PR #8, opened 2016-01-08
Updated io_gdal. Fixes JIRA AUTOCONG-58

jcwbacker:


jlaura:
This looks super good.  Two things that would be good to update before merging:

1. The `import numpy as np` should stay.  That is canonical Python.  For other modules, If the module name is long, the `as` is traditionally used.
2. It is not necessary to document `@property` methods as they are documented in the `Attributes` section at the top of the class.  The double documentation is hard to keep synchronized sometimes.

jcwbacker:
Ok. Thanks Jay. Will update as recommended and let you know when it's ready to be pulled.

PR #21, opened 2016-01-15
documentation and other tweaks

kree:


jlaura:
I see that build #84 passed - did you rerun this same PR with a new build?  The code looks clean and I see a seg fault on Travis' side.  Saw this earlier today and it looks like an issue with --with-coverage that is external to us.

kree:
Build #84 seems to be for PR #20. I did try re-starting the travis build for this PR, but it consistently segfaults. I just tried again and the results are the same.

kree:
Yes, that was it!

jlaura:
Awesome!  Looks like it would make sense to check what the change from 3.0.0 to 3.1.0 is - a future item.

PR #22, opened 2016-01-22
Added functionality to compute homography to CandidateGraph

kree:
I added compute_homography to the CandidateGraph class. It calculates the homography for an edge of the graph defined by an input source node and destination node, and returns the transformation matrix and also a bad-match mask from RANSAC. Looking for feedback on this, especially on how I can improve testing and error-checking. 

References AUTOCONG-65.


kree:
I'll need to update this to incorporate @jlaura's changes to the matches dataframe.

kree:
I've actually updated this now to incorporate the matches df change.

PR #23, opened 2016-01-25
Nimage matching

jlaura-usgs:


jlaura:
Merged without finishing Travis (local tests are passing) and Travis is having network issues...

PR #24, opened 2016-01-26
Outlier detection functionality added

kree:
OutlierDetector.distance_ratio() currently only works as expected if you pass it a "matches" dataframe with duplicates already dropped. I'm working on a way to not require this, but figured I should get my existing code in. Fixes JIRA #29


jlaura:
:+1:

PR #25, opened 2016-01-26
Added convenience methods to CandidateGraph class. Fixes JIRA AUTOCONG-72

jcwbacker:
Updated node attributes image_name and image_path to be prepended with current working directories if relative paths are given.


kree:
I'm going to go ahead and declare a -0.6% decrease in code coverage as "epsilon" at this point and merge.

PR #28, opened 2016-01-27
Added visualization utilities and example notebooks. Fixes JIRA AUTOCONFIG 68, 70

jcwbacker:


.gitignore
```diff
@@ -1,5 +1,6 @@
 #iPython
 /.ipynb*
+/.ipynb*/
```
kree: It looks like this line was accidentally repeated?

.gitignore
```diff
@@ -1,5 +1,6 @@
...
```
kree: It looks like this line was accidentally duplicated?

kree:
Also, just to check, did you deliberately switch back to using from_adjacency, rather than from_adjacency_file ?

autocnet/utils/visualization.py
```diff
@@ -0,0 +1,288 @@
...
+    # create a subplot with its own axes
+    ax1 = fig.add_subplot(rows, columns, 1)
+    ax1.axis('off')
+    ax1.margins(tight=True)
+    plotFeatures(imageName1, 
+                 graph.get_keypoints(imageName1), 
+                 'ro', 
+                 featurePointSize)
+    ax2 = fig.add_subplot(rows, columns, 2)
+    ax1.axis('off')
```
jlaura: @jcwbacker Should this be `ax2.axis('off')`?

.gitignore
jcwbacker: The line was, in fact, intentionally duplicated. 
> +/.ipynb*/

When running "git status" my ipython notebook backup folders keep showing up in the list of files. Jay thought adding the slash might help, but it didn't seem to make a difference, so if you deleted it, then that's fine

autocnet/utils/visualization.py
jcwbacker: yes

jcwbacker:
from_adjacency_file() was changed back to from_adjacency() since it now accepts a file or a dict

PR #31, opened 2016-01-28
Subpixel offset calculation added

kree:
I added a subpixel offset calculation both as a standalone function and also to the graph class, so that ithe graph can calculate subpixel offsets for each keypoint on each edge and then add it as an attribute on the edge.

It's still inefficient and doesn't have the best way of dealing with "can't find a match," so I'm going to continue work on this, but it seems to work so someone can pull this now if they want to use it.


jlaura:
:+1:

PR #35, opened 2016-01-29
Added adaptive non-maximal suppression. Fixes JIRA #71

kree:
Added adaptive_non_max_suppression, which will compute and return a mask for the keypoint list we have stored as an attribute for each node. The mask will have 1's in the places of the top n maximum minimum distances to another point that is significantly stronger. See http://graphics.cs.cmu.edu/courses/15-463/2005_fall/www/Papers/MOPS.pdf for more information.


jlaura:
:+1:

PR #36, opened 2016-02-01
Offset validation

jlaura:
This PR fixes a regression where extraction parameters could not be specified via convenience functions, and moved template extraction for sub pixel matching to a stand alone function.


kree:
This looks great to me!

PR #38, opened 2016-02-06
Offset validation

jlaura:
This is a number of small fixes that grew from testing  The biggy is that downsampling is not a good idea, so backed completely out.

### DIFF
--- a/autocnet/fileio/io_gdal.py
+++ b/autocnet/fileio/io_gdal.py
@@ -142,6 +142,11 @@ class GeoDataset(object):
 
     @property
     def geotransform(self):
+      """
+      Where the array is in the form:
+      [top left x, w-e pixel resolution, x-rotation,
+      top left y, y-rotation, n-s pixel resolution]
+      """
         if not getattr(self, '_geotransform', None):
             self._geotransform = self.dataset.GetGeoTransform()
         return self._geotransform
--- a/autocnet/fileio/io_gdal.py
+++ b/autocnet/fileio/io_gdal.py
@@ -142,11 +142,11 @@ class GeoDataset(object):
 
     @property
     def geotransform(self):
-      """
-      Where the array is in the form:
-      [top left x, w-e pixel resolution, x-rotation,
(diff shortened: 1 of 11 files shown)


kree:
Hey @jlaura, could you take a look at the failing tests? It's not just the usual "n isn't in range (a,b)"

jlaura:
I will fix these - tests need to be updated for a bug.

autocnet/fileio/io_gdal.py
```diff
@@ -192,10 +192,10 @@ def xy_extent(self):
         if not getattr(self, '_xy_extent', None):
             geotransform = self.geotransform
             minx = geotransform[0]
-            maxy = geotransform[3]
+            miny = geotransform[3]
 
             maxx = minx + geotransform[1] * self.dataset.RasterXSize
-            miny = maxy + geotransform[5] * self.dataset.RasterYSize
+            maxy = miny + geotransform[5] * self.dataset.RasterYSize
```
kree: Could you document somewhere what geotransform[0-5] are? It's pretty obvious what the 4 here mean because of context, but it might be nice to have something up where self.geotransform is set to succinctly explain what it contains.

autocnet/matcher/outlier_detector.py
```diff
@@ -56,15 +56,22 @@ def distance_ratio(matches, ratio=0.8):
...
-            unique = sorted['distance'].unique()
-            if unique[0] < ratio * unique[1]:
+            sorted_group = group.sort_values(by=['distance'])
+            unique = sorted_group['distance'].unique()
+
+            if len(unique) == 1:
+                # The distances from the unique points are identical
+                mask[counter: counter + group_size] = False
+                counter += group_size
+            elif unique[1] / unique[0] < ratio:
```
kree: Are you sure this is right? I think things sort by ascending by default (?), so unique[0] should be less than unique[1]. So, that fraction will be > 1, not < 1, as desired.

PR #39, opened 2016-02-08
Added ability to produce a fromlist (for isis) from the graph. References JIRA #83

kree:
I added the ability for **the graph** to produce a fromlist (and then later write it to a file), since the fromlist only cares about which images are included. Right now we start with an adjacency dict we turn into a graph, so we know that all nodes(images) should be included in the output fromlist. 

This is very simplistic, so please let me know if any of my above assumptions are wrong or anything else needs to be changed! We could later call this on a connected-subgraph to produce the fromlist for a connected subset of images.


jlaura:
:+1:

PR #43, opened 2016-02-09
Convex Hull Computations

jlaura:


autocnet/graph/network.py
```diff
@@ -226,13 +228,34 @@ def compute_homography_overlap(self):
...
         Returns
         -------
+        source_geom : object
+                      PySAL Polygon object of the source pixel bounding box
+
+        projected_geom : object
+                         PySAL Polygon object of the destination geom projected
+                         into the source reference system using the current
+                         homography
+
```
kree: From this, is it implied that the return value is actually the tuple (source_geom, projected_geom, area) ?

jlaura: Yes, but you would not document it that way, because the function return could be handled as:

```python
def foo():
    return a, b

x, y = foo()
```

In fact, this is how you would expect the return to be used as it avoids:

```python
x = foo
x[0].do_something
z = bar(x[1])
```

autocnet/graph/network.py
```diff
@@ -226,13 +228,34 @@ def compute_homography_overlap(self):
...
 
-    def plot(self, clean_keys=[], **kwargs):
-        return plot_edge(self, clean_keys=clean_keys, **kwargs)
+        source_geom = self.source.handle.pixel_polygon
+        destination_geom = self.destination.handle.pixel_polygon
+
+        # Project using the homography
+        vertices_to_project = destination_geom.vertices
+        for i, v in enumerate(vertices_to_project):
+            vertices_to_project[i] = tuple(np.array([v[0], v[1], 1]).dot(self.homography)[:2])
```
kree: Is the [:2] here because this is 2d? (If not, maybe you can walk me through what this does? Is v[0] a single scalar value or multidimensional?)

kree:
I'm happy to merge this whenever. I had one minor documentation comment and a question about whether I understand your math or not.

autocnet/graph/network.py
kree: Yeah, the x, y = foo() usage makes more sense to me--I just wanted to make sure that whatever parses the docstrings to give help information/type hints would know what to do.

autocnet/graph/network.py
jlaura: Because the homography is a 3x3, so we have to extend the point out to be a 3d vector.  The polygon coordinates are 2 though, so `[:2]`. 

`v = [x, y]`, so this is just slicing the first two entries and adding in a 1 to get the 3,1 shape.

kree: Awesome! That's what I thought/meant by "because it is 2d," just wanted to make sure I understood.

PR #45, opened 2016-02-10
Jigsaw test notebook + json file added.

kree:
References JIRA #77, but doesn't close the issue. Jigsaw should probably be tested more extensively.


kree:
I'm pretty sure this test is failing for the same reason as @jlaura's PR #44: problems with the gdal package.

jlaura:
It is. Working on a gdal build.

PR #49, opened 2016-02-16
Added example notebook for ANMS

kree:
I just added a quick example notebook which shows a couple of images pre- and post- running ANMS (adaptive non-maixmal suppression,) so that we can demo this if we need to.


jlaura:
:+1:

PR #51, opened 2016-02-24
Fixes serialization for save/load

jlaura:


kree:
Looks reasonable to me! Should I go ahead and merge this?

PR #53, opened 2016-02-29
Updates to ratio test for performance

jlaura:


autocnet/examples/Apollo15/three_image_adjacency.json
```diff
@@ -1,3 +1,3 @@
 {"AS15-M-0297_SML.png": ["AS15-M-0298_SML.png", "AS15-M-0299_SML.png"],
-"AS15-M-0298_SML.png": ["AS15-M-0297_SML.png", "AS15-M-0299_SML.png"],
-"AS15-M-0299_SML.png": ["AS15-M-0297_SML.png", "AS15-M-0298_SML.png"]}
+"AS15-M-0298_SML.png": ["AS15-M-0299_SML.png", "AS15-M-0297_SML.png"],
+"AS15-M-0299_SML.png": ["AS15-M-0298_SML.png", "AS15-M-0297_SML.png"]}
```
kree: What does switching the order here do?

autocnet/matcher/outlier_detector.py
```diff
@@ -1,8 +1,126 @@
...
+
+        self._action_stack = deque(maxlen=10)
+        self._current_action_stack = 0
+        self._observers = set()
+        self.matches = matches
+        self.mask = None
+
+    @property
+    def nvalid(self):
+        return self.mask.sum()
```
kree: I know that nvalid is pretty self-descriptive, but could you put a small comment in here anyway about "returns number of valid matches" or something like that? nvalid is visually very close to invalid, so it seems good to make this explicit.

autocnet/matcher/outlier_detector.py
```diff
@@ -1,8 +1,126 @@
...
+                the ratio between the first and second-best match distances
+                for each keypoint to use as a bound for marking the first keypoint
+                as "good". Default: 0.8
+        Returns
+        -------
+         mask : ndarray
+                Intended to mask the matches dataframe. Rows are True if the associated keypoint passes
+                the ratio test and false otherwise. Keypoints without more than one match are True by
+                default, since the ratio test will not work for them.
+
```
kree: Unless I misunderstand what is happening in func, the last sentence of this comment is not true and should be updated. Keypoints without more than one match will be False, *not* True.

autocnet/matcher/outlier_detector.py
```diff
@@ -1,8 +1,126 @@
...
+            res = [False] * len(group)
+            if len(res) == 1:
+                return res
+            if group.iloc[0] < group.iloc[1] * ratio:
+                res[0] = True
+            return res
+
+        if mask is not None:
+            self.mask = mask.copy()
+            new_mask = self.matches[mask].groupby('source_idx')['distance'].transform(func).astype('bool')
```
kree: Nice! :+1:

kree:
Okay, those are all of my comments. I'm happy to merge after you add/change the two documentation things I mentioned and let me know what switching the order in the json adjacency file does (I didn't think that it mattered.)

autocnet/examples/Apollo15/three_image_adjacency.json
jlaura: I was trying to induce a bug - we have an issue where edges are getting added with (higher node key, lower node key).  This breaks logic downstream.  I can reproduce intermittently with 4+ images.

autocnet/matcher/outlier_detector.py
jlaura: Addressed as part of a large docos update.

PR #56, opened 2016-03-09
Create a graph from a filelist (calculate adjacency structure)

kree:
This works for our "Projection" examples, but not the Apollo pngs or the ISIS cubes. I'm still working on figuring out how to make this work for Cubes, especially.


jlaura:
sweet!  Will check the out soon.

autocnet/graph/network.py
```diff
@@ -5,7 +5,9 @@
 import numpy as np
 import pandas as pd
 
-
+from pysal import cg
+from autocnet.examples import get_path
```
jlaura: extraneous import?

autocnet/graph/network.py
```diff
@@ -83,6 +85,44 @@ def from_graph(cls, graph):
...
+          A Network graph object
+        """
+
+        # TODO: Reject unsupported file formats + work with more file formats
+
+        dataset_list = []
+        for file in filelst:
+            dataset = GeoDataset(file)
+            dataset_list.append(dataset)
+
```
jlaura: Can these two for loops be combined into one so that a potentially large number of file handles are not open at the same time?

jlaura:
We should chat about the GeoDataSet class and supporting ISIS cubes in there.  This could take the form of a fallback to using PVL to try and get some data.

autocnet/graph/network.py
```diff
@@ -83,6 +85,44 @@ def from_graph(cls, graph):
...
+        filelst : list
+                  A list containing the files (with full paths) to construct an adjacency graph from
+
+        Returns
+        -------
+        : object
+          A Network graph object
+        """
+
+        # TODO: Reject unsupported file formats + work with more file formats
```
jlaura: The rejection logic should probably live the with geodataset object.  This might mean that this call needs to get wrapped in a `try/except` so that it can fail gracefully, warn, and move on.

jlaura:
@kree If possible, ca we close this today.  I have a sprint review tentatively scheduled for Friday (I think you will be away, which is fine), so we can demo this functionality.

kree:
I'm still working on the ISIS Polygon functionality. I'll open a new PR for this?

jlaura:
A new one would be great.  I wanted to get this in so it is 'demoable'.

PR #58, opened 2016-03-09
Adding Minimum Spanning Tree Method to CandidateGraph

Kelvinrr:
The mst in a Pandas DataFrame as a boolean mask. The DataFrame now an attribute of CandidateGraph.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -45,6 +45,7 @@ class CandidateGraph(nx.Graph):
         self.node_counter = 0
         node_labels = {}
         self.node_name_map = {}
+        self.graph_masks = {}
 
         # the node_name is the relative path for the image
         for node_name, node in self.nodes_iter(data=True):
@@ -290,6 +291,45 @@ class CandidateGraph(nx.Graph):
                  of keyword arguments to be passed through to the func
         """
         _, self.clusters = func(self, *args, **kwargs)
+
+    def apply_func_to_edges(self, func, *args, graph_mask_keys=[], **kwargs):
+        """
+        Iterates over edges using an optional mask and and applies the given function.
+        If func is not an attribute of Edge, raises AttributeError
+        Parameters
+        ----------
+        func : string
+               function to be called on every edge
+        graph_mask_keys : list
+                          of keys in graph_masks
+        """
+
+        merged_graph_mask = self.graph_masks[graph_mask_keys].all(axis=1)
+        edges_to_iter = merged_graph_mask[merged_graph_mask].index
+
+        for s, d in edges_to_iter:
+            curr_edge = self.get_edge_data(s, d)
+            try:
+                function = getattr(curr_edge, func)
+            except:
+                raise AttributeError('The passed function is not an attribute of Edge')
+            else:
+                function(*args, **kwargs)
+
+    def minimum_spanning_tree(self):
+        """
+        Calculates the minimum spanning tree of the graph
+
+        Returns
+        -------
+
+         : DataFrame
+           boolean mask for edges in the minimum spanning tree
+        """
+
+        self.graph_masks = pd.DataFrame(False, index=self.edges(), columns=['mst'])
+        mst = nx.minimum_spanning_tree(self)
+        self.graph_masks['mst'][mst.edges()] = True
 
     def symmetry_checks(self):
         """
--- a/autocnet/graph/tests/test_network.py
+++ b/autocnet/graph/tests/test_network.py
@@ -99,3 +99,27 @@ class TestEdge(unittest.TestCase):
     def setUpClass(cls):
         cls.graph = network.CandidateGraph.from_adjacency(get_path('adjacency.json'))
 
+
+class TestMSTGraph(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        cls.test_dict = {"0": ["4", "2", "1", "3"],
+                         "1": ["0", "3", "2", "6", "5"],
+                         "2": ["1", "0", "3", "4", "7"],
+                         "3": ["2", "0", "1", "5"],
+                         "4": ["2", "0"],
+                         "5": ["1", "3"],
+                         "6": ["1"],
+                         "7": ["2"]}
+
+        cls.graph = network.CandidateGraph.from_adjacency(cls.test_dict)
+        cls.graph.minimum_spanning_tree()
+        cls.mst_graph = cls.graph.copy()
+
+        for s, d, edge in cls.graph.edges_iter(data=True):
+            if not cls.graph.graph_masks['mst'][(s, d)]:
+                cls.mst_graph.remove_edge(s, d)
+
+    def test_mst_output(self):
+        self.assertEqual(self.mst_graph.nodes(), self.graph.nodes())
+        self.assertEqual(self.mst_graph.number_of_edges(), self.graph.number_of_edges()-5)
--- a/autocnet/graph/tests/test_network.py
+++ b/autocnet/graph/tests/test_network.py
@@ -40,6 +40,23 @@ class TestCandidateGraph(unittest.TestCase):
     def test_island_nodes(self):
         self.assertEqual(len(self.disconnected_graph.island_nodes()), 1)
 
+    def test_apply_func_to_edges(self):
+        graph = self.graph.copy()
+        graph.minimum_spanning_tree()
+
+        try:
+            graph.apply_func_to_edges('incorrect_func')
+        except AttributeError:
+            pass
+
+        graph.extract_features(extractor_parameters={'nfeatures': 500})
+        graph.match_features()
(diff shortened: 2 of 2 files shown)


autocnet/graph/network.py
```diff
@@ -492,6 +493,21 @@ def island_nodes(self):
...
+        Returns
+        -------
+         : DataFrame
+           boolean mask for edges in the minimum spanning tree
+        """
+
+        self.graph_masks = pd.DataFrame(False, index=self.edges(), columns=['mst'])
+        mst = nx.minimum_spanning_tree(self)
+        self.graph_masks['mst'].loc[mst.edges()] = True
+        return self.graph_masks
```
jlaura: This return is not needed, line 508 adds the attribute.  Python will implicitly return None if a return is not specified.  If you remove line 509 and then recommit, this PR will be updated.

jlaura:
Nice PR.  The next step here will be to test using the mask.  Anywhere that the `CandidateGraph` class is iterating over edges, we need to now have the ability to use said mask.  I will leave this PR open so that you can figure out how to add that functionality.

Kelvinrr:
Alright, I'll begin looking into that tonight.

Kelvinrr:
I imagine I can ignore failed tests for now. Current error seems to be from PR #60, not my tests.

Kelvinrr:
hmmmmm, I don't know what happened with this last update. I think it included changes from other commits from other people.

jlaura:
Please merge the latest updates and then resubmit.

Kelvinrr:
Alright, I think I got them all. Writing some tests.

jlaura:
The other action item here is going to be to squash all of these commits: http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

Kelvinrr:
Done, one commit for the mst method and one for adding the calls to the other methods.

jlaura:
@Kelvinrr The build is failing.  Please fix.

Kelvinrr:
yeah, I put assertTrue(False) for one of the tests. I just wanted to push it to my remote repo is I can possibly work on it while home sick. I'll fix it as soon as I am OK.

autocnet/graph/edge.py
```diff
@@ -113,7 +113,7 @@ def symmetry_check(self):
         else:
             raise AttributeError('No matches have been computed for this edge.')
 
-    def ratio_check(self, clean_keys=[], **kwargs):
+    def ratio_check(self, clean_keys, **kwargs):
```
jlaura: Why are you removing an optional kwarg and making it a required positional argument?

autocnet/graph/network.py
```diff
@@ -307,21 +309,49 @@ def add_matches(self, matches):
...
+
         for s, d, edge in self.edges_iter(data=True):
-            edge.symmetry_check()
+            if use_mst and self.graph_masks['mst'][(s, d)]:
+                edge.symmetry_check()
+            else:
+                edge.symmetry_check()
 
-    def ratio_checks(self, clean_keys=[], **kwargs):
+    def ratio_checks(self, clean_keys, use_mst=False):
```
jlaura: This should not alter the `clean_keys` optional kwarg, or remove `**kwargs` that piping through this method to the method with the logic on the edge.

autocnet/graph/network.py
```diff
@@ -307,21 +309,49 @@ def add_matches(self, matches):
...
+        ----------
+        clean_keys : list
+                     Of keys in the mask dict
+
+        use_mst : Boolean
+                  Set to true if you only want to iterate through mst,
+                  false if you want to iterate through all nodes
+
         """
+        if use_mst and self.graph_masks['mst'].isnull().values.any():
```
jlaura: we do not want a `use_mst` boolean as this couples all of these methods to using a `mst` mask.  What if I have a different mask?  This is where you should be passing in a list of strings that are keys into a pandas dataframe (stored in graph_masks).  You can check the logic in the Edge class for how clean keys are managed on the edge.

The code should take the list of graph clean_keys and make a single mask, where the index is the edge key ((0,1) for example, and the value is a bool).

You can then do:

```python
merged_graph_mask = # implement me
edges_to_iter = merged_graph_mask[merged_graph_mask == True].index
for s, d, e in self.edges(edges_to_iter, data=True):
    e.some_func_on_the_edge()
    
```

autocnet/graph/network.py
Kelvinrr: deleting `**kwargs` was definitely unintentional. I removed the `=[]` to stop pep8 from complaining. I didn't think it would be very disadvantageous. I can change it back. 

Alternatively, I can change it to `clean_keys=none` and in the body put something like `if clean_keys is none:
    clean_keys=[]` and get the same behavior without the mutable list and without pep8 throwing a fit.

jlaura: Please look at [this tutorial](http://www.tutorialspoint.com/python/python_functions.htm) for the difference between required and optional arguments.  Also, take a look at the difference between `False`, `None`, and `[]`.

`clean_keys=[]` is intentional.

autocnet/graph/network.py
Kelvinrr: If `merged_graph_mask` contain booleans, wouldn't `merged_graph_mask == True` be redundant?  `edges_to_iter = merged_graph_mask[merged_graph_mask].index` seems to work. Unless you want the explicit `== True`?

Kelvinrr:
I added the function to process edges, though I am not sure of the name. I am open to hear any suggestion to change it.

autocnet/graph/network.py
```diff
@@ -307,17 +310,51 @@ def add_matches(self, matches):
...
+        edges_to_iter = merged_graph_mask[merged_graph_mask].index
+
+        for s, d in edges_to_iter:
+            curr_edge = self.get_edge_data(s, d)
+            function = getattr(curr_edge, func)
+            if function:
+                function(*args, **kwargs)
+            else:
+                raise AttributeError(func, ' is not an attribute of Edge')
+
```
Kelvinrr: I don't like that the if statement is in every iteration, if someone can suggest a better way to do this, I am all ears.

autocnet/graph/edge.py
jlaura: This looks like it is still broken?

Kelvinrr: I definitely changed all those lines, I double checked I changed that like in particular because, for some reason, it is still here. I'll look at it tomorrow morning, but I definitely changed it in my local copy.  Don't know why it is not reflected here.

autocnet/graph/network.py
```diff
@@ -309,26 +310,38 @@ def add_matches(self, matches):
...
-    def ratio_checks(self, clean_keys, use_mst=False):
+    def symmetry_checks(self):
+        """
+        Perform a symmetry check on all edges in the graph
+        """
+
+        for s, d, edge in self.edges_iter(data=True):
+            edge.symmetry_check()
+
+    def ratio_checks(self, clean_keys=[], **kwargs):
```
Kelvinrr: Here, it is back to normal

autocnet/graph/network.py
```diff
@@ -307,17 +309,53 @@ def add_matches(self, matches):
...
+        edges_to_iter = merged_graph_mask[merged_graph_mask].index
+
+        for s, d in edges_to_iter:
+            curr_edge = self.get_edge_data(s, d)
+            try:
+                function = getattr(curr_edge, func)
+            except:
+                raise AttributeError('The passed function is not an attribute of Edge')
+            else:
+                function(*args, **kwargs)
```
jlaura: Is the function call in an else instead of inside the try intentionally?

autocnet/graph/network.py
```diff
@@ -523,7 +565,7 @@ def to_json_file(self, outputfile):
         ----------
 
         outputfile : str
-                     PATH where the JSON will be written
+                     PATH where the JSON will be writtenself.graph.graph_masks['mst'].loc(s, d)
```
jlaura: weird line concatenation only in the diff?

jlaura:
Looks good. Two comments inline.

Can you give this one more squash to get it down to a single commit?

autocnet/graph/network.py
Kelvinrr: Yeah, my thinking was only attempt to call the function if it exists as an attribute and raise that particular error if (and only if) the function didn't exist.

jlaura:
The other thing to do, before merge, is to get this PR in sync with master.

Kelvinrr:
I imagine the decrease in coverage is from some other PR?

jlaura:
No. This is your code.

Kelvinrr:
Never worked with something like coveralls, but I am guessing this is because I didn't include tests for apply_func_to_edges. Correct?

jlaura:
:+1:

PR #60, opened 2016-03-11
Basic 'calculate overlap' funcationality working* on cubes

kree:
This seems to be working, but I didn't test it very well (see my tests.)

latlon_extent is not used for Cubes right now. I added _pvl_header and _footprint_polygon to GeoDataset to support Cubes.

I'm not sure that reorganizing footprint_polygon.GetEnvelope() into the right shape for a bounding box is doing the right thing, so someone should double-check.

### DIFF
--- a/autocnet/fileio/io_gdal.py
+++ b/autocnet/fileio/io_gdal.py
@@ -1,6 +1,10 @@
+import pvl
 import os
+import warnings
 
 from pysal.cg.shapes import Polygon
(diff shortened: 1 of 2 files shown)


kree:
The online tests will fail because I didn't move the (large) Apollo cubes into the repository. They'll pass locally if you move them into the examples area. Thoughts on how to deal with this?

jlaura:
Merging and will PR some modifications later.

PR #65, opened 2016-03-16
Pep8 example for --aggressive --ignore E501(line length)

acpaquette:
Basic example of running autopep8 on edge with only one aggressive argument and ignoring line length.


jlaura:
@acpaquette Is this completed or do you have a few additions (another commit) to make?

acpaquette:
I have one more commit to make, should be done in a few minutes.

jlaura:
:+1:

PR #66, opened 2016-03-16
Graph Abstraction

jlaura:
and docos updates.


autocnet/graph/markov_cluster.py
```diff
@@ -0,0 +1,108 @@
...
+
+    inflate_factor : float
+                     Parameter to strengthen and weaken flow between nodes.  The larger the value
+                     the more granular the resultant clusters are.
+
+    expand_factor : int
+                    Parameter to manage flow connection between different regions of the graph.
+
+    mult_factor : int
+                  Value to set for self loops.
```
kree: Maybe explain this one a little better? I can't tell what this means.

autocnet/graph/markov_cluster.py
```diff
@@ -0,0 +1,108 @@
...
+    if isinstance(g, nx.Graph):
+        arr = np.array(nx.adjacency_matrix(g).todense())
+    elif isinstance(g, np.ndarray):
+        arr = g
+
+    arr = _add_diag(arr, mult_factor)
+    arr = _normalize(arr)
+
+    for i in range(max_loop):
+        #logging.info("loop", i)
```
kree: Did you want to take this comment out?

autocnet/graph/markov_cluster.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def _expand(arr, expand_factor):
+        return np.linalg.matrix_power(arr, expand_factor)
+
+    def _add_diag(arr, mult_factor):
+        return arr + mult_factor * np.identity(arr.shape[0])
+
+    def _stop(arr, i):
+
+        if i%5==4:
```
kree: Why check when i%5 == 4?

autocnet/graph/markov_cluster.py
```diff
@@ -0,0 +1,108 @@
...
+        column_sums = arr.sum(axis=0)
+        new_matrix = arr / column_sums[np.newaxis, :]
+        return new_matrix
+
+    def _inflate(arr, inflate_factor):
+        return _normalize(np.power(arr, inflate_factor))
+
+    def _expand(arr, expand_factor):
+        return np.linalg.matrix_power(arr, expand_factor)
+
```
kree: I don't really understand _inflate or _deflate or the 'flow between nodes' but just to make sure: the `np.power` vs. `np.linalg.matrix_power` is deliberate, right?

autocnet/graph/network.py
```diff
@@ -307,6 +272,25 @@ def add_matches(self, matches):
...
+               Markov Clustering Algorithm
+
+        args : list
+               of arguments to be passed through to the func
+
+        kwargs : dict
+                 of keyword arguments to be passed through to the func
+        """
+        _, self.clusters = func(self, *args, **kwargs)
+
```
kree: This is a nice way of dealing with not knowing what func might be supplied.

autocnet/graph/network.py
```diff
@@ -583,3 +567,20 @@ def plot(self, ax=None, **kwargs):
...
+        return subgraph
+
+    @create_subgraph.register(pd.DataFrame)
+    def _(self, arg, g, key=None):
+        if key == None:
+            return
+        col = arg[key]
+        nodes = col[col == True].index
+        subgraph = g.subgraph(nodes)
+        return subgraph
```
kree: PEP wants a newline.

kree:
Okay, I'm done commenting. I don't totally understand the @singledispatch & the @subgraph.register stuff yet (reading up on it), but other than that, everything looks good to me!

jlaura:
For how (and why) inflate and deflate are being used, checkout: https://www.cs.ucsb.edu/~xyan/classes/CS595D-2009winter/MCL_Presentation2.pdf.

The presentation describes the algorithm pretty well.  The logic basically has to do with powering and then normalizing the matrix - do this enough and the 'big' numbers get bigger and the 'small' numbers get smaller.

autocnet/graph/markov_cluster.py
jlaura: they are.  I added a comment below with a link to a powerpoint that describes this.  The thesis and technical report that describe the algorithms should be linked in the references, but I would not start there! ;)

autocnet/graph/markov_cluster.py
jlaura: Just a, check every so often for convergence, line.  The powering and comparison could be expensive with many nodes.

kree:
Thanks for the inflate/deflate reference!

PR #67, opened 2016-03-17
Basic implimentation of image_match.py

acpaquette:


autocnet/examples/Apollo15/user_image_adjacency.json
```diff
@@ -0,0 +1,5 @@
+{
+	"AS15-M-0414_sub4.cub": ["AS15-M-0413_sub4.cub", "AS15-M-0412_sub4.cub"],
+	"AS15-M-0413_sub4.cub": ["AS15-M-0414_sub4.cub", "AS15-M-0412_sub4.cub"],
+	"AS15-M-0412_sub4.cub": ["AS15-M-0413_sub4.cub", "AS15-M-0414_sub4.cub"]
+}
```
jlaura: You can actually completely omit this file.  We would ship it if we know the user would have this data and / or we were testing the script.

bin/image_match.py
```diff
@@ -0,0 +1,42 @@
...
+cnet = cg.to_cnet(clean_keys=['subpixel'], isis_serials=True)
+
+filelist = cg.to_filelist()
+write_filelist(filelist, 'TestList.lis')
+
+to_isis('TestList.net', cnet, mode='wb', targetname='Moon')
+
+# Ticket calls for a user specified "file list".
+# What kind of "file list" should this ui take? Should it be in the form of a .json file or should we allow the user
+# to enter the images he/she wants to look at in particular and parse them into a .json?
```
jlaura: Once you get to taking argument via arg parse, you could have a flag where the user tells you whether the file is a list of images or a json adjacency matrix.  In the case of the former the call to create the graph becomes `from_filelist`.

bin/image_match.py
```diff
@@ -0,0 +1,42 @@
...
+import sys
+import argparse
+
+sys.path.insert(0, os.path.abspath('../autocnet'))
+
+from autocnet.graph.network import CandidateGraph
+from autocnet.fileio.io_controlnetwork import to_isis
+from autocnet.fileio.io_controlnetwork import write_filelist
+
+cg = CandidateGraph.from_adjacency(args.filename, basepath='/home/acpaquette/Desktop/')
```
jlaura: Can you clean out the basepath that has your name, etc.  You can pass `basepath=''`.  This is just sanitation for deployment.

acpaquette:
Managed to delete the entire docs directory, fun. I have reverted my local branch but I am having trouble pushing that to the main due to the tip of the branch not being current with origin.

jlaura:
:+1:

PR #69, opened 2016-03-18
Adds HDF io Module

jlaura:
This is something to discuss tomorrow re: API.  Questions:

* Descriptors and keypoints are stored in an HDF5 file.  How does the user specify when these might exist already?  Is this done into the network?  Or when the adjacency matrix / file list are read?
* How do we let the user indicate that they want to save the keypoints?
* Image identifiers.  ISIS uses serial numbers.  How do we want to handle this?

Please do not merge.


jlaura:
Good to know - failure is due to a Cython dependency and this OS.  Something to watch for.

jlaura:
This is finally ready for review.  Anyone can have a whack at it whenever is convenient.

Something to discuss - right now the image name is used as the key in the HDF.  I really dislike this approach.  My preferred approach would be to compute an MD5 hash of the image.  This would work fine in the case of most formats, but not ISIS, because ISIS can modify the PVL header and append history.  Those actions change the hash.

An potential solution would be to compute the hash using the binary data if we hit an ISIS cube.  I do not want to have custom, `if isis` all over the place...

autocnet/graph/network.py
```diff
@@ -239,7 +241,64 @@ def extract_features(self, method='orb', extractor_parameters={}):
...
+
+        if os.path.exists(out_path):
+            mode = 'a'
+        else:
+            mode = 'w'
+        hdf = io_hdf.HDFDataset(out_path, mode=mode)
+
+        # Cleaner way to do this?
+        if nodes:
+            for i, n in self.nodes_iter(nodes, data=True):
```
Kelvinrr: Is this valid? I can't find anywhere in the networkx docs that shows `nodes_iter` can traverse a subset of nodes.

autocnet/graph/network.py
```diff
@@ -239,7 +241,64 @@ def extract_features(self, method='orb', extractor_parameters={}):
...
+                n.save_features(hdf)
+        else:
+            for i, n in self.nodes_iter(data=True):
+                n.save_features(hdf)
+
+        hdf = None
```
Kelvinrr: Why `hdf = none`? Wouldn't the `__del__` be called when `hdf` leaves the stack frame regardless and close the file?

autocnet/graph/network.py
```diff
@@ -239,7 +241,63 @@ def extract_features(self, method='orb', extractor_parameters={}):
...
+        hdf = io_hdf.HDFDataset(in_path, 'r')
+
+        if nodes:
+            for i, n in self.nodes_iter(nodes, data=True):
+                n.load_features(hdf)
+        else:
+            for i, n in self.nodes_iter(data=True):
+                n.load_features(hdf)
+
+        hdf = None
```
Kelvinrr: Probably should get rid of all of them for the sake of consistency.

jlaura:
Updated with an intentional `hdf=None`.  This ensures that the file is properly closed and does not need to wait for garbage collection.  Corrected the iterator in the load method.

autocnet/fileio/io_hdf.py
```diff
@@ -0,0 +1,84 @@
...
+            return b
+        if index_column is not None:
+            index = sarray[index_column]
+            clean_array = remove_field_name(sarray, 'index')
+        else:
+            clean_array = sarray
+            index = None
+        columns = clean_array.dtype.names
+
+        return pd.DataFrame(data=sarray, index=index, columns=columns)
```
kree: PEP 8 wants a newline.

autocnet/utils/tests/test_utils.py
```diff
@@ -92,3 +92,12 @@ def test_make_homogeneous(self):
...
         np.testing.assert_array_equal(pts[:, -1], np.ones(25))
+
+
+    def test_remove_field_name(self):
+        starray = np.array([(1 ,2.,'String'), (2, 3.,"String2")],
+              dtype=[('index', 'i4'),('bar', 'f4'), ('baz', 'S10')])
+        truth = np.array([(2.,'String'), (3.,"String2")],
+              dtype=[('bar', 'f4'), ('baz', 'S10')])
+        cleaned_array = utils.remove_field_name(starray, 'index')
+        np.testing.assert_array_equal(cleaned_array, truth)
```
kree: PEP8 newline needed

autocnet/utils/utils.py
```diff
@@ -159,3 +159,28 @@ def make_homogeneous(points):
        n x m + 1 array of homogeneous points
     """
     return np.hstack((points, np.ones((points.shape[0], 1))))
+
+
+def remove_field_name(a, name):
+    """
+    Given a numpy structured array, remove a column and return
+    a copy of the remaineder of the array
+
```
kree: remaineder -> remainder

autocnet/utils/utils.py
```diff
@@ -159,3 +159,28 @@ def make_homogeneous(points):
...
+    Returns
+    -------
+    b : ndarray
+        Numpy structured array with the 'name' column removed
+    """
+    names = list(a.dtype.names)
+    if name in names:
+        names.remove(name)
+    b = a[names]
+    return b
```
kree: PEP 8 newline needed.

kree:
@jlaura, @Kelvinrr asked me to take a quick look at this, so I added some quick PEP & documentation comments.

PR #70, opened 2016-03-18
Image match

acpaquette:
General modulation of image_match.py and the addition of argument parsing to the script.


bin/image_match.py
```diff
@@ -8,30 +8,47 @@
...
-# Apply SIFT to extract features
-cg.extract_features(method='sift', extractor_parameters={'nfeatures': 1000})
+def match_images(args):
 
-# Match
-cg.match_features()
+    # Matches the images in the input file using various candidate graph methods
+    # produces two files usable in isis
+    try:
+        cg = CandidateGraph.from_adjacency(args.input_file, basepath='/home/acpaquette/Desktop/')
```
jlaura: This base path will not work.  I would add an argument above and then pipe that in.

jlaura:
Looks good.  Single comment and then good to merge.

jlaura:
:+1:

PR #72, opened 2016-03-22
Pep8 modifications 03/22/16

acpaquette:
Both the matcher and the graph directories are pep8 complient ignoring line length, and imports not being at top of file.


acpaquette:
Pull request update failed for some reason. The nosetests should be passing but the PR has taken my most recent commit.

PR #73, opened 2016-03-22
Pep8 modifications 03/22/16

acpaquette:
Both the matcher and the graph directories are pep8 complient ignoring line length, and imports not being at top of file.


acpaquette:
PR isn't updating with the pep8 branch, not sure why.

PR #75, opened 2016-03-24
Updated image_match to work with a config file

acpaquette:
Basic implementation of a config file for image_match. Easily editable down the road.


jlaura:
@Kelvinrr Please do the code review on this, add comments inline, and then let @kree or @jlaura know so we can do a super quick double check.

bin/image_match.py
```diff
@@ -47,10 +54,11 @@ def match_images(args):
...
 
     filelist = cg.to_filelist()
-    write_filelist(filelist, args.output_file + '.lis')
+    write_filelist(filelist, config['match_image']['outputfile_path'] + args.output_file + '.lis')
 
-    to_isis(args.output_file + '.net', cnet, mode='wb', targetname='Moon')
+    to_isis(config['match_image']['outputfile_path'] + args.output_file + '.net', cnet, mode='wb', targetname='Moon')
 
 if __name__ == '__main__':
+    config = read_config('/home/acpaquette/autocnet/.image_match_config.yml')
```
Kelvinrr: Did you intentionally put your desktop's config file path here?

bin/image_match.py
```diff
@@ -1,31 +1,38 @@
...
+from autocnet.fileio.io_yaml import read_yaml
 
-# parses command line arguments into a single args variable
+# reads in config file
+def read_config(yaml_file):
+    config_dict = read_yaml(yaml_file)
+
+    return config_dict
+
+# parses command line arguments into a single args list
```
Kelvinrr: Probably not a big deal, but these inline comments might be better as docstrings. Alternatively, you could just remove them since they don't say anything you can't get from reading the signature.

.image_match_config.yml
```diff
@@ -0,0 +1,5 @@
+match_image: 
+  inputfile_path: /home/acpaquette/autocnet/autocnet/examples/Apollo15/
+  basepath: /home/acpaquette/Desktop/
+  outputfile_path: /home/acpaquette/autocnet/autocnet/examples/Apollo15/
```
Kelvinrr: Did you mean to nest these? It seems to me that it unnecessarily adds some verbosity when trying to access them. e.g `config['match_image'][basepath]` vs `config[basepath]`. Feel free to correct me if I am misunderstanding something.

Kelvinrr:
It seem good to me, what I found are minor things. @kree or @jlaura can catch anything i might have missed and/or correct any of my comments.

acpaquette:
It just depends on how we want to organize the config file. We will be adding other fields for the methods that are used in the script. There are many ways to do it and we can talk about it at some point. I will probably remove the majority of the path inserts since it is specific to my system.

jlaura:
Look at the function find_in_dict in utils.utils. This will make working
with a nested dict much easier.  (And yes, a nested dict is they way to
organize this.)

bin/image_match.py
```diff
@@ -1,31 +1,36 @@
...
 
     # Matches the images in the input file using various candidate graph methods
     # produces two files usable in isis
     try:
-        cg = CandidateGraph.from_adjacency(args.input_file, basepath=args.basepath)
+        cg = CandidateGraph.from_adjacency(config['inputfile_path'] +
+                                           args.input_file, basepath=config['basepath'])
     except:
-        cg = CandidateGraph.from_filelist(args.input_file)
+        cg = CandidateGraph.from_filelist(config['inputfile_path'] + args.input_file)
```
Kelvinrr: Shouldn't this be config_dict?

jlaura:
@acpaquette Can you check the comment from @Kelvinrr and address if necessary so that we can get this merged?

PR #76, opened 2016-03-29
Image match with working Config file

acpaquette:
Basic implementation of config file with image_match


.image_match_config.yml
```diff
@@ -0,0 +1,3 @@
+basepath:
+inputfile_path:
+outputfile_path:
```
jlaura: Why is this a dot prepended file?

bin/image_match.py
```diff
@@ -5,27 +5,31 @@
 sys.path.insert(0, os.path.abspath('../autocnet'))
```
jlaura: This line should be removed, it is an artifact of development and not needed in deployment.

bin/image_match.py
```diff
@@ -5,27 +5,31 @@
...
 from autocnet.graph.network import CandidateGraph
-from autocnet.fileio.io_controlnetwork import to_isis
-from autocnet.fileio.io_controlnetwork import write_filelist
+from autocnet.fileio.io_controlnetwork import to_isis, write_filelist
+from autocnet.fileio.io_yaml import read_yaml
+
+def read_config(yaml_file):
+    config_dict = read_yaml(yaml_file)
+
+    return config_dict
```
jlaura: This is a function wrapping a function that is not a closure....why?

PR #79, opened 2016-04-01
fixed mst node comparison issue

Kelvinrr:
This concerns a problem where the second mst graph in `TestGraphMasks`might have it's nodes slightly out of order after mst computation on rare occasions. This causes the node list comparison to fail even though node order is inconsequential.


jlaura:
:+1:

PR #86, opened 2016-04-08
Fixes subgraphs from node lists and add subgraph extract from edge lists

jlaura:
@Kelvinrr this one is for you.  I will be in the air until this afternoon, so I will not be able to make review changes until tonight.  If the review is minor stuff, go ahead comment and merge.  This way you can use the code.  If it is major hold off.  Wither way, I'll address comments after the flight.


Kelvinrr:
looks good.

PR #87, opened 2016-04-08
Create sub-graph from matched images

Kelvinrr:
Something to think about: would it be useful if we add a more generalized method that creates a sub-graph given nodes/edges that meet some condition (e.g. have matches)?


autocnet/graph/network.py
```diff
@@ -698,3 +698,21 @@ def create_node_subgraph(self, nodes):
 
         H.graph = self.graph
         return H
+
+    def subgraph_from_matches(self):
+        """
+        Returns a sub-graph where all images have matches.
```
jlaura: where all edges have matches.

autocnet/graph/network.py
```diff
@@ -698,3 +698,21 @@ def create_node_subgraph(self, nodes):
...
+        Returns
+        -------
+        : Object
+          A networkX graph object
+        """
+
+        # get all edges that have matches
+        matches = ((u, v) for u, v in self.edges_iter()
+                   if hasattr(self.get_edge_data(u, v), 'matches') and
+                   not self.edge[u][v].matches.empty)
```
jlaura: Can you use [filter](https://docs.python.org/3/library/functions.html#filter) here and `self.edges_iter(data=True)`.

Kelvinrr: It could use `filter`, but from what I know, I would have to use a lambda and I imagine that would be worse performance-wise and won't necessarily be more readable.

PR #89, opened 2016-04-13
simple node/edge filter

Kelvinrr:
Simple methods that filter through nodes or edges using some function and returns a sub-graph from matches. Mimics python's `filter` function.


jlaura:
:+1:

PR #90, opened 2016-04-15
Fix for GeoDataset.foorprint failing on read-only cubes 

kree:
It doesn't look like we actually need write permission, so I changed r -> r+


Kelvinrr:
That little plus was a pain.

PR #91, opened 2016-04-15
removed redundant filter_from_matches method

Kelvinrr:
With the general case available, I think it is safe to say this isn't needed anymore.


jlaura:
Could a syntax sugar be implemented / retained?

Kelvinrr:
I guess we can leave this alone for syntactic sugar. I personally can't think of an alternative that would be cleaner.

PR #92, opened 2016-04-15
Use slopes to refine fundamental matrix

kree:
References JIRA #116


jlaura:
@kree From the test error, it looks like the action stack test might just need an update because refine is now being called twice?

PR #95, opened 2016-04-18
Added robustness to missing footprints and more meaningful warnings 

Kelvinrr:


kree:
This all looks completely reasonable to me. Would you be willing to give it a quick look-over, too, @jlaura?

jlaura:
@Kelvinrr Is handle used in any of the notebook visualizations that we are regularly using?  If not, good to merge, if so let's get those updated.

Kelvinrr:
@jlaura Didn't think of that. After a quick `grep`ing of the notebook directory, looks like some do reference it. I'll look through them when I have a moment. 

```bash
./ISIS Serial Numbers.ipynb
./Sprint1Demo.ipynb
./UnderstandingAutoCNetInputAndOutputStructures.ipynb
./VisualizingOutputsExample.ipynb
```

jlaura:
@Kelvinrr These look like they are candidates for removal.  @kree what do you think?

kree:
@jlaura Yes, I think those can all be removed.

PR #96, opened 2016-04-19
Added syntactic sugar to network.py

acpaquette:
Allows the user to access the individual methods easily.


jlaura:
:+1:

autocnet/graph/network.py
```diff
@@ -395,6 +395,42 @@ def apply_func_to_edges(self, function, *args, graph_mask_keys=[], **kwargs):
...
+
+    def ratio_checks(self, *args, **kwargs):
+        '''
+        Apply a symmetry check to all edges in the graph
+        '''
+        self.apply_func_to_edges('ratio_check', *args, **kwargs)
+
+    def compute_homographies(self, *args, **kwargs):
+        '''
+        Apply a symmetry check to all edges in the graph
```
Kelvinrr: you might want to update all these doc-strings, they all say symmetry check. Might as well include the parameters and return values. It would probably be good to have some way users can infer the parameters these methods take since they all take fairly different params.

jlaura: Instead of copying the parameters, how about using the See Also syntax:
https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt
 (Number 9)

On Tue, Apr 19, 2016 at 11:15 AM, Kelvin Rodriguez <notifications@github.com

acpaquette: So for ratio checks would the see also section look like this?

See Also
-------------
outlier_detector.DistanceRatio.compute

Since that is where the **kwargs are ultimately being funneled into. I am
just not sure if I have the naming conventions right for the "path". I
believe the DistanceRatio.compute is correct but I don't know if I should
add on anything else.

jlaura: Give that a try.  Then cd to docs, and do a make clean, make install.  The _build directory will have an index.html file that is a local version of the docs.  Give a few PATHs a try to see if one works.

acpaquette:
The See Also links are working properly, figured out that you need the whole "path" for the link to work. Need documentation for a few of the methods used i.e. compute_fundamental_matrix in edge.py has no documentation but the only thing that needs documentation is the ability to pass in clean keys since everything else is passed to compute_fundamental_matrix in outlier_detector.py

Kelvinrr:
I believe changing the functional tests back to using these methods should increase the coverage.

jlaura:
:+1:

acpaquette:
I changed the apply func methods in one of the funcitonal tests to the new methods which increased coverage but not by much.

Kelvinrr:
Still missing coverage for `ratio_check`, `symmetry_check`, `compute_fundamental_matrices` and `suppress`. The two_image_test seems to cover some of these by iterating through edges instead of using the network methods. Making those modifications should bring the coverage up further. 

e.g. This:
```Python
for source, destination, edge in cg.edges_iter(data=True):
    # Perform the symmetry check
    edge.symmetry_check()
    self.assertIn(edge.masks['symmetry'].sum(), range(400, 600))

    # Perform the symmetry check
    edge.ratio_check(clean_keys=['symmetry'])
    self.assertIn(edge.masks['ratio'].sum(), range(30, 100))
```
can be rewritten as something like:

```Python
# Perform the symmetry check
cg.symmetry_checks()
# Perform the ratio check
cg.ratio_checks()

for source, destination, edge in cg.edges_iter(data=True):
    self.assertIn(edge.masks['symmetry'].sum(), range(400, 600))
    self.assertIn(edge.masks['ratio'].sum(), range(30, 100))
```

(Obviously there are some things missing in the latter code snippet but you get the idea)

Do the same for `compute_fundamental_matrices` and then write a test for `suppress`.

jlaura:
@acpaquette This PR is going to need to get aimed at the dev branch and checked for merging.  Something for Tuesday.

PR #94, opened 2016-04-18
Fixes size rounding issues and alters how the graph backend stores no…

jlaura:
…des and edges.

PR #97, opened 2016-04-19
Updated keypoints method in Vis

acpaquette:
It seems like someone already did the majority of these but missed one in the process. Either that or I am looking at the wrong file.


Kelvinrr:
Yeah, Jay got most of them in #94

autocnet/vis/graph_view.py
```diff
@@ -94,7 +94,7 @@ def plot_node(node, ax=None, clean_keys=[], **kwargs):
     keypoints = node.get_keypoints()
     if clean_keys:
         matches, mask = node._clean(clean_keys)
-        keypoints = node.keypoints[mask]
+        keypoints = node.get_keypoints[mask]
```
Kelvinrr: Isn't the proper use here is `get_kepoints(index = mask)`?

jlaura: `get_keypoints[mask]` should be failing.

Kelvinrr: yeah, I think `get_keypoints()[mask]` seems to be the way to go.

acpaquette:
Corrected the syntax error in the get_keypoints method.

Kelvinrr:
👍

PR #100, opened 2016-04-20
added robustness to poor footprints

Kelvinrr:
This should give added robustness when dealing with polys that fail both the PVL and lat/lon check. Also added robustness when dealing with polys that pass the checks but still generate an invalid polygon. 

Calls to `match_features()` should no longer fail in the presence of isolated vertices.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -8,6 +8,7 @@ import pandas as pd
 import warnings
 
 from autocnet.fileio.io_gdal import GeoDataset
+from autocnet.fileio import io_utils
 from autocnet.fileio import io_hdf
 from autocnet.control.control import C
 from autocnet.fileio import io_json
@@ -108,11 +109,8 @@ class CandidateGraph(nx.Graph):
         : object
           A Network graph object
         """
-        if not isinstance(filelist, list):
-            with open(filelist, 'r') as f:
-                filelist = f.readlines()
-                filelist = map(str.rstrip, filelist)
-                filelist = filter(None, filelist)
+        if isinstance(filelist, str):
+            filelist = io_utils.file_to_list(filelist)
 
         # TODO: Reject unsupported file formats + work with more file formats
         if basepath:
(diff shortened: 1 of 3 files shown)


jlaura:
@Kelvinrr Please add a commit that gets the coverage passing.

Kelvinrr:
I left a comment on gitter about it. But the gist is that the only way I can imagine getting coverage back up involves adding cubes (or something with a footprint that can be used with `from_file_list` that has both passing and failing cases) to the repo under examples.

autocnet/graph/network.py
```diff
@@ -121,23 +122,29 @@ def from_filelist(cls, filelist, basepath=None):
```
jlaura: The above four lines can be extracted into a separate function that returns an iterable of datasets (this could even be a generator).  Then...

autocnet/graph/network.py
```diff
@@ -121,23 +122,29 @@ def from_filelist(cls, filelist, basepath=None):
...
+        valid_datasets = []
 
-        for i, j in itertools.permutations(datasets,2):
-            if not i.file_name in adjacency_dict.keys():
-                adjacency_dict[i.file_name] = []
-            if not j.file_name in adjacency_dict.keys():
-                adjacency_dict[j.file_name] = []
+        # filter bad/malformed footprints
+        for i in datasets:
+            adjacency_dict[i.file_name] = []
```
jlaura: Utilize the `Mock` module to test this section by passing in a list of Mocked geo datasets.  The only thing that each Mock object needs to return is a footprint.  Make one of those bad.

jlaura:
https://docs.python.org/3/library/unittest.mock.html

jlaura:
I added some thoughts inline.

Kelvinrr:
Thanks for for info!

jlaura:
Looks good.  Nice job!  Perhaps you can chat for a couple at the standup today and let everyone know how / why Mocking worked here?

PR #110, opened 2016-04-26
CandidateGraph.Size() fix

Kelvinrr:
This will hopefully fix the CandidateGragh.number_of_edges() issue mentioned on Gitter. Plus, make it marginally more efficient.

(I don't know why it is including all these extra commits)

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1,5 +1,4 @@
 import itertools
-import math
 import os
 import warnings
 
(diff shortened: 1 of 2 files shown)


jlaura:
Checkout the gitter conversation on the commits.  I *think* that you pulled commits from master.  Those need to get removed.  Once those are out, this looks good and can get merged in to dev.

Kelvinrr:
That is exactly what happened, muscle memory just wants to write upstream/origin master

PR #112, opened 2016-04-28
Made both the graph and matcher directories pep8 complaint

acpaquette:
Reverted the changes to the line lengths. I forgot to ignore line length in the autopep8 command, along with the flake8 command.


jlaura:
:+1:

PR #115, opened 2016-05-12
Adds better F support, concurrency support, guided matching, camera estimation (E)

jlaura:
Anyone can merge this.  I can explain what is going on here and why if anyone wants.  Follow on to this will be guided matching to support added control depth.


autocnet/graph/edge.py
```diff
@@ -148,6 +207,12 @@ def compute_fundamental_matrix(self, clean_keys=[], **kwargs):
         # Set the initial state of the fundamental mask in the masks
         self.masks = ('fundamental', mask)
 
+    def add_putative_matches(self):
```
jlaura: Remove def.

autocnet/graph/edge.py
```diff
@@ -178,17 +243,13 @@ def compute_homography(self, method='ransac', clean_keys=[], pid=None, **kwargs)
...
-        transformation_matrix, ransac_mask = od.compute_homography(s_keypoints.values,
-                                                                   d_keypoints.values,
-                                                                   **kwargs)
+        self.homography = Homography(np.zeros((3,3)), index=self.masks.index)
+        self.homography.compute(s_keypoints.values,
+                                d_keypoints.values)
 
         # Convert the truncated RANSAC mask back into a full length mask
-        mask[mask] = ransac_mask.ravel()
+        mask[mask] = self.homography.mask.ravel()
```
jlaura: yank ravel, Series are flat.

autocnet/fileio/io_gdal.py
```diff
@@ -386,9 +400,9 @@ def pixel_to_latlon(self, x, y):
         
         """
         try:
-            geotransform = self.geotransform
-            x = geotransform[0] + (x * geotransform[1]) + (y * geotransform[2])
-            y = geotransform[3] + (x * geotransform[4]) + (y * geotransform[5])
+            #geotransform = self.geotransform
+            #x = geotransform[0] + (x * geotransform[1]) + (y * geotransform[2])
+            #y = geotransform[3] + (x * geotransform[4]) + (y * geotransform[5])
             lon, lat, _ = self.coordinate_transformation.TransformPoint(x, y)
```
acpaquette: This needs to be changed from lon, lat to lat, lon. Otherwise it inverts the output point.

autocnet/fileio/io_gdal.py
```diff
@@ -203,20 +209,21 @@ def geospatial_coordinate_system(self):
...
-            self._latlon_extent = [(lowerlat, lowerlon), (upperlat, upperlon)]
+                self._footprint = [(upperlat, lowerlon),
+                                   (lowerlat, lowerlon),
+                                   (lowerlat, upperlon),
+                                   (upperlat, upperlon)]
+            else:
+                xy_corners = self.xy_corners
+                self._latlon_extent = []
+                self.coordinate_transformation
+                for x, y in xy_corners:
```
acpaquette: Self.coordinates_transform from what I can tell is doing nothing, the latlon_extent also assumes it is being given pixel coordinates. I imagine that no matter what type the image is we should still be able to return a latlon or pixel extent.

acpaquette:
In pixel_to_latlon, change line 392 from lon, lat, .... to lat, lon. Also are you planning on doing anything with the Maximum Likelihood Estimation method in transformations.py before merging?

autocnet/graph/edge.py
```diff
@@ -6,11 +6,12 @@
 from pysal.cg.shapes import Polygon
 
 from autocnet.cg.cg import convex_hull_ratio
-from autocnet.cg.cg import overlapping_polygon_area
+from autocnet.cg import cg
```
Kelvinrr: duplicate import?

autocnet/camera/camera.py
```diff
@@ -0,0 +1,165 @@
...
+    residuals : ndarray
+                (n, 1) residuals for each correspondence
+
+    cumulative_error : float
+                       sum of the residuals
+
+
+    """
+    if p1.shape != (3,4):
+        p1 = p1.reshape(3,4)
```
Kelvinrr: Why force a reshape instead of warning/error?

jlaura: This method can be called by an optimization routine that requires a 1D array, but the camera (and associated matrix multiplication) is in 2D.  This function can get hit hundreds, if not thousands of times, when optimizing the fundamental matrix.  Therefore, the reshape is intentional and should be transparent to the user.

jlaura:
@acpaquette I do not plan to add the MLE stuff in in this PR.  Too big already and the poor optimization performance is not inspiring me to keep pushing ahead on cleaning / testing that code.  Is the comment re: 391 in io_gdal.py still valid?  The coordination transformation appears to return lon, lat and not lat, lon.  Java docs: http://gdal.org/java/org/gdal/osr/CoordinateTransformation.html#TransformPoint(double[],%20double,%20double)

So, double x, double y or lon, lat.

PR #116, opened 2016-05-18
Ciratefi

Kelvinrr:
This is a notebook that contains the first filter for Ciratefi.


jlaura:
:+1: Looking forward to seeing phase II - I think it should be a bunch easier!

PR #117, opened 2016-05-18
Update to the Spatial Suppression function in outlier detector and various tests

acpaquette:


autocnet/matcher/tests/test_outlier_detector.py
```diff
@@ -91,11 +91,44 @@ def test_suppress_non_optimal(self):
...
 
         with warnings.catch_warnings(record=True) as w:
             self.suppression_obj.k = 101
             self.suppression_obj.suppress()
             self.assertEqual(len(w), 1)
             self.assertTrue(issubclass(w[0].category, UserWarning))
 
+class testSuppressionRanges(unittest.TestCase):
+
+    def setUp(self):
```
jlaura: Here you could do:

```python
@classmethod
def setUpClass(cls):
    cls.r = np.random.RandomState(12345)
```

This method would then get called once, when the test class is initially instantiated.  As it currently stands `self.r` is instantiated once for every test method.  Not something to change, but something to be aware of in case you ever need to dig into tests that are running slowly.

jlaura:
:+1: Checkout the inline comment.  If you want to update that line, go for it.  If not, let me know and anyone can merge this.

PR #118, opened 2016-05-19
Added overlap method to edge

acpaquette:
Gathers the overlap data of the two images, and calculates the area being overlapped and the percentage of the total area being overlapped. Does not use real world measurements and I am not sure if we want super accurate measurements. Also still needs to be tested not sure how to do so without using cub data.


autocnet/graph/edge.py
```diff
@@ -80,6 +87,21 @@ def masks(self, v):
     def health(self):
         return self._health.health
 
+    def overlap(self):
```
jlaura: Since line 49 adds the weight attr, the `if hasattr(self, 'weight')` is redundant.  The class will always have this attribute (unless a user manually removed it, but we do not need to protect against that...).

autocnet/graph/edge.py
```diff
@@ -80,6 +87,21 @@ def masks(self, v):
...
+        if hasattr(self, 'weight'):
+            poly1 = self.source.geodata.footprint
+            poly2 = self.destination.geodata.footprint
+
+            intersection = poly1.Intersection(poly2)
+            self._overlap_area = intersection.GetArea()
+
+            total_area = poly1.GetArea()
+            intersection_area = intersection.GetArea()
+            self._overlap_percn = (intersection_area/total_area)*100
```
jlaura:
```math
[z / (x + y - z)] * 100, where z is the intersection area, x is poly1_area, and y is poly2_area
```

Quick logic check (please punch a hole!):

- If two polygons are exactly aligned, with area 2, this will result in 100%.
- If x = 4, y = 9, and z = 2, we have ~18% overlap.
- If z is increased to 4, e.g. x is entirely covered by y, the result is ~44%.

acpaquette:
Pushed to update the PR since it has been inactive for a few days now. Everything should be good to go except for a few tests and documentation.

autocnet/cg/cg.py
```diff
@@ -1,5 +1,8 @@
 import json
 import ogr
+import pandas as pd
+
+from autocnet.fileio import io_gdal
```
jlaura: Unused import?

autocnet/cg/tests/test_cg.py
```diff
@@ -19,3 +20,15 @@ def test_area_single(self):
...
+        wkt1 = "POLYGON ((0 40, 40 40, 40 0, 0 0, 0 40))"
+        wkt2 = "POLYGON ((20 60, 60 60, 60 20, 20 20, 20 60))"
+
+        poly1 = ogr.CreateGeometryFromWkt(wkt1)
+        poly2 = ogr.CreateGeometryFromWkt(wkt2)
+
+        info = cg.two_poly_overlap(poly1, poly2)
+
+        self.assertEqual(info[1], 400)
+        self.assertEqual(info[0], 14.285714285714285)
```
jlaura: AssertAlmostEqual might be safer here.  To say 6 places?

autocnet/graph/edge.py
```diff
@@ -1,5 +1,7 @@
 import warnings
 from collections import MutableMapping
+import json
+import ogr
```
jlaura: Are either of these used? (ogr, json)

autocnet/graph/network.py
```diff
@@ -476,6 +477,16 @@ def suppress(self, *args, **kwargs):
...
 
+    def overlap(self, *args, **kwargs):
+        '''
+        Compute the percentage and area coverage of two images
+
+        See Also
+        --------
+        autocnet.cg.cg.two_image_overlap
+        '''
+        self.apply_func_to_edges('overlap')
```
jlaura: Make sure to pass *args and **kwargs if they can be.  Otherwise, you can remove them from the signature in 480.

autocnet/graph/node.py
```diff
@@ -1,11 +1,15 @@
 from collections import MutableMapping
 import os
 import warnings
+import json
+import ogr
 
 import numpy as np
 import pandas as pd
 from scipy.misc import bytescale
+from scipy.spatial import ConvexHull
```
jlaura: Unused?

autocnet/graph/node.py
```diff
@@ -1,11 +1,15 @@
 from collections import MutableMapping
 import os
 import warnings
+import json
```
jlaura: Unused?

autocnet/graph/node.py
```diff
@@ -1,11 +1,15 @@
...
+import ogr
```
jlaura: Unused?

functional_tests/test_two_image.py
```diff
@@ -105,3 +110,28 @@ def tearDown(self):
             os.remove('TestTwoImageMatching.net')
             os.remove('fromlist.lis')
         except: pass
+
+    def test_edge_overlap(self):
```
jlaura: Should this be here or in the edge test?  This looks like it is testing edge functionality, as opposed to being a true, pipeline style functional test.

jlaura:
Check the one remaining note and this should be good to go.  Re: line 480 in network.py.

jlaura:
:1: Nice job.

PR #119, opened 2016-05-19
Ciratefi: Rafi + cleanup

Kelvinrr:
Added Rafi to the notebook. Image shows normalized cross correlation for first grade candidate pixels only. 

Edit: Added image for both the original Cifi pass and Tefi on first grade candidate pixels.


jlaura:
:+1:  This looks really good!  Going to crank out the final step?

Kelvinrr:
Yeah, I imagine it would be done before I leave for today.

PR #120, opened 2016-05-20
Notebook continuous integration

kree:
This includes the runipy script which will run all notebooks in autocnet for each build and the deletion of many obsolete notebooks. 

The remaining notebooks (which @jlaura) asked me to keep are:
ISIS Serial Numbers
Testing Slopes
Suppression Vs. Disk Covering
Tutorial with Visualization 
Kaguya Spectral Profilier

I had to modify many of these in temporary ways (commenting things out) to force this build to pass, and someone at some point should look into actually fixing them. 

Also -- I tried to squash my commits and clearly screwed something up. I ended up with a detached HEAD and tried to fix it. If anyone is willing to help me figure out what went wrong, that would be great.


kree:
This is failing right now because the new Ceritefi notebook is failing. @Kelvinrr could you take a quick look at the failure and see if it's something you would expect?

Kelvinrr:
Yeah, it is from forgetting to import warnings. I fixed it PR #119

jlaura:
Just merged #119.  I will rerun these tests in 5 or so, and then this should be good to go.  Nice job getting the notebooks testing! 

We can open a ticket in JIRA to get the failures up to snuff.  If you have a moment, can you toss a one liner, how to test the notebooks locally into [here](https://github.com/USGS-Astrogeology/autocnet/wiki/Testing)?  Thanks!

kree:
@jlaura -- I added the information about how to test the notebooks.

PR #121, opened 2016-05-23
Ciratefi: Tefi

Kelvinrr:
Adds the last filter to Ciratefi. It is still a little janky (the way I am picking the best fit scale/rotation is broken, trivial fix) but it still works to show how the algorithm functions.


jlaura:
:+1: For the full on implementation we will want to make sure that the up sampling is working as well.

PR #123, opened 2016-06-02
Added coverage method to the edge.

acpaquette:
Needs serious refactoring which I will probably get done today then work on point clustering.


autocnet/graph/edge.py
```diff
@@ -424,3 +425,60 @@ def overlap(self):
...
 
+    def _construct_json_serial(self, coords):
+        hull = cg.convex_hull(coords)
+        poly_points = [hull.points[i] for i in hull.vertices]
+        coordinates = [[i, j] for [i, j] in poly_points]
+        coordinates.append((poly_points[0][0], poly_points[0][1]))
+        return coordinates
+
+    def coverage(self, image='source'):
+        mask = self.masks.all(axis = 1)
```
Kelvinrr: Might want to check if matches have been computed first.

Something like:

```
 if not hasattr(self, 'matches'):
            raise AttributeError('No matches have been computed for this edge.')

```

autocnet/graph/edge.py
```diff
@@ -424,3 +425,60 @@ def overlap(self):
         self.weight['overlap_area'] = overlapinfo[1]
         self.weight['overlap_percn'] = overlapinfo[0]
 
+    def _construct_json_serial(self, coords):
+        hull = cg.convex_hull(coords)
+        poly_points = [hull.points[i] for i in hull.vertices]
+        coordinates = [[i, j] for [i, j] in poly_points]
+        coordinates.append((poly_points[0][0], poly_points[0][1]))
+        return coordinates
```
Kelvinrr: As per our discussion, maybe simplify this and put it into `coverage`

autocnet/graph/edge.py
```diff
@@ -424,3 +425,60 @@ def overlap(self):
...
+
+    def coverage(self, image='source'):
```
Kelvinrr: Doc-String

autocnet/graph/edge.py
```diff
@@ -424,3 +425,60 @@ def overlap(self):
...
+            node = self.destination
+
+        convex_coordinates = self._construct_json_serial(image_covered)
+
+        # Convert the convex hull pixel coordinates to latlon
+        # coordinates
+        convex_points = []
+        for i, j in convex_coordinates:
+            point = node.geodata.pixel_to_latlon(i, j)
+            convex_points.append(point)
```
Kelvinrr: This can be simplified via list comprehension.

autocnet/graph/edge.py
```diff
@@ -424,3 +425,60 @@ def overlap(self):
...
+        convex_coords = [[i, j] for i, j in convex_points]
```
Kelvinrr: i, j should be something more like x, y

autocnet/graph/edge.py
```diff
@@ -424,3 +425,60 @@ def overlap(self):
...
+        return coordinates
+
+    def coverage(self, image='source'):
+        mask = self.masks.all(axis = 1)
+        matches = self.matches[mask]
+        source_array = self.source.get_keypoint_coordinates(index=matches['source_idx'])
+        destination_array = self.destination.get_keypoint_coordinates(index=matches['source_idx'])
+
+        source_points = np.array(source_array)
+        destination_points = np.array(destination_array)
```
Kelvinrr: You might want to use `np.asarray` as that guarantees a copy won't be made. Alternatively, you can keep `np.array` and set the optional `copy` parameter to `False` (as the default is `True`).

autocnet/graph/edge.py
```diff
@@ -424,3 +425,60 @@ def overlap(self):
...
+
+        source_verts = self.source.geodata.latlon_corners
+        destination_verts = self.destination.geodata.latlon_corners
+
+        source_coordinates = self._construct_json_serial(source_verts)
+
+        destination_coordinates = self._construct_json_serial(destination_verts)
```
Kelvinrr: Calling a convex hull algorithm on what are suppose to be corners is redundant.

autocnet/graph/edge.py
```diff
@@ -424,3 +425,60 @@ def overlap(self):
...
+
+        if image == 'source':
+            image_covered = source_points
+            node = self.source
+        else:
+            image_covered = destination_points
+            node = self.destination
+
```
Kelvinrr: I do not know the full requirements for this PR, but I imagine we would want to consolidate the (s,d) and (d,s) convex hulls. Probably something @jlaura should comment on.

autocnet/graph/edge.py
```diff
@@ -424,3 +425,60 @@ def overlap(self):
...
+        # coordinates
+        convex_points = []
+        for i, j in convex_coordinates:
+            point = node.geodata.pixel_to_latlon(i, j)
+            convex_points.append(point)
+        convex_coords = [[i, j] for i, j in convex_points]
+
+        source_geom = {"type": "Polygon", "coordinates": [source_coordinates]}
+        destination_geom = {"type": "Polygon", "coordinates": [destination_coordinates]}
+        convex_geom = {"type": "Polygon", "coordinates": [convex_coords]}
```
Kelvinrr: There should be a way to convert an array directly to a polygon without awkward conversions to json.

autocnet/graph/edge.py
```diff
@@ -423,3 +426,66 @@ def overlap(self):
...
+        -------
+        total_overlap_percentage : float
+                                   returns the overlap area
+                                   covered by the keypoints
+        """
+        if self.matches is None:
+            raise AttributeError('Edge needs to have features extracted and matched')
+            return
+        mask = self.masks.all(axis = 1)
+        matches = self.matches[mask]
```
jlaura: What about just using the `_clean` method on the edge?  `matches, mask = self._clean(clean_keys)`.  This requires that the coverage method take a `clean_keys` kwarg that defaults to `[]`.

autocnet/graph/edge.py
```diff
@@ -423,3 +426,66 @@ def overlap(self):
...
+        source_array = self.source.get_keypoint_coordinates(index=matches['source_idx'])
+        destination_array = self.destination.get_keypoint_coordinates(index=matches['destination_idx'])
+
+        source_points = np.array(source_array)
+        destination_points = np.array(destination_array)
+
```
jlaura: You can remove 454 and 454 by changing 451/452 to: `source_array = self.source.get_keypoint_coordinates(index=matches['source_idx']).values`  and `.values` on 452.

autocnet/graph/edge.py
```diff
@@ -423,3 +426,66 @@ def overlap(self):
...
+            image_covered = source_points
+            node = self.source
+        else:
+            image_covered = destination_points
+            node = self.destination
+
+        # pixel space
+        convex_hull = cg.convex_hull(image_covered)
+
+        convex_points = [node.geodata.pixel_to_latlon(row[0], row[1]) for row in convex_hull.points[convex_hull.vertices]]
```
jlaura: Here a change to `pixel_to_latlon` is probably in order to avoid the list comprehension completely.  For now, keep it in, but I want to get this comment documented: `pixel_to_latlon` should be a multi-method (single dispatch).  If the arg is an ndarray of (x,y) pixels, then the dispatch is to a function that reshapes to (-1, 2), applies the geotransform, and then calls the `TransformPoints` method.  If the arg is an (x,y) tuple then the current method is completely appropriate.  This is accomplished using: https://docs.python.org/3/library/functools.html#functools.singledispatch

autocnet/graph/edge.py
```diff
@@ -423,3 +426,66 @@ def overlap(self):
...
+        # Convert the convex hull pixel coordinates to latlon
+        # coordinates
+        source_geom = utils.array_to_geom(destination_coords)
+        destination_geom = utils.array_to_geom(source_coords)
+        convex_geom = utils.array_to_geom(convex_coords)
+
+        source_poly = ogr.CreateGeometryFromJson(json.dumps(source_geom))
+        destination_poly = ogr.CreateGeometryFromJson(json.dumps(destination_geom))
+        convex_poly = ogr.CreateGeometryFromJson(json.dumps(convex_geom))
+
```
jlaura: Could we move lines 475 to 481 off of the edge and into `cg`?  It seems that a function there could take a list of a coordinates and return an ogr geometry object.  This means that edge never needs to know about `ogr` or `json`.

autocnet/graph/edge.py
```diff
@@ -423,3 +426,66 @@ def overlap(self):
...
+        image_intersection = source_poly.Intersection(destination_poly)
+        image_intersection_area = image_intersection.GetArea()
+
+        hull_overlap_poly = image_intersection.Intersection(convex_poly)
+        hull_overlap_area = hull_overlap_poly.GetArea()
```
jlaura: Here too, could 483 - 487 be moved into a function in `cg`that returns area of an arbitrary polygon?

autocnet/graph/edge.py
jlaura: I am not 100% that I am following the logic here.  Does it matter if the pixel coordinates come from the source or the destination?  Either way, the overlap is being projected to lat/lon using the geotransform from one of those two images.  The overlap is also computed in lat/lon space.  So we should be invariant to which geotransform we use?

autocnet/graph/tests/test_edge.py
```diff
@@ -72,3 +75,58 @@ def test_edge_overlap(self):
...
+            return vals[(i, j)]
+
+        e.source.geodata.pixel_to_latlon = MagicMock(side_effect = pixel_to_latlon)
+        e.destination.geodata.pixel_to_latlon = MagicMock(side_effect = pixel_to_latlon)
+
+        e.matches = matches_df
+
+        self.assertRaises(AttributeError, cg.edge[0][1].coverage)
+        self.assertEqual(e.coverage(image='destination'), 0.3)
+        self.assertEqual(e.coverage(image='source'), 0.3)
```
jlaura: Nice use of Mocking here!  This is a great isolation of just the components we want tested.

autocnet/utils/utils.py
```diff
@@ -305,3 +305,29 @@ def cartesian(arrays, out=None):
         out[:, n] = arrays[n][ix[:, n]]
 
     return out
+
+
+def array_to_geom(array):
```
jlaura: I might add a `kwarg` here `geom='Polygon'`, as this function is not an arbitrary array to geom, but assumes the geom is a polygon.  You could take the `geom` arg and support a polyline without issue.  This ignores the complexity of multi-geoms, etc., but we are fine with that for now.

jlaura:
Comments in.  Let me know if anything does not make sense.

autocnet/graph/edge.py
acpaquette: It was due to variance between the convex hulls of the source and destination. I discovered that I was passing destination to a source variable vis versa throwing off the area results. Now that everything is being passed correctly there is at most about a 2% difference of area covered between images. Should I do max, min or an average of the two or should I just do with one of the nodes(source or destination)?

autocnet/graph/edge.py
acpaquette: It was due to variance between the convex hulls of the source and destination. I discovered that I was passing destination to a source variable vis versa throwing off the area results. Now that everything is being passed correctly there is at most about a 2% difference of area covered between images. Should I do max, min or an average of the two or should I just do with one of the nodes(source or destination)?

jlaura: I would just pick on.  Maybe make a note in the function documentation.  Nice job tracking this down.

autocnet/graph/edge.py
acpaquette: Is there a way to just grab the clean keys in string format to pipe into _clean? I understand that this would be if we only wanted certain clean keys, but if there isn't a very easy/efficient way to do that I think just grabbing all of the clean_keys that have been calculated is fine.

autocnet/graph/edge.py
acpaquette: Having a little trouble implementing this:
`    @singledispatch
    def pixel_to_latlon(self, x, y):`

`    @pixel_to_latlon.register(list)
    def _(self, arg):
        latlon_points = [self.geodata.pixel_to_latlon(row[0], row[1]) for row in arg]
        return latlon_points`

It doesn't seem to be picking up the dispatch into the list overload

Kelvinrr:
The single dispatch is a bit more complicated when dealing with methods. We can do probably handle that in a different PR at a different time.

jlaura:
@Kelvinrr Absolutely.  Here is a nice example for methods.  Nice to see functools wrapper being put to good use.

http://stackoverflow.com/questions/24601722/how-can-i-use-functools-singledispatch-with-instance-methods

PR #129, opened 2016-09-21
Generalize the feature extractor

jlaura:
for use outside of autocnet.  Now the extractor interface is standardized whether the underlying feature extractor is from vlfeat or opencv.


jlaura:
Dog food again...

PR #136, opened 2016-09-30
First draft of configure.py and bundle.sip

Kelvinrr:
configure.py still relies on the user to have ISIS set in order to take advantage of its make system. The bundle.sip file still needs to have more Bundle classes wrapped to be useful.


jlaura:
@kelvinrr Lets close this here and push to the newly created repo.  I think that the decomposition of this library and the wrapper to an optional dependency is going to be cleaner.

PR #144, opened 2016-10-18
Fixes Suppression

jlaura:
This PR incorporates @acpaquette's changes in another PR and includes additional fixes to the spatial suppression check.  The essence of the fix is that the binary search space was far too granular and the correct cell size was never being selected because the discrete search space did not include it. 

The fix was to increase the granularity by an order of magnitude.  Not sure if this is a long term fix given the potential for images that are many orders of magnitude larger.  The alternative will be to rewrite the binary search to operate on a more continuous space; do not preallocate the search space.

### DIFF
--- a/autocnet/graph/node.py
+++ b/autocnet/graph/node.py
@@ -251,8 +251,10 @@ class Node(dict, MutableMapping):
 
         allkps = pd.DataFrame(data=clean_kps, columns=columns, index=index)
 
-        self._keypoints = allkps.sort_values(by='response', ascending=False)
-
+        if 'response' in allkps.columns:
+            self._keypoints = allkps.sort_values(by='response', ascending=False)
+        elif 'size' in allkps.columns:
+            self._keypoints = allkps.sort_values(by='size', ascending=False)
         if isinstance(in_path, str):
             hdf = None
 
@@ -299,7 +301,7 @@ class Node(dict, MutableMapping):
         if isinstance(out_path, str):
             hdf = None
 
-    def group_correspondences(self, cg, *args, clean_keys=['fundamental'], deepen=False, **kwargs):
+    def group_correspondences(self, cg, *args, deepen=False, **kwargs):
         """
 
         Parameters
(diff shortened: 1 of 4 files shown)


jlaura:
Looks like an error in the suppression testing.  Do not merge until I am passing please.

jlaura:
Testing this with much larger images, I am still seeing issues with the suppression.  I suspect that the portioning of the space needs to be a function of the longest image edge.  This needs to be tested and integrated if accurate. 

The logic of the binary search and cell size shifting appears to be correct. (I am just not brave enough to claim it is 100%.)

jlaura:
@acpaquette Give this a check and see if it is working more consistently for you.

You can pull this local to play around using:

    git checkout -b jlaura-master dev
    git pull https://github.com/jlaura/autocnet.git master

acpaquette:
@jlaura In regards to suppression, I don't really know what consistent should look like. The suppression on the source image is consistently good but the destination seems like it is going to be inconsistent. From what I can tell we never look at the destinations points so nothing will be suppressed on the destination image. When you visualize the suppression, the source points correspond  to "random" values in the destination. The values obviously aren't random, just whatever the source points correspond to on the destination. There is no check or guarantee that the values on the destination will be suppressed like they are on the source node. I may be completely off here, but this is what I understand after looking through the code and running my testing notebook.

jlaura:
@acpaquette We are seeing very different things then.  Please post your notebook as a gist and link back to here. (gist.github.com) to get access to get to the gist homepage.  Then you can open the ipynb in a text editor and copy/paste the contents into the gist.

## On the matches
If we take the matches the from a single image, they have some distribution.  Theoretically, the distribution is identical in the other image after applying some transformation function.  So (f() = T(g()), where f and g are the two distributions.  If we spatially suppress one distribution and reduce the total number of matches, we are also suppressing the other distribution.  The extent of the suppression (x, y size) might be smaller in the second image, but the suppressed distribution will be the same.

If this is not the case, we have bug - so show me some pictures! :)

jlaura:
@acpaquette These are conda related test failures that look like issues with Travis being able to access the conda repos.  The build is clean on the linux side.  If you review and this looks that it does in fact fix the issue, go ahead and merge.  The tests will run again and, if the issue is then on our side, I can roll the merge back out and try again.

acpaquette:
@jlaura https://gist.github.com/acpaquette/edec9240b28d6db7ac92d069fd65c38b Here is the gist for my notebook.

jlaura:
Can you add .ipynb so that it renders as an ipython notebook? Thanks!

jlaura:
@acpaquette Restarted the build.  If this looks clean, lets merge today so that the ticket can be closed this sprint.  If it does not look clean, post a comment for cleanup.

PR #169, opened 2016-12-03
Cuda

jlaura:
This PR incorporates CUDA support into autocnet.  By default, the cuda support will not load.  If the user calls `autocnet.cuda(enable=True)` the pyCudaSift library will attempt to import and a GPU will be allocated (the above func has a second kmart `gpu` that is the device id for multi-device systems. 

Once loaded, the GPU enabled funcs are patched into the Node and Edge classes where appropriate.


jlaura:
Not really sure what is going on with Travis.  The Linux build is failing because of a lack of a minstar key, but the env. variable is set (and not echoing properly).  This might be because the build is never supposed to upload though - this is a PR.  Either way, the build exits `0` and then hangs.  This looks like Travis and not us.

PR #174, opened 2017-01-03
Attributes (Closes #165)

### LINKED ISSUE #165: As the process, the Node and Edge class attributes are accessible via dict notation so that functionality is inline with NetworkX standards
```python
n = Node()
n.attribute # The current functionality to maintain
n['attribute'] # The functionality expected via NetworkX as the node has an attributes dict.
```

acpaquette:


Kelvinrr:
Travis passed but is reporting failed, not our fault. Code looks fine to me 👍

PR #182, opened 2017-02-18
Make Travis Great Again

Kelvinrr:
~~Trying to make travis work again. Seems to pass on Linux. MacOS is a pain because it takes 1 hour before the tests even start. Even then, often fails before using `.travis.yml`. Maybe some way to mimic the Travis test env locally?~~

Seems to be working, although it feels a bit hack-y

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,6 @@ os:
   - linux
   - osx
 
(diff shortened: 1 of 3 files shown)


jlaura:
:+1:

PR #184, opened 2017-02-27
Fixes a bug where the ratio test would fail because it assumed implic…

jlaura:
…it ordering


jlaura:
In the ratio check function if iloc[0] and iloc[1] are being used to grab the two values to be compared it is essential that the group (data frame) be explicitly sorted on the column of interest.  This changes adds that sorting.

Kelvinrr:
Travis is suffering from an outage, there AWS servers are down. Im sure this is fine, but will hold off merge until they we get a green checkmark here assuming it doesn't take long.

PR #185, opened 2017-03-10
Edge keypoint function

evindunn:
Function on edge that returns keypoints for source or destination node

### DIFF
--- a/autocnet/graph/tests/test_edge.py
+++ b/autocnet/graph/tests/test_edge.py
@@ -249,3 +249,75 @@ class TestEdge(unittest.TestCase):
         for i in e.matches['vor_weights']:
             self.assertAlmostEquals(i, weights['vor_weights'][k])
             k += 1
+
+    def test_get_keypoints(self):
+        src_keypoint_df = pd.DataFrame({'x': (0, 1, 2, 3, 4), 'y': (5, 6, 7, 8, 9),
+                                        'response': (10, 11, 12, 13, 14), 'size': (15, 16, 17, 18, 19),
+                                        'angle': (20, 21, 22, 23, 24), 'octave': (25, 26, 27, 28, 29),
+                                        'layer': (30, 31, 32, 33, 34)})
+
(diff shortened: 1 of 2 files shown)


autocnet/graph/tests/test_edge.py
```diff
@@ -249,3 +249,52 @@ def test_voronoi_homography(self):
         for i in e.matches['vor_weights']:
             self.assertAlmostEquals(i, weights['vor_weights'][k])
             k += 1
+
```
acpaquette: This is going to be pretty slow for a test and potentially inconsistent. Rather than building a candidate graph, try mocking the attributes that you need using Mock and MagicMock to develop the test. If you checkout the test_coverage test in test_edge, there should already be a fair amount there that you can borrow for testing.

evindunn: I've updated the test function to use MagicMock, could you take a look and let me know what you think?

autocnet/graph/tests/test_edge.py
```diff
@@ -249,3 +249,65 @@ def test_voronoi_homography(self):
...
+        src_keypoint_df = pd.DataFrame({'x': (0, 1, 2, 3, 4), 'y': (5, 6, 7, 8, 9),
+                                        'response': (10, 11, 12, 13, 14), 'size': (15, 16, 17, 18, 19),
+                                        'angle': (20, 21, 22, 23, 24), 'octave': (25, 26, 27, 28, 29),
+                                        'layer': (30, 31, 32, 33, 34)})
+
+        dst_keypoint_df = pd.DataFrame({'x': (34, 33, 32, 31, 30), 'y': (29, 28, 27, 26, 25),
+                                        'response': (24, 23, 22, 21, 20), 'size': (19, 18, 17, 16, 15),
+                                        'angle': (14, 13, 12, 11, 10), 'octave': (9, 8, 7, 6, 5),
+                                        'layer': (4, 3, 2, 1, 0)})
+
```
acpaquette: I'm not sure if you need to have response, size, angle, etc. If you only need the x and y coordinates then the others probably aren't necessary. Other than that, everything looks good! :+1:

PR #188, opened 2017-04-27
Code Coverage & Updates to Edge

evindunn:
- Conditionals using Edge.matches were broken, updated to use isinstance(), determining whether Edge.matches is df or None
- There was a fundamental matrix lookup loop in Edge.masks(); Removed this, no longer necessary
- Lots of coverage updates to Edge
- control.py classes & their methods should now be completely covered


Kelvinrr:
Try pulling from upstream and then update this PR. The error should go away.

PR #203, opened 2017-05-17
Updates to documentation and a single bug fix that was apparently made in the repo...

jlaura:


jlaura:
Merging for the docs - if the single line change is breaking I can roll that back out.

PR #212, opened 2017-07-20
"overlap" keypoint constraint & match mask

evindunn:
- overlap=bool param for Edge.get_keypoints(); Masks returned keypoints that aren't within their Node's MBR
- Edge.overlap_check() that populates Edge.masks["overlap"]; Masks matches whose keypoints aren't within their Node's MBR
- Unit tests for both of the above


jlaura:
Linux passed and Coverage is up.  Nice job on this.  The overlap == mask insight is great.

Issue #192, opened 2017-05-05
The matcher.feature_matcher funcs should not have access to self.  Rewrite as funcs without class access.


jlaura:

jlaura:
Related to #191.  This is mainly a code smell issue.

evindunn:
- cpu_matcher.match() / cuda_matcher.edge() take "edge" instead of self
- static matcher methods are patched in to static Edge._match(edge)
- member method Edge.match() calls Edge._match(self)

Issue #191, opened 2017-05-05
The edge object has a method that is monkey patched in names `match`.  This method needs to reset the mask attribute to be empty before being called.  Decorators are an option.


jlaura:

evindunn:
@jlaura I want to do this while keeping  the` _add_matches()` functionality in mind, but I can't find where it's implemented. If it isn't implemented yet, should I add a keyword to Edge.match()? Something like `append_matches=bool`

PR #218, opened 2017-07-25
cpu/cuda_matcher.match() is static & clears Edge.masks

evindunn:
- cpu_matcher.match() & cuda_matcher.match() are static, take "edge" instead of self
- cpu_matcher.match() & cuda_matcher.match() reset Edge.masks to empty df
- matcher methods are patched in to static Edge._match(edge)
- member method Edge.match() calls static Edge._match(self)
- TestMatcher.test_cpu_match() adds coverage for cpu matcher

#191 and #192


jlaura:
:+1:

PR #221, opened 2017-07-27
Functional additions + decorator utils

Kelvinrr:


jlaura:
@Kelvinrr Apologies - the fix for the builds caused a tiny conflict in here - I can resolve your you can - nice and easy via the web editor these days.

PR #224, opened 2017-07-28
Updates for plio change

jlaura:
This PR removes the to_isis tests from the functional testing as this functionality should be handled in another class (such as a ControlMediator).  This also should fix the intermittent issues being caused with serialization/deserialization and edge ordering.  Finally, looks like a removal of cyvlfeat as a mandatory dependency snuck in.  Minimal change to `setup.py`.


jlaura:
Merging to get the other PRs building.  The coverage reduction is due to add_depth not being tested.  Since that functionality is moving out of the candidate graph (tentatively), I am not super worried.

PR #225, opened 2017-08-09
Added CandidateGraph.add_image()

evindunn:
- Allows user to add an image to an existing graph
- Optional 'adjacency' arg that can be dict or json file: Without it, tries to calculate adjacency using footprints; Without footprints, adds the image as a disconnected Node on the graph
- Optional 'basepath' arg: User can provide full path to image w/o basepath, or image file name with basepath
- Optional 'apply_func' arg: User can provide a static method or list of static methods that take an edge as their argument; These function(s) will be applied to any new edges generated when adding the image
```
    def extract_and_match(edge):
        for n in [edge.source, edge.destination]:
            n.extract_features(n.get_array(band=1), extractor_parameters={'nfeatures': 800})
        edge.match()
        edge.compute_overlap()
        edge.overlap_check()
        edge.symmetry_check()
        edge.ratio_check(clean_keys=['overlap', 'symmetry'])
        edge.compute_fundamental_matrix(clean_keys=['overlap', 'symmetry', 'ratio']))
```

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -26,11 +26,11 @@ from autocnet.vis.graph_view import plot_graph, cluster_plot
 
 
 # The total number of pixels squared that can fit into the keys number of GB of RAM for SIFT.
-MAXSIZE = {0:None,
-           2:6250,
-           4:8840,
-           8:12500,
-           12:15310}
+MAXSIZE = {0: None,
+           2: 6250,
+           4: 8840,
+           8: 12500,
+           12: 15310}
 
 class CandidateGraph(nx.Graph):
     """
@@ -165,7 +165,7 @@ class CandidateGraph(nx.Graph):
                     adjacency_dict[i.file_name].append(j.file_name)
                     adjacency_dict[j.file_name].append(i.file_name)
             except:
-                warnings.warn('Failed to calculated intersection between {} and {}'.format(i, j))
+                warnings.warn('Failed to calculate intersection between {} and {}'.format(i, j))
 
         return cls(adjacency_dict)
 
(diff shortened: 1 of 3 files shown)


autocnet/graph/network.py
```diff
@@ -223,16 +224,235 @@ def get_name(self, node_index):
...
+                      "when adding {0}; Could not add {0} to " \
+                      "CandidateGraph".format(image_name)
+                # Type Check
+                try:
+                    assert callable(apply_func) or type(apply_func) is list
+                except AssertionError:
+                    raise TypeError(ERR)
+
+                # If it's a function, apply it
+                if callable(apply_func):
```
Kelvinrr: you seem to check type twice (checking if a function or list of functions), might be a good idea to reduce them.

Kelvinrr:
Comments are in line, let me know what you think about some of these issues. Overall, the amount of logic in this function makes me nervous. Must be a way to simplify or displace some things here and there. Either that or I'm full of it.

autocnet/graph/network.py
```diff
@@ -223,16 +224,235 @@ def get_name(self, node_index):
...
+                # Grab footprints from nodes that have them
+                fp = ds.footprint
+                if fp and fp.IsValid():
+                    valid_datasets.append(ds)
+                else:
+                    warnings.warn('Missing or invalid geospatial data for '
+                                  '{}'.format(os.path.basename(ds.file_name)))
+
+            # Grab the footprints and test for intersection
+            for ds in valid_datasets:
```
Kelvinrr: Probably a faster way to determine overlap. Might be worth trying to take advantage of an r-tree? Would it be worth the extra logic? Or does simply iterating though everything good enough for now?

evindunn:
Thanks @Kelvinrr, I'll get to work on these suggestions

autocnet/graph/network.py
```diff
@@ -223,16 +224,235 @@ def get_name(self, node_index):
...
         Parameters
         ----------
+        image_name : str
+                     The file name of or path to the image to add
+
+        adjacency : dict or path to json
+                    The an adjacency dictionary or json file containing
+                    the adjacency dictionary for the image
+
+        basepath : str
```
evindunn: I was thinking so that the user doesn't have to provide the full path to all adjacent images; The basepath can be inserted into the image being added as well as the names of its adjacent images. Do you think I should get rid of it?

Kelvinrr: I imagine it is simple enough in python to glob things together that if the user really wanted to batch large number of images it wouldn't an issue.

autocnet/graph/network.py
jlaura: I would suggest that brute force is ideal for now.  Moving to an r-Tree in the future would be a great enhancement not only for this use case but also for initial construction.

autocnet/graph/network.py
```diff
@@ -223,16 +224,198 @@ def get_name(self, node_index):
...
+        apply_func : function
+                     A static function that takes an Edge as its parameter
+                     Function will be applied to all Edges generated when adding
+                     the image
 
         """
-        raise NotImplementedError
-        self._order_adjacency()
+
+        def add_node(img_pth):
```
jlaura: This is good as a closure because it is naturally named and does not override the networkx add_node method

autocnet/graph/network.py
```diff
@@ -223,16 +224,198 @@ def get_name(self, node_index):
...
+            self.graph['node_name_map'][img_nm] = id
+            self.graph['node_counter'] += 1
+
+            # Return the Node reference
+            return node
+
+        # Check if image is already in the graph
+        if image_name in self.graph['node_name_map']:
+            warnings.warn("{} is already in the graph".format(image_name))
+            return
```
jlaura: 👍

PR #229, opened 2017-10-24
Control

jlaura:
This fixes (for good...) the way that epipolar geometry and distances are being computed.


jlaura:
This should be updated and good to go.  All tests are passing locally.  Could still have env. issues with networkx, etc.  I did not check those in this PR.

PR #230, opened 2017-10-28
Added Dockerfile for autocnet environment

evindunn:
Containerized Linux dev environment for autocnet

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,9 +10,7 @@ RUN apt-get install -y libgl1-mesa-glx
 
 # Anaconda environment
 ADD environment.yml /root
-RUN conda install -y python=3.5.4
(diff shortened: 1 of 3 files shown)


evindunn:
@jlaura with the environment.yml included in this pr, my container runs the tests no problem. Maybe we should try it with Travis?

Kelvinrr:
The problem is that `.travis.yml` is installing the latest version of NetworkX which updated to 2.0 like a month ago. We need to set the NetworkX requirement to 1.11 until we can go in and update everything to correspond with the 2.0 release.

evindunn:
Reworking the deps for docker image to match up with new networkx

jlaura:
:+1: The linux builds are all passing.  Travis OS X is slow as dirt recently...

PR #234, opened 2017-12-04
Updates to NetworkX2

jlaura:
This PR updates the API to networkX 2.0.  

Major changes:
`CandidateGraph.add_image` is intentionally commented out.  This method is going to need some thought before adding it back to the graph for working with the new API.

Otherwise, I tried to keep most of the changes centralized inside of network.  

## _iters
I added backwards compatibility methods to the CandidateGraph that should not break our existing examples / docs.  These should be yielding out the node and edge objects in a way that previous users are expecting.  These are very lightly tested and could use some more robust testing (either inside of the graph or inside the 3 image functional test).

This PR also provides an example for porting standard unittest testing to PyTest and using fixtures / parametrization.  The ciratefi test has been updates with 2 different styles of parametrized tests.  Comments on that section solicited.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1175,6 +1175,20 @@ class CandidateGraph(nx.Graph):
         files = self.files()
         self.controlnetwork.to_isis(outname, serials, files, *args, **kwargs)
 
+    def nodes_iter(self, data=False):
+        for i, n in self.nodes.data('data'):
+            if data:
+                yield i, n
+            else:
+                yield i
+
+    def edges_iter(self, data=False):
+        for s, d, e in self.edges.data('data'):
+            if data:
+                yield s, d, e
+            else:
+                yield s, d
+
 class SubCandidateGraph(nx.graphviews.SubGraph, CandidateGraph):
     def __init__(self, *args, **kwargs):
         super(SubCandidateGraph, self).__init__(*args, **kwargs)
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -34,7 +34,6 @@ MAXSIZE = {0: None,
            8: 12500,
            12: 15310}
 
-
 class CandidateGraph(nx.Graph):
     """
     A NetworkX derived directed graph to store candidate overlap images.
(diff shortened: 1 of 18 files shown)


jlaura:
@all I am working on `plio` now so that our builds are clean again.

jlaura:
@all plio builds are working on linux and should be working on OSX (though Travis is very slow recently for OSX): https://travis-ci.org/USGS-Astrogeology/plio

Autocnet builds are working on linux as well: https://travis-ci.org/USGS-Astrogeology/autocnet
Same deal with the OSX build speed.  No promises on coverage with the bump...

autocnet/graph/network.py
```diff
@@ -345,7 +301,7 @@ def add_node(img_pth):
                 return
 
             # Detect adjacency between our new node and the CG's nodes
-            target_nodes = [self.node[idx] for idx in self.nodes()]
+            target_nodes = [self.node[idx] for idx in self.nodes()]  # This is broken too
```
Kelvinrr: outdated comment?

autocnet/graph/network.py
```diff
@@ -893,6 +881,7 @@ def create_edge_subgraph(self, edges):
         H : object
             A networkx subgraph object
         """
+        return self.edge_subgraph(edges)
```
Kelvinrr: return statement before the rest of the body seems odd. Did you mean to delete the rest of the body after this line?

autocnet/graph/network.py
```diff
@@ -1225,3 +1174,23 @@ def to_isis(self, outname, *args, **kwargs):
...
+            if data:
+                yield s, d, e
+            else:
+                yield s, d
+
+class SubCandidateGraph(nx.graphviews.SubGraph, CandidateGraph):
+    def __init__(self, *args, **kwargs):
+        super(SubCandidateGraph, self).__init__(*args, **kwargs)
+
+nx.graphviews.SubGraph = SubCandidateGraph
```
Kelvinrr: why override the `SubGraph` class rather than simply using `SubCandidateGraph` directly?

autocnet/graph/tests/test_network.py
```diff
@@ -429,22 +441,21 @@ def test_set_maxsize(graph):
...
-def test_is_complete(graph):
+"""def test_is_complete(graph):
     # Create a small incomplete graph with three nodes and two edges
     incomplete_graph = network.CandidateGraph()
     incomplete_graph.add_nodes_from([1, 2, 3])
     incomplete_graph.add_edges_from([(1, 2), (2, 3)])
 
     assert False == incomplete_graph.is_complete()
-    assert True == graph.is_complete()
+    assert True == graph.is_complete()"""
```
Kelvinrr: Is no longer testing completeness intentional? if so, then should we delete this rather than commenting it out?

autocnet/graph/network.py
jlaura: The whole method is commented out and not working properly.  I left in this comment for when one of use gets back to working on add_image.

autocnet/graph/network.py
jlaura: Yup.  Yanked it.

autocnet/graph/network.py
jlaura: So that we can use edge_subgraph and node_subgraph direct from networkx without having to reimplement them.  Your previous comment about remove the stuff after the return illustrates why I did this; to avoid reimplementation.

jlaura:
The Travis failure is an abort trap 6 on OS X (https://travis-ci.org/USGS-Astrogeology/autocnet/jobs/315035653#L611).  I have been seeing this failure intermittently on my laptop as well. Given where the failure is happening in the testing, I think that this is an issue with GDAL, GEOS, or OGR though this is conjecture (I have not run through a debugger).  I have not been able to reproduce this failure on a linux box.  Is anyone else experiencing this issue?

PR #236, opened 2018-01-02
Adds add_node() and add_edge() to CandidateGraph

evindunn:
Extends networkx.graph.add_node() and add_edge() to take image file names


jlaura:
:+1: This is really tight. Nice.  I like the dispatching off to nx2 functionality.  I learned another in line use of super - thanks!

PR #248, opened 2018-06-26
improving conda commands

michaelaye:
The addition of channels like menpo and conda-forge drastically change the dependency chain, so it's better to have these configurations only added for the newly created environment. This is achieved by adding the `--env` option to the `conda config` command.
Also, after adding the `usgs-astrogeology` channel, it's not required in the `install` command.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,4 +47,4 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
    * ``conda config --env --add channels conda-forge``
    * ``conda config --env --add channels menpo``
    * ``conda config --env --add channels usgs-astrogeology``
-#. Finally, install autocnet: ``conda install autocnet-dev``
+#. Finally, install autocnet: ``conda install autocnet``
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -38,9 +38,13 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
...
+#. Bring up a command line and add three channels to your conda environment-specific config file:
    
-   * ``conda config --add channels conda-forge``
-   * ``conda config --add channels menpo``
-   * ``conda config --add channels usgs-astrogeology``
-#. Finally, install autocnet: ``conda install -c usgs-astrogeology autocnet``
+   * ``conda config --env --add channels conda-forge``
+   * ``conda config --env --add channels menpo``
+   * ``conda config --env --add channels usgs-astrogeology``
+#. Finally, install autocnet: ``conda install autocnet-dev``
```
jlaura: Can you pull the `-dev` out of this line?  We have moved away from using `-dev` and therefore, only push the autocnet package to anaconda.org.  Thanks!

Kelvinrr:
MacOS is eating it for dumb Travis reasons, so ignore it.

PR #254, opened 2018-07-03
CN refactor + bug fixes 

Kelvinrr:
* Control network generation is now much faster 
* removed ControlNetwork class, the control network now exists as a simple pandas dataframe on the network 
* tutotial with vis notebook updated with past autocnet changes


autocnet/graph/edge.py
```diff
@@ -395,6 +400,7 @@ def suppress(self, suppression_func=spf.correlation, clean_keys=[], maskname='su
         # Massage the dataframe into the correct structure
         coords = self.source.get_keypoint_coordinates()
         merged = matches.merge(coords, left_on=['source_idx'], right_index=True)
+        print(merged.columns)
         merged['strength'] = merged.apply(suppression_func, axis=1, args=([self]))
```
jlaura: yank the print.

autocnet/control/control.py
```diff
@@ -9,31 +9,37 @@
...
+        df = pd.DataFrame(
+            offs, columns=['x_off', 'y_off', 'corr'], index=group.index)
         return df
-    gps = cn.data.groupby('point_id').apply(subpixel_group,threshold=0.9,max_x_shift=5, max_y_shift=5,template_size=template_size, search_size=search_size,**kwargs)
-    cn.data[['x_off', 'y_off', 'corr']] = gps.reset_index()[['x_off', 'y_off', 'corr']]
+    gps = cn.data.groupby('point_id').apply(subpixel_group, threshold=0.9, max_x_shift=5,
+                                            max_y_shift=5, template_size=template_size, search_size=search_size, **kwargs)
+    cn.data[['x_off', 'y_off', 'corr']] = gps.reset_index()[
+        ['x_off', 'y_off', 'corr']]
+
```
jlaura: This looks fine. I just opened #256 to address refactoring this out in the future.

autocnet/graph/network.py
```diff
@@ -26,13 +27,16 @@
 from autocnet.vis.graph_view import plot_graph, cluster_plot
 from autocnet.control import control
 
+np.warnings.filterwarnings('ignore')
+
```
jlaura: 👍

autocnet/graph/network.py
```diff
@@ -120,7 +142,8 @@ def maxsize(self):
     @maxsize.setter
     def maxsize(self, value):
         if not value in MAXSIZE.keys():
-            raise KeyError('Value must be in {}'.format(','.join(map(str,MAXSIZE.keys()))))
+            raise KeyError('Value must be in {}'.format(
+                ','.join(map(str, MAXSIZE.keys()))))
         else:
```
jlaura: :+1:

autocnet/graph/network.py
```diff
@@ -348,14 +374,17 @@ def extract_features_with_downsampling(self, downsample_amount=None, *args, **kw
         """
         for node in self.nodes:
             if downsample_amount == None:
-                total_size = node.geodata.raster_size[0] * node.geodata.raster_size[1]
+                total_size = node.geodata.raster_size[0] * \
+                    node.geodata.raster_size[1]
                 downsample_amount = math.ceil(total_size / self.maxsize**2)
-            node.extract_features_with_downsampling(downsample_amount, *args, **kwargs)
+            node.extract_features_with_downsampling(
+                downsample_amount, *args, **kwargs)
```
jlaura: I think that this got refactored to use apply in the merge that is going to need to happen.

autocnet/graph/network.py
```diff
@@ -348,14 +374,17 @@ def extract_features_with_downsampling(self, downsample_amount=None, *args, **kw
...
 
-    def extract_features_with_tiling(self, tilesize=1000, overlap=500, *args, **kwargs): #pragma: no cover
+    def extract_features_with_tiling(self, tilesize=1000, overlap=500, *args, **kwargs):  # pragma: no cover
         for i, node in self.nodes(data='data'):
             print('Processing {}'.format(node['image_name']))
-            node.extract_features_with_tiling(tilesize=tilesize, overlap=overlap, *args, **kwargs)
+            node.extract_features_with_tiling(
+                tilesize=tilesize, overlap=overlap, *args, **kwargs)
```
jlaura: As above - merge covered this?

autocnet/graph/network.py
```diff
@@ -403,6 +430,7 @@ def match(self, *args, **kwargs):
         autocnet.graph.edge.Edge.match
         """
         self.apply_func_to_edges('match', *args, **kwargs)
+        # self.generate_control_network()
```
jlaura: just remove?

autocnet/graph/network.py
```diff
@@ -467,8 +495,8 @@ def compute_triangular_cycles(self):
         cycles = []
         for s, d in self.edges:
             for n in self.nodes:
-                if(s,n) in self.edges and (d,n) in self.edges:
-                    cycles.append(tuple(sorted([s,d,n])))
+                if(s, n) in self.edges and (d, n) in self.edges:
+                    cycles.append(tuple(sorted([s, d, n])))
         return set(cycles)
```
jlaura: What do you think about using: https://github.com/ambv/black

We could set up a linter to just apply this to every PR. Removes stuff like this and helps us (me) be more consistent.

autocnet/graph/network.py
```diff
@@ -1088,8 +1123,205 @@ def edges_iter(self, data=False):
             else:
                 yield s, d
 
+    def generate_control_network(self, clean_keys=[], mask=None):
+        """
+        TODO: Find a way to make this less dirt slow
```
jlaura: I think this is or is an update to the PR needed?

autocnet/graph/network.py
```diff
@@ -1088,8 +1123,205 @@ def edges_iter(self, data=False):
...
+            point_id : hashable
+                       The point to link the node to.  This is most likely an integer, but
+                       any hashable should work.
+            """
+            if key in self.measure_to_point.keys():
+                return
+            if point_id == None:
+                point_id = self._point_id
+            self.measure_to_point[key] = point_id
+            # The node_id is a composite key (image_id, correspondence_id), so just grab the image
```
jlaura: This is going to have to go to a setter/getter if we want to have it persist to a DB in autocnet_server. If creating the network is fast though, no need. Thoughts?  Either way, I would not do that in this PR, rather open an issue if needed / desired.

autocnet/graph/network.py
```diff
@@ -1088,8 +1123,205 @@ def edges_iter(self, data=False):
...
+        if all(map(lambda x:x[2].empty, mask_list)):
+            return
+
+        new_graph_mask = pd.DataFrame()
+        for s,d, mask in mask_list:
+                mask = mask.reset_index().rename(columns={'index':'match_idx'})
+                mask['edge'] = [(s['node_id'], d['node_id'])]*mask.shape[0]
+                if 'source' in mask.columns and 'destination' in mask.columns:
+                    mask = mask.drop(['source', 'destination'], axis=1)
+                new_graph_mask = new_graph_mask.append(mask)
```
jlaura: Is it cheaper to append here or to have a python list to append to and then concat in a data frame? (or is this not the slow point?)

autocnet/graph/network.py
```diff
@@ -1088,8 +1123,205 @@ def edges_iter(self, data=False):
...
+
+        # updated previous column
+        for column in new_mask_sorted:
+            current_mask_sorted[column] = new_mask_sorted[column]
+
+        self.masks = current_mask_sorted
+        return self.masks
+
+
+    def add_measure(self, key, edge, match_idx, fields, point_id=None):
```
jlaura: Is this double defined - line 1130?

autocnet/graph/network.py
```diff
@@ -1088,8 +1123,205 @@ def edges_iter(self, data=False):
...
+
+    def clean_singles(self):
+        """
+        Take the `controlnetwork` dataframe and return only those points with
+        at least two measures.  This is automatically called before writing
+        as functions such as subpixel matching can result in orphaned measures.
+        """
+        return self.controlnetwork.groupby('point_id').apply(lambda g: g if len(g) > 1 else None)
+
+    def to_isis(self, outname, serials, olist, *args, **kwargs):  # pragma: no cover
```
jlaura: Also double defined I believe

Kelvinrr: The one in the CN class was moved here replacing CG's, since CG's simply dispatched to the CN's

autocnet/graph/network.py
Kelvinrr: if you mean setter/getter for measure_to_point, no because it's made fresh every time the network is generated.

autocnet/graph/network.py
Kelvinrr: Since we agreed to generate each time, this method no longer has a purpose is removed.

autocnet/graph/network.py
Kelvinrr: removed, since we dont need to manually add measures and it seems it only has a purpose within the scope of the `generate_control_network` functions.

Kelvinrr:
Should be good to go whenever. Probably worth another quick look through but after the previous comments, I only removed things.

autocnet/graph/network.py
jlaura: :+1:

PR #262, opened 2018-09-12
Fixes serialization and refactors to properties

jlaura:


.travis.yml
```diff
@@ -37,15 +34,10 @@ before_install:
...
   - conda config --add channels conda-forge
-  - conda config --set ssl_verify false
-  - conda install -c conda-forge vlfeat scipy networkx numexpr cython matplotlib pillow runipy geopandas opencv numpy libgdal gdal scikit-image
-  - conda install -c menpo cyvlfeat
-  - conda install -c usgs-astrogeology plio
-  
-  # Development installation
-  - conda install -c conda-forge pytest pytest-cov coverage anaconda-client
-  - pip install coveralls
+  - conda env update -n test -f environment.yml
```
jlaura: Much cleaner way to force synchronization with the environment.yml (for installers) and the testing env.

autocnet/control/tests/test_control.py
```diff
@@ -66,3 +66,4 @@ def test_potential_overlap(controlnetwork, candidategraph):
                                  (1,), (1,),
                                  (0,), (0,)],
                                  index=[6,7,8,9,10,11]))
+"""
```
jlaura: @Kelvinrr  Need to verify that all of this logic is now in the network?  Therefore, these tests should be removed / migrated.

autocnet/graph/edge.py
```diff
@@ -70,12 +69,25 @@ def __eq__(self, other):
...
+            self._masks = pd.DataFrame()
+        return self._masks
+
+    @masks.setter
+    def masks(self, value):
+        if isinstance(value, pd.DataFrame):
+            self._masks = value
+        else:
+            raise(TypeError)
+
```
jlaura: To support autocnet_server we need properties.

autocnet/graph/edge.py
```diff
@@ -189,12 +201,14 @@ def add_coordinates_to_matches(self):
...
+        self.matches['source_y'] = skps.values[:,1]
         dkps = self.get_keypoints(self.destination, index=self.matches.destination_idx)
-        self.matches['source_x'] = skps.x.values
-        self.matches['source_y'] = skps.y.values
-        self.matches['destination_x'] = dkps.x.values
-        self.matches['destination_y'] = dkps.y.values
-
+        dkps.reindex(self.matches['destination_idx'])
+        self.matches['destination_x'] = dkps.values[:,0]
+        self.matches['destination_y'] = dkps.values[:,1]
```
jlaura: This is the one point I need to triple check. The logic here is that the key points (x,y) should be added to the matches data frame. This is easy, except that one has to be careful that the matches[source_idx] ordering matches the index of the x/y key points data frame.

autocnet/graph/edge.py
```diff
@@ -210,14 +224,16 @@ def project_matches(self, semimajor, semiminor, on='source', srid=None):
...
+        
         gnd = np.empty((len(coords), 3))
         # Project the points to the surface and reproject into latlon space
         for i in range(gnd.shape[0]):
-            gnd[i] = camera.imageToGround(coords[i][0], coords[i][1], 0)
-        lon, lat, alt = reproject(gnd.T, semimajor, semiminor,
+            ic = csmapi.ImageCoord(coords[i][0], coords[i][1])
+            ground = camera.imageToGround(ic, 0)
+            gnd[i] = [ground.x, ground.y, ground.z]
+        lon, lat, alt = spatial.reproject(gnd.T, semimajor, semiminor,
```
jlaura: This refactor uses the new and improved CSMAPI interface that is more C++ like.

autocnet/graph/edge.py
```diff
@@ -446,36 +461,46 @@ def subpixel_register(self, method='phase', clean_keys=[],
...
+                                    size_x=updated_size, size_y=updated_size)         
             shift_x, shift_y, metrics = func(s_template, d_search, **kwargs)
 
             # ROIs and clipping all work using whole pixels. The clip_roi func returns
             # the subpixel components that are lost when converting to whole pixels
             # reapply those here.
-            shift_x += dx
-            shift_y += dy
+            shifts_x[i] = shift_x + dxr
+            shifts_y[i] = shift_y + dyr
```
jlaura: These updates the sub pixel matcher to simply return the new coordinate and not shifts that the user then needs to track mentally and apply when they want them.

autocnet/graph/edge.py
```diff
@@ -484,13 +509,13 @@ def subpixel_register(self, method='phase', clean_keys=[],
...
 
         if method == 'phase':
-            self.costs.loc[mask, 'phase'] = [i[0] for i in strengths]
-            self.costs.loc[mask, 'rmse'] = [i[1] for i in strengths]
+            self.costs.loc[mask, 'phase_diff'] = strengths[:,0]
+            self.costs.loc[mask, 'rmse'] = strengths[:,1]
         elif method == 'template':
-            self.costs.loc[mask, 'correlation'] = strengths
+            self.costs.loc[mask, 'correlation'] = strengths[:,0]
+
```
jlaura: This allows the method to return 'goodness' measures of 1 or 2 dimensions.

autocnet/graph/network.py
```diff
@@ -27,7 +27,7 @@
 from autocnet.vis.graph_view import plot_graph, cluster_plot
 from autocnet.control import control
 
-np.warnings.filterwarnings('ignore')
+#np.warnings.filterwarnings('ignore')
```
jlaura: This is awesome, until is suppresses warning that we are testing for inside of pytest.

autocnet/graph/network.py
```diff
@@ -133,7 +133,9 @@ def __eq__(self, other):
         if sorted(self.edges()) != sorted(other.edges()):
             return False
         for s, d, e in self.edges.data('data'):
-            if not e == other.edges[s, d]['data']:
+            if s > d:
+                s, d = d, s
+            if not e == other.edges[(s, d)]['data']:
```
jlaura: Checks the serialization to avoid weird behavior where (s,d) != (d,s)

autocnet/graph/node.py
```diff
@@ -335,22 +327,19 @@ def extract_features(self, array, xystart=[], camera=None, *args, **kwargs):
...
         concat_kps.reset_index(inplace=True, drop=True)
         concat_kps.drop_duplicates(inplace=True)
         #descriptor_mask = descriptor_mask[count:]        
         # Removed duplicated and re-index the merged keypoints
         
-        
         if self.descriptors is not None:
             concat = np.concatenate((self.descriptors, new_descriptors))
-            new_descriptors = concat[concat_kps.index]
+            new_descriptors = concat[concat_kps.index.values]
```
jlaura: Need to be very careful with numpy index vs. pandas index objects.

autocnet/graph/node.py
```diff
@@ -407,8 +396,10 @@ def project_keypoints(self):
         # Project the sift keypoints to the ground
         def func(row, args):
             camera = args[0]
-            gnd = getattr(camera, 'imageToGround')(row[1], row[0], 0)
-            return gnd
+            imagecoord = csmapi.ImageCoord(float(row[1]), float(row[0]))
+            # An elevation at the ellipsoid is plenty accurate for this work
+            gnd = getattr(camera, 'imageToGround')(imagecoord, 0)
+            return [gnd.x, gnd.y, gnd.z]
```
jlaura: for the CSMAPI change

autocnet/io/keypoints.py
```diff
@@ -187,3 +187,30 @@ def to_npy(keypoints, descriptors, out_path):
              keypoints=keypoints,
              keypoints_idx=keypoints.index,
              keypoints_columns=keypoints.columns)
+
```
jlaura: Migrated from plio because this is hyper specific to autocnet / autocnet_server

autocnet/io/network.py
```diff
@@ -76,10 +76,11 @@ def save(network, projectname):
...
                     ndarrays_to_write[k+'_columns'] = v.columns
-            # Handle the matches dataframe that is a property
-            ndarrays_to_write['matches'] = data.matches
-            ndarrays_to_write['matches_idx'] = data.matches.index
-            ndarrays_to_write['matches_columns'] = data.matches.columns
+            # Handle DataFrames that are properties
+            for k in ['_matches', '_masks', '_costs']:
+                ndarrays_to_write[k] = getattr(data, k, np.array([]))
+                ndarrays_to_write['{}_idx'.format(k)] = getattr(data, k, pd.DataFrame()).index
+                ndarrays_to_write['{}_columns'.format(k)] = getattr(data, k, pd.DataFrame()).columns
```
jlaura: Since we have multiple data frames on the object, serialization needed to be more generic.

autocnet/io/network.py
```diff
@@ -152,8 +153,10 @@ def load(projectname):
                 edge[k] = v
             try:
                 nzf = np.load(BytesIO(pzip.read('{}_{}.npz'.format(s,d))))
-                edge.masks = pd.DataFrame(nzf['masks'], index=nzf['masks_idx'], columns=nzf['masks_columns'])
-                edge.matches = pd.DataFrame(nzf['matches'], index=nzf['matches_idx'], columns=nzf['matches_columns'])
+                for j in ['_matches', '_masks', '_costs']:
+                    setattr(edge, j, pd.DataFrame(nzf[j], 
+                                                  index=nzf['{}_idx'.format(j)], 
+                                                  columns=nzf['{}_columns'.format(j)]))
```
jlaura: and so do the reads then as well.

jlaura: Really, the enumeration of these attributes should be defined in one place - probably inside of the class itself. This is not very maintainable.

autocnet/io/tests/test_keypoints.py
```diff
@@ -31,6 +31,21 @@ def test_read_write_hdf(tmpdir, kd):
...
+
+def test_read_hdf_with_index(tmpdir, kd):
+    kps, desc = kd                                                                 
+    path = tmpdir.join('out.h5')                                                   
+    keypoints.to_hdf(path.strpath, keypoints=kps, descriptors=desc) 
+
+    reloaded_kps, reloaded_desc = keypoints.from_hdf(path.strpath, index=np.arange(10))
+    assert len(reloaded_kps) == len(reloaded_desc) == 10
+    np.testing.assert_array_equal(np.arange(10), reloaded_kps.index.values)
+
```
jlaura: real test!

autocnet/matcher/cpu_ring_matcher.py
```diff
@@ -99,7 +101,7 @@ def ransac_permute(ref_points, tar_points, tolerance_val, target_points):
             for j in range(i):
                 m[i,j] = q[i].dot(q[j])
                 m[j,i] = m[i,j]
-        qm = m > target_points
+        qm = m >= target_points
```
jlaura: so writing tests shows that these need to be >= or <= (above lines as well). If you look at the test, you will see how many good points are being removed by using < >.

autocnet/matcher/tests/test_cpu_ring_matcher.py
```diff
@@ -41,4 +43,28 @@ def test_check_pidx_duplicates(arr, expected):
...
+                                      ])
+def test_ransac_permut(x, y, eidx):
+    xp, yp, idx = rm.ransac_permute(x, y, 0.2, 2)
+    np.testing.assert_array_equal(idx, eidx)
+
+
+def test_add_correspondences():
+    func = 'autocnet.matcher.cpu_ring_matcher.ring_match_one'
+    with mock.patch(func, return_value=1):
+        in_feats = np.array([[1,1], [2,2]])
```
jlaura: This test is really just testing that the func is being called. I need to do a lot more thinking about how the heck to write tests for the ring matcher. This is going to require refactoring the ring matcher down a bunch more to get the logic more modularized. We know empirically that this is working, but I would be nervous making too many changes because the tests are poor or non-existent.

autocnet/transformation/spatial.py
```diff
@@ -0,0 +1,42 @@
...
+    -------
+    : list
+      Transformed coordinates as y, x, z
+
+    """
+    source_pyproj = pyproj.Proj(proj = source_proj, a = semi_major, b = semi_minor)
+    dest_pyproj = pyproj.Proj(proj = dest_proj, a = semi_major, b = semi_minor)
+
+    y, x, z = pyproj.transform(source_pyproj, dest_pyproj, record[0], record[1], record[2], **kwargs)
+    return y, x, z
```
jlaura: direct from plio

autocnet/transformation/tests/test_spatial.py
```diff
@@ -0,0 +1,10 @@
+from unittest import mock
+import pytest
+
+from autocnet.transformation import spatial
+
+def test_reproject():
+    with mock.patch('pyproj.transform', return_value=[1,1,1]) as mock_pyproj:
+        res = spatial.reproject([1,1,1], 10, 10, 'geocent', 'latlon')
+        mock_pyproj.assert_called_once()
+        assert res == (1,1,1)
```
jlaura: we do not want to test pyproj, we want to test our interface. This should be doing that by mocking pyproj.transform.

environment.yml
```diff
@@ -1,26 +1,23 @@
...
-    - pillow
-    - pvl
-    - libpysal
-    - protobuf
+  - ncurses
+  - geopandas
+  - pytest
+  - pytest-cov
+  - coveralls
+  - anaconda-client
```
jlaura: So much cleaner....one place to change...happiness.

jlaura:
Locally, this has coverage back up to 80%. Still lots to do on the testing side. 

https://travis-ci.org/USGS-Astrogeology/autocnet/builds/428468618

tthatcher95:
@jlaura It passed all checks and looks good to me. The only thing, is maybe for the ring matcher tests we can open a ticket and get an idea for a few concrete tests? Just to improve upon our quality of life, and your sanity. 

Other than that :+1:

jlaura:
@tthatcher95 :+1: on the ring matcher tests. I am actually working on those right now.

autocnet/control/tests/test_control.py
Kelvinrr: Should be in the network. I think there is still some work to do to clean it up (some problems go through on that PR, like two `to_isis` function being defined). Should be rewritten to ensure functionality works from the graph.

PR #271, opened 2019-03-22
Matcher Update

acpaquette:
Changed ring matcher to have the same columns as the k nearest neighbors matcher for the source_image and destination_image


setup.py
```diff
@@ -32,20 +32,6 @@ def setup_package():
...
-              'vlfeat',
-              'cyvlfeat',
-              'csmapi',
-              'networkx',
-              'pillow',
-              'plio',
-              'ncurses',
-              'geopandas',
-              'imageio'],
         classifiers=[
```
jlaura: Why this change? Is it because everything is in the environment.yml?

autocnet/graph/edge.py
```diff
@@ -185,11 +185,11 @@ def ring_match(self, *args, **kwargs):
...
         matches[:,3] = tar_kps.index[pidx[:,1]].values
 
-        matches = pd.DataFrame(matches, columns=['source',
+        matches = pd.DataFrame(matches, columns=['source_image',
                                                  'source_idx',
-                                                 'destination',
+                                                 'destination_image',
                                                  'destination_idx']).astype(np.float32)
-        
+
```
jlaura: :+1:

setup.py
Kelvinrr: If I recall correctly, he was trying to run `setup.py develop` and it couldn't find one of the packages required despite it being conda installed, so I just told him to blast the list as a quick fix. Since things are all in the env file I don't see why they can't stay gone.

jlaura: :+1:

PR #276, opened 2019-05-19
Refactor

jlaura:


autocnet/__init__.py
```diff
@@ -26,6 +21,39 @@
...
+                                            config['database']['name'])
+    hostname = socket.gethostname()
+    engine = create_engine(db_uri, poolclass=pool.NullPool,
+                    connect_args={"application_name":"AutoCNet_{}".format(hostname)},
+                    isolation_level="AUTOCOMMIT")                   
+    Session = orm.session.sessionmaker(bind=engine)
+except: 
+    Session = None
+    engine = None
+
```
jlaura: This block is setting up the DB connection that autocnet might use. If it does not need one, the `Session` and `engine` variables are set because they are imports inside of autocnet.

autocnet/__init__.py
```diff
@@ -26,6 +21,39 @@
 else:
     __version__ = _dist.version
 
+#Load the config file and setup a global DB session factory
+try:
+    with open(os.environ['autocnet_config'], 'r') as f:
+        config = yaml.safe_load(f)
+except:
+    warnings.warn('No autocnet_config environment variable set. Defaulting to an en empty configuration.')
+    config = {}
```
jlaura: We use the config file if the project needs things like DB access. Otherwise, this is vanilla autocnet.

autocnet/graph/edge.py
```diff
@@ -1,14 +1,17 @@
...
 import warnings
-from collections import MutableMapping
+from collections import defaultdict, MutableMapping
 
+from geoalchemy2.elements import WKBElement
 import numpy as np
 import pandas as pd
 import networkx as nx
-
+import pyproj
```
jlaura: the server has sensor model support so can use projection

autocnet/graph/edge.py
```diff
@@ -1,14 +1,17 @@
...
 from scipy.spatial.distance import cdist
+from shapely.geometry import Point
+import sqlalchemy
 
-import autocnet
+from autocnet import Session, engine
```
jlaura: Here is that import from above of the session/engine. None in a vanilla instance.

autocnet/graph/edge.py
```diff
@@ -20,6 +23,8 @@
 from autocnet.transformation import spatial
 from autocnet.vis.graph_view import plot_edge, plot_node, plot_edge_decomposition, plot_matches
 from autocnet.cg import cg
+from autocnet.io.db.model import Images, Keypoints, Matches, Cameras, Network, Base, Overlay, Edges, Costs
+from autocnet.io.db.wrappers import DbDataFrame
```
jlaura: I colocated the object and the Networkobject variant in the same file, so the imports get long. Thoughts?

autocnet/graph/edge.py
```diff
@@ -704,8 +707,269 @@ def get_match_coordinates(self, clean_keys=[]):
...
         if self.matches.empty:
             return pd.DataFrame()
-        self.add_coordinates_to_matches()
+        #self.add_coordinates_to_matches()
         matches, _ = self.clean(clean_keys=clean_keys)
         skps = matches[['source_x', 'source_y']]
         dkps = matches[['destination_x', 'destination_y']]
         return matches
+
+class NetworkEdge(Edge):
```
jlaura: The idea here is that NetworkEdge object just extends the Edge class with the pieces necessary to get/put data into the database and or use remote services.

autocnet/graph/edge.py
```diff
@@ -704,8 +707,269 @@ def get_match_coordinates(self, clean_keys=[]):
...
+
+    default_msg = {'sidx':None,
+                    'didx':None,
+                    'task':None,
+                    'param_step':0,
+                    'success':False}
+
+    def __init__(self, *args, **kwargs):
+        super(NetworkEdge, self).__init__(*args, **kwargs)
+        self.job_status = defaultdict(dict)
```
jlaura: The status stuff is stubbed in but largely turned off because this is running pseudo locally, e.g., no cluster atm.

autocnet/graph/edge.py
```diff
@@ -704,8 +707,269 @@ def get_match_coordinates(self, clean_keys=[]):
...
+                                        first()
+
+        try:
+            df = pd.DataFrame.from_records(res[0])
+            df.index = df.index.map(int)
+        except:
+            ids = list(map(int, self.matches.index.values))
+            df = pd.DataFrame(index=ids)
+        df.index.name = 'match_id'
+        return DbDataFrame(df, parent=self, name='masks')
```
jlaura: Get the edge from the database and pull the masks column. This is a dict (json) so go ahead and load this from the record into a dataframe.

autocnet/graph/edge.py
```diff
@@ -704,8 +707,269 @@ def get_match_coordinates(self, clean_keys=[]):
...
+        res = session.query(Edges).\
+                                filter(Edges.source == self.source['node_id']).\
+                                filter(Edges.destination == self.destination['node_id']).first()
+        if res:
+            as_dict = df.to_dict()
+            dict_check(as_dict)
+            # Update the masks
+            res.masks = as_dict
+            session.add(res)
+            session.commit()
```
jlaura: If we are able to pull an edge, then we can store a masks object. Logic is as above except now we merge/update the json object with the `v` dataframe.

autocnet/graph/edge.py
```diff
@@ -704,8 +707,269 @@ def get_match_coordinates(self, clean_keys=[]):
...
+
+        if res:
+        # Parse the JSON dicts in the cost field into a full dimension dataframe
+            costs = {r.match_id:r._cost for r in res}
+            df = pd.DataFrame.from_records(costs).T  # From records is important because from_dict drops rows with empty dicts
+        else:
+            df = pd.DataFrame(index=ids)
+
+        df.index.name = 'match_id'
+        return DbDataFrame(df, parent=self, name='costs')
```
jlaura: The idea here is that these are arbitrary costs associated with each match. We can use these to weight measures through the match.

autocnet/graph/edge.py
```diff
@@ -704,8 +707,269 @@ def get_match_coordinates(self, clean_keys=[]):
...
+                costs = row.pop('_costs', {})
+                for k, v in row.items():
+                    if np.isnan(v):
+                        v = None
+                    costs[k] = v
+                cost = Costs(match_id=idx, _cost=costs)
+                to_db_add.append(cost)
+        if to_db_add:
+            session.bulk_save_objects(to_db_add)
+        session.commit()
```
jlaura: The update logic is a PITA, but this allows for arbitrary updates to the data frame.

autocnet/graph/edge.py
```diff
@@ -704,8 +707,269 @@ def get_match_coordinates(self, clean_keys=[]):
...
+        qf = q.filter(Matches.source == self.source['node_id'],
+                      Matches.destination == self.destination['node_id'])
+        odf = pd.read_sql(qf.statement, q.session.bind).set_index('id')
+        df = pd.DataFrame(odf.values, index=odf.index.values, columns=odf.columns.values)
+        df.index.name = 'id'
+        # Explicit close to get the session cleaned up
+        session.close()
+        return DbDataFrame(df,
+                           parent=self,
+                           name='matches')
```
jlaura: Literally just read matches from the DB.

autocnet/graph/edge.py
```diff
@@ -704,8 +707,269 @@ def get_match_coordinates(self, clean_keys=[]):
...
+               filter(Edges.destination == self.destination['node_id']).first()
+        if res:
+            res.ring = ring
+        else:
+            edge = Edges(source=self.source['node_id'],
+                         destination=self.destination['node_id'],
+                         ring=ring)
+            session.add(edge)
+            session.commit()
+        return
```
jlaura: These two support the ring matcher and propagating the ring information to the column in the db.

autocnet/graph/edge.py
```diff
@@ -704,8 +707,269 @@ def get_match_coordinates(self, clean_keys=[]):
...
+               filter(Edges.source == self.source['node_id']).\
+               filter(Edges.destination == self.destination['node_id']).first()
+        if res:
+            res.fundamental = v
+        else:
+            edge = Edges(source=self.source['node_id'],
+                         destination=self.destination['node_id'],
+                         fundamental = v)
+            session.add(edge)
+            session.commit()
```
jlaura: Another setter into the DB.

autocnet/graph/edge.py
```diff
@@ -704,8 +707,269 @@ def get_match_coordinates(self, clean_keys=[]):
...
+        ecef = pyproj.Proj(proj='geocent',
+			               a=self.parent.config['spatial']['semimajor_rad'],
+			               b=self.parent.config['spatial']['semiminor_rad'])
+        lla = pyproj.Proj(proj='longlat',
+			              a=self.parent.config['spatial']['semiminor_rad'],
+			              b=self.parent.config['spatial']['semimajor_rad'])
+        lons, lats, alts = pyproj.transform(ecef, lla, kps.xm.values, kps.ym.values, kps.zm.values)
+        points = [Point(lons[i], lats[i]) for i in range(len(lons))]
+        mask = [i for i in range(len(points)) if self.intersection.contains(points[i])]
+        return mask
```
jlaura: Helper to identify which polygons a particular get of keypoints intersect.

autocnet/graph/network.py
```diff
@@ -750,11 +763,12 @@ def serials(self):
             serials[n] = generate_serial_number(node['image_path'])
         return serials
 
+    @property
     def files(self):
         """
         Return a list of all full file PATHs in the CandidateGraph
         """
-        return [node['image_path'] for node in self.nodes]
+        return [node['image_path'] for _, node in self.nodes(data='data')]
```
jlaura: This is a bug fix for nx2.x

autocnet/graph/network.py
```diff
@@ -965,7 +977,7 @@ def compute_weight(self, clean_keys, **kwargs):  # pragma: no cover
                 ['x', 'y']]
             reproj_geom = source_node.reproject_geom(
                 overlap.geometry.values[0].__geo_interface__['coordinates'][0])
-            initial_mask = geom_mask(kps, reproj_geom)
+            initial_mask = cg.geom_mask(kps, reproj_geom)
```
jlaura: Just trying to be explicit

autocnet/graph/network.py
```diff
@@ -1270,3 +1282,269 @@ def to_bal(self):
         http://grail.cs.washington.edu/projects/bal/
         """
         pass
+
+class NetworkCandidateGraph(CandidateGraph):
+    node_factory = NetworkNode
+    edge_factory = NetworkEdge
```
jlaura: Standard overrides on the node/edge factories. These need to be the network variants now.

autocnet/graph/network.py
```diff
@@ -1270,3 +1282,269 @@ def to_bal(self):
...
+        if config.get('redis', None):
+            self._setup_queues()
+        self._setup_db_connection()
+        # Job metadata
+        self.job_status = defaultdict(dict)
+
+        for i, d in self.nodes(data='data'):
+            d.parent = self
+        for s, d, e in self.edges(data='data'):
+            e.parent = self
```
jlaura: This is a hack method to have the node/edge know about the graph and vice-versa. If anyone has a better solution that would be awesome!

autocnet/graph/network.py
```diff
@@ -1270,3 +1282,269 @@ def to_bal(self):
...
+        job_counter += 1
+
+        # Submit the jobs
+        submitter = Slurm('acn_submit',
+                     mem_per_cpu=config['cluster']['processing_memory'],
+                     time=walltime,
+                     partition=config['cluster']['queue'],
+                     output=config['cluster']['cluster_log_dir']+'/slurm-%A_%a.out')
+        submitter.submit(array='1-{}'.format(job_counter))
+        return job_counter
```
jlaura: This is an extension to the apply method that now submits the job to the cluster instead of a local apply. We can still call specific methods on the candidate graph and they will be performed locally. Only apply is wired to go to the cluster.

autocnet/graph/network.py
```diff
@@ -1270,3 +1282,303 @@ def to_bal(self):
...
+        Set up a database connection and session(s)
+        """
+        try:
+            Base.metadata.bind = engine
+            # If the table does not exist, this will create it. This is used in case a
+            # user has manually dropped a table so that the project is not wrecked.
+            Base.metadata.create_all(tables=[Network.__table__, Overlay.__table__,
+                                             Edges.__table__, Costs.__table__, Matches.__table__,
+                                             Cameras.__table__, Points.__table__,
+                                             Measures.__table__])
```
jlaura: This gets the tables created if they do not already exist.

autocnet/matcher/subpixel.py
```diff
@@ -261,11 +265,12 @@ def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=251, reduction=11, conver
     # Apply the shift to d_search and compute the new correspondence location
     dx += (shift_x + dxr)
     dy += (shift_y + dyr)
+    
     # Break if the solution has converged
     size -= reduction
-    if size < 1:
-        return None, None, None
-    elif abs(shift_x) < convergence_threshold and abs(shift_y) < convergence_threshold:
+    if abs(shift_x) <= convergence_threshold and abs(shift_y) <= convergence_threshold:
```
jlaura: This is a bug fix for the <= vs <

autocnet/__init__.py
Kelvinrr: Is there an example config file or schema somewhere?

autocnet/graph/edge.py
Kelvinrr: I personally don't have an issue with colocating a large number of entities in a single module overall. But, what exactly do you mean by object vs Netwrokobject?

autocnet/__init__.py
jlaura: Added on your PR.

autocnet/graph/edge.py
jlaura: exactly

PR #277, opened 2019-05-24
Added mosaic match code 

Kelvinrr:
Uses #276 as a base 

notebook: https://gist.github.com/Kelvinrr/f7fb0dd683302077ac34f196b1fe4cf8


jlaura:
:+1: I'll review this now.

jlaura:
Should I get #276 cleaned and merged or are you okay with getting this merged as an omnibus PR? If okay to just merge this one, can you add me as as a collaborator so I can push to this branch directly?

Kelvinrr:
Added you as a collaborator, let me know if you can push directly. I imagine it's best to clean this up since it'll auto-merge #276

autocnet/spatial/overlap.py
```diff
@@ -1,6 +1,6 @@
 import warnings
 from autocnet import config
-from autocnet.cg import cg
+from autocnet.cg import cg as compgeom
```
Kelvinrr: Oooooh, so that's what CG stands for. This thing always confused be because I thought it was short for Candidate Graph or something.

autocnet/matcher/subpixel.py
```diff
@@ -261,11 +267,105 @@ def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=251, reduction=11, conver
...
+    b = mosaic.metadata["IsisCube"]["Mapping"]["PolarRadius"]
+    ecef = pyproj.Proj(proj='geocent', a=a, b=b)
+    lla = pyproj.Proj(proj='latlon', a=a, b=b)
+    lons, lats, alts = pyproj.transform(ecef, lla, np.asarray(cnet["adjustedX"]), np.asarray(cnet["adjustedY"]), np.asarray(cnet["adjustedZ"]))
+    gdf = gpd.GeoDataFrame(cnet, geometry=geopandas.points_from_xy(lons, lats))
+    points = gdf.geometry
+
+
+    # get footprint
+    image_arr = cube.read_array()
```
jlaura: Can we pass the footprint into this func? That way we can use it with a non-ISIS cube, e.g., grab the footprint from the database.

autocnet/vis/graph_view.py
```diff
@@ -272,36 +269,37 @@ def plot_edge(edge, ax=None, clean_keys=[], image_space=100, downsampling=1,
...
+            color = line_kwargs['color']
+            line_kwargs.pop('color', None)
+
+        s_kps = s_kps[['x', 'y']].values
+        d_kps = d_kps[['x', 'y']].values
+        d_kps[:, 0] += x_offset
+
+        for l in zip(s_kps, d_kps):
+            ax.plot((l[0][0], l[1][0]), (l[0][1], l[1][1]), color=color, **line_kwargs)
+
```
jlaura: :+1:

autocnet/spatial/overlap.py
jlaura: Yeah..so it is. Perhaps rename cg (computational geometry) to something else?

autocnet/matcher/subpixel.py
Kelvinrr: I don't see why not.

autocnet/spatial/overlap.py
Kelvinrr: probably just computational_geometry.py

jlaura: I am not going to do this in this PR so this thing doesn't get even larger...I'll get an issue opened though so we don't forget.

autocnet/matcher/subpixel.py
jlaura: I got to thinking on this function. I think I would like to merge as is; we know it is working. Then refactor if desired. 

Question - does this method require a projected cube? I assume it does, but want to confirm. If so, we are going to need to then back the projection back out to get into image space. I think, let's merge and then we can worry about how to get the data we need from this for a network.

jlaura:
I am working through the CI issues now. I am also ignoring OS X since the issue is not us.

autocnet/matcher/subpixel.py
Kelvinrr: So, in this case, I used projected images because I was matching against a projected mosaic. Do we really need to project back into image space? I thought the ground points were the only thing that really mattered?

PR #280, opened 2019-06-03
Changes overlap to sql only

jlaura:


autocnet/io/db/model.py
```diff
@@ -218,7 +219,7 @@ def __repr__(self):
 class Overlay(Base):
     __tablename__ = 'overlay'
     id = Column(Integer, primary_key=True, autoincrement=True)
-    overlaps = Column(ARRAY(Integer))
+    intersections = Column(ARRAY(Integer))
```
Kelvinrr: So I imagine in the new process the intersections will simply be stored in the Overlay table. Where would this get populated by manually running the compute_overlapping_geoms.sql script?

jlaura: yup - fire up psql or use pgadmin4.  For ~2.8k images this takes a while so this way it can be totally asynchronous.

PR #287, opened 2019-06-04
Clsub

jlaura:
I am going to update this as I work a little on the point stuff. 

- Initial commit updates the setup.py to actually install the script. Then the cluster jobs will be able to find the script.


jlaura:
Crud - I need to get the other commits merged in properly.

PR #289, opened 2019-06-04
installs bin script

jlaura:


jessemapel:
I'm not sure why CI is failing on those tests.

jlaura:
I have a testing branch going trying to get this working. I think the issue is one of the packages that installs after openCV. This is then causing opencv to downgrade or come from another channel. The result is that the opencv so is then mismatched.  This accounts for the two cv2 errors. I think that the save/load error might be a side effect issue.

PR #290, opened 2019-06-04
fixes footprint loads

jlaura:
Fix to have the footprint get pushed


jlaura:
@ladoramkershner I think that this can be closed. The code is working as expected. I am going to close, but let's reopen if we encounter the issue again.

PR #293, opened 2019-06-04
Performance improvement in sql overlap

jlaura:
This adds an ST_INTERSECT query to the inner query as a performance improvement. I believe that the query planner should use that WHERE statement first and then perform the actual intersections. This should take the inner query from best case n^2 to worst case n^2.

### DIFF
--- a/sql/compute_overlapping_geometries.sql
+++ b/sql/compute_overlapping_geometries.sql
@@ -15,7 +15,7 @@ FROM (
 		SELECT overlay.id, array_agg(images.id) as iid
 		FROM overlay, images
 		WHERE images.footprint_latlon is NOT NULL AND
-		WHERE ST_INTERSECTS(overlay.geom, images.footprint_latlon) AND
+		ST_INTERSECTS(overlay.geom, images.footprint_latlon) AND
 		ST_AREA(ST_INTERSECTION(overlay.geom, images.footprint_latlon)) > 0.000001
 		GROUP BY overlay.id
 	) AS imgs
--- a/sql/compute_overlapping_geometries.sql
+++ b/sql/compute_overlapping_geometries.sql
@@ -15,6 +15,7 @@ FROM (
 		SELECT overlay.id, array_agg(images.id) as iid
 		FROM overlay, images
 		WHERE images.footprint_latlon is NOT NULL AND
+		WHERE ST_INTERSECTS(overlay.geom, images.footprint_latlon) AND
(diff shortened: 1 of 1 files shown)


jlaura:
Here is the query plan and execution time for the changed version:

```sql
"Update on "overlay"  (cost=20801.06..27638.86 rows=35649 width=312) (actual time=61443.625..61443.625 rows=0 loops=1)"
"  ->  Merge Join  (cost=20801.06..27638.86 rows=35649 width=312) (actual time=60899.261..61030.779 rows=35666 loops=1)"
"        Merge Cond: (imgs.id = "overlay".id)"
"        ->  Subquery Scan on imgs  (cost=20800.77..21870.24 rows=35649 width=96) (actual time=60897.376..61002.517 rows=35666 loops=1)"
"              ->  GroupAggregate  (cost=20800.77..21513.75 rows=35649 width=36) (actual time=60897.367..60993.042 rows=35666 loops=1)"
"                    Group Key: overlay_1.id"
"                    ->  Sort  (cost=20800.77..20889.89 rows=35649 width=8) (actual time=60897.356..60943.613 rows=221272 loops=1)"
"                          Sort Key: overlay_1.id"
"                          Sort Method: external merge  Disk: 3888kB"
"                          ->  Nested Loop  (cost=0.28..18105.42 rows=35649 width=8) (actual time=3.790..60672.119 rows=221272 loops=1)"
"                                ->  Seq Scan on images  (cost=0.00..1309.18 rows=2878 width=644) (actual time=1.524..5.376 rows=2878 loops=1)"
"                                      Filter: (footprint_latlon IS NOT NULL)"
"                                      Rows Removed by Filter: 40"
"                                ->  Index Scan using idx_overlay_geom on "overlay" overlay_1  (cost=0.28..5.83 rows=1 width=214) (actual time=1.614..21.046 rows=77 loops=2878)"
"                                      Index Cond: (geom && images.footprint_latlon)"
"                                      Filter: ((st_area(st_intersection(geom, images.footprint_latlon)) > '1e-06'::double precision) AND _st_intersects(geom, images.footprint_latlon))"
"                                      Rows Removed by Filter: 115"
"        ->  Index Scan using overlay_pkey on "overlay"  (cost=0.29..5232.74 rows=36108 width=220) (actual time=1.874..14.007 rows=36108 loops=1)"
"Planning time: 8.569 ms"
"Execution time: 61444.710 ms"
```

jlaura:
The other query has been running for > 20 minutes...I will let it finish, but the above is significantly faster.

jlaura:
After 2 hours, the connection to the DB was lost, so yeah - the speedup is real...

PR #297, opened 2019-06-04
cluster overlap point seeding

jessemapel:
This adds a function to seed points on overlaps via the cluster and a simple submission script.


autocnet/spatial/overlap.py
```diff
@@ -1,12 +1,14 @@
...
+from autocnet import config, Session, engine
 from autocnet.cg import cg as compgeom
+from autocnet.graph.node import NetworkNode
 from autocnet.io.db.model import Images, Measures, Overlay, Points
 from autocnet.matcher.subpixel import iterative_phase
-from autocnet import Session, engine
 
+import json
+from redis import StrictRedis
+from plurmy import Slurm
```
jessemapel: The imports on this file are a bit of a mess, is there a standard format or ordering to use?

autocnet/spatial/overlap.py
```diff
@@ -103,3 +103,127 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
...
+                point.measures.append(Measures(sample=dx,
+                                               line=dy,
+                                               imageid=destination['node_id'],
+                                               serial=destination.isis_serial,
+                                               measuretype=3))
+        if len(point.measures) >= 2:
+            points.append(point)
+    session.add_all(points)
+    session.commit()
+    session.close()
```
jessemapel: Should I modify place_points_in_overlaps to use this logic instead of the duplicated code? This is slightly different in that is queries the DB instead of using an actual candidate graph object.

autocnet/spatial/overlap.py
acpaquette: Generally in autocnet we have done python standard library imports like os, warnings, etc. followed by 3rd party libraries like redis, pyproj, etc. and finally any of our own libraries like autocnet, plurmy, plio, etc.

autocnet/spatial/overlap.py
```diff
@@ -103,3 +102,124 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
...
+
+    iterative_phase_kwargs : dict
+        Dictionary of keyword arguments for the iterative phase matcher function
+
+    walltime : str
+        Cluster job wall time as a string HH:MM:SS
+    """
+    # Get all of the overlaps over the size threshold
+    session = Session()
+    overlaps = session.query(Overlay.id, Overlay.geom, Overlay.intersections).\
```
acpaquette: Not sure how much it matters but this session should probably be closed.

bin/acn_overlaps
```diff
@@ -8,19 +8,46 @@ import sys
...
 if __name__ == '__main__':
-    main()
+    conf = config['redis']
+    queue = StrictRedis(host=conf['host'], port=conf['port'], db=0)
+
+    msg = pop_computetime_push(queue,
+                               conf['processing_queue'],
+                               conf['working_queue'])
+
+    response = main(msg, config)
```
acpaquette: Response is never populated by main, will always return None. I'm not sure if we even need this response. It's a hold over from the redis queue passing back completed jobs.

bin/acn_overlaps
```diff
@@ -8,19 +8,46 @@ import sys
...
+    if res is None:
+        print('Could not find overlap with ID', id)
+        sys.exit(1)
+    overlap = res.first()
+    geom = overlap.geom
+    nodes = []
+    for id in overlap.intersections:
+        res = session.query(Images).filter(Images.id == id).first()
+        nodes.append(NetworkNode(node_id=id, image_path=res.path))
+    print('Placing points in overlap', id)
```
acpaquette: Not sure if we want to leave these prints in here

autocnet/spatial/tests/test_overlap_points.py
```diff
@@ -0,0 +1,76 @@
...
+        Dummy method so that node['node_id'] works
+        """
+        if key == 'node_id'
+            return self.id
+
+    def isis_serial(self):
+        return str(self.id)
+
+    def geodata(self):
+        return GeoDataset()
```
acpaquette: I'm not sure if we want to do these dummy classes or just mock the attributes with Mock. I think mock would be preferable, probably (not 100% sure) less code to maintain

autocnet/spatial/overlap.py
```diff
@@ -103,3 +102,124 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
...
+
+    # Push the job messages onto the queue
+    queuename = config['redis']['processing_queue']
+    for overlap in overlaps:
+        msg = {'id' : overlap.id,
+               'height' : height,
+               'iterative_phase_kwargs' : iterative_phase_kwargs,
+               'walltime' : walltime}
+        rqueue.rpush(queuename, json.dumps(msg))
+    job_counter = len([*overlaps]) + 1
```
acpaquette: Might just want to use an enumeration and save off the last value +1 but I think this works just as well

bin/acn_overlaps
Kelvinrr: Some logging would help (presumably this would be written out to logs). But maybe logger instead of prints?

jessemapel: The prints will be logged to the slurm logs, so I'd like to keep them in here.

jessemapel:
CI on this is passing now, but it's not quite done yet

autocnet/spatial/tests/test_overlap_points.py
```diff
@@ -0,0 +1,55 @@
...
+    third_node = MagicMock()
+    third_node.camera = MagicMock()
+    third_node.camera.groundToImage.return_value = csmapi.ImageCoord(0.0, 1.0)
+    third_node.isis_serial = '3'
+    third_node.__getitem__.return_value = 3
+    fourth_node = MagicMock()
+    fourth_node.camera = MagicMock()
+    fourth_node.camera.groundToImage.return_value = csmapi.ImageCoord(0.0, 0.0)
+    fourth_node.isis_serial = '4'
+    fourth_node.__getitem__.return_value = 4
```
jessemapel: I changed this away from being a fixture because the patches the fixtures do not play nicely together

autocnet/spatial/overlap.py
jessemapel: To me this approach makes more sense. Having an enumeration that isn't used in the actual loop is strange.

Kelvinrr:
Looks good

PR #298, opened 2019-06-05
Updates with context on what is happening

jlaura:


notebooks/server/Demo.ipynb
```diff
@@ -9,18 +25,30 @@
    "outputs": [],
    "source": [
     "import os\n",
-    "os.environ['autocnet_config'] = '/home/jlaura/autocnet_projects/elysium.yml'\n",
+    "os.environ['autocnet_config'] = 'config/sample.yml'\n",
```
ladoramkershner: Should we add line of code here, such as
os.environ['PROJ_LIB'] = '/home/ladoramkershner/miniconda3/envs/autocnet_local/share/proj'

that can point to your location, but serve as a reminder that, in order to run on jupyterhub, this variable needs to be set?

jlaura: I think doing this in a follow on PR makes a ton of sense.

PR #302, opened 2019-06-05
Cnet match first edition 

Kelvinrr:
Will need changes to plio to successfully write the returned cnet to disk.


autocnet/io/db/model.py
```diff
@@ -20,7 +20,17 @@
...
+spatial = config.get('spatial', {'srid': 949900})
+srid = spatial['srid']
+
+class BaseMixin(object):
+    @classmethod
+    def create(cls, session, **kw):
+        obj = cls(**kw)
+        session.add(obj)
+        session.commit()
+        return obj
```
Kelvinrr: I did a git reset --hard then added my changes, why are these other changes here? Was something merged then reverted?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+                                          dbconfig["name"])
+
+    engine = create_engine(db_uri, poolclass=pool.NullPool,
+                           isolation_level="AUTOCOMMIT")
+
+    ctx_images = gpd.GeoDataFrame.from_postgis("select * from images", engine, geom_col="footprint_latlon")
+
+    themis_cnet = from_isis(cnet)
+    # we need image path information for each measure
+    themis_cnet = themis_cnet.merge(themis_images, how='left', left_on="serialnumber", right_on="serial")
```
acpaquette: I'm not sure exactly why these two are being merged together, other than to associate footprints with images. Just seems like something that could be handled via the database.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+        resolutions.append(res["GroundPoint"]["LineResolution"].value)
+
+    themis_cnet["resolution"] = resolutions
+    themis_cnet = gpd.GeoDataFrame(themis_cnet, geometry=gpd.points_from_xy(lons, lats))
+
+    spatial_index = ctx_images.sindex
+
+
+    # We are going to iterate on points
+    groups = themis_cnet.groupby("id_x").groups
```
acpaquette: groups is a little vague, any particular reason to group by x values?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+
+
+    # append to list if images, mostly used for working with the network in python
+    # after this step, is this uncecceary outside of debugging? Maybe actually should return
+    # more info of where everything was sourced in the original DataFrames?
+    images = []
```
acpaquette: Not really sure what you mean with the comment in regards to more info in the "original" dataframes

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+
+    # append CNET info into structured Python list
+    ctx_constrained_net = []
+
+    # easily parrallelized on the cpoint level, dummy serial for now
+    for cpoint, indices in groups.items():
+        measures = themis_cnet.loc[indices]
+        measure = measures.iloc[0]
+        p = measure.geometry
+        possible_matches_index = list(spatial_index.intersection(p.bounds))
```
acpaquette: How much does intersecting on the spatial_index save computation wise? Just curious

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+                                       0,                      # line residual
+                                       ctx_sample,             # apriorisample
+                                       ctx_line,               # aprioriline
+                                       0,                      # sample sigma
+                                       0                       # line sigma
+                                       ])
+
+
+    # These should be defined somewhere in Autocnet/plio, if so it should be imported
+    columns = ['point_id', 'PointType', 'chooserName', 'datetime', 'editLock', 'ignore',
```
acpaquette: This is in kind of in plio. 
https://github.com/USGS-Astrogeology/plio/blob/10ef4a8245e53dcd2c413c85b824b1f513b5fc76/plio/io/io_controlnetwork.py#L215

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+
+    themis_images = gpd.GeoDataFrame.from_postgis("select * from themis_ir", engine, geom_col="footprint_latlon")
+
+    # useful for masking out paths, sometimes data is split between shared directories and
+    # scratch causing errors when working anywhere besides the custer
+    themis_images["valid_path"] = [os.path.isfile(p) for p in themis_images["path"]]
+    themis_images = themis_images[themis_images["valid_path"]]
+
+    # We need to join the CNET dataframe on serial numbers
+    themis_images["serial"] = [plio.io.isis_serial_number.generate_serial_number(d) if d else None for d in themis_images["path"]]
```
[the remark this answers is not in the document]
Kelvinrr: Last I checked the themis specific Database does not have that column. Can't double check remotely but definitely the case while I was writing this.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: File paths are what I need, the CNET does not have image paths for measures. Isis uses a separate image list which would be an alternative, but it's better to figure out the image situation across instruments and have serials there.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: `from_isis` sets id as the point id, during the merge the two DFs have id columns so renames them id_x id_y with id_x being the point id. The comment implies the groups as point ids as that his ow control points are handled in our tables. Not worth doing much here since I see these tables changing sooner than later.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: The data frames obtained at the start of the function. Merging cnet with image list and there isn't much info other than IDs to reverse the extra data in the CNET and image list if the user cares.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: Depends on size, never quantified it but was a significant difference at larger scales with hundreds of points.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+    Hardcoded to be named exactly what it does, if we get meaningful results should
+    be simple to generalize to a cross instrument CNET matcher type thing.
+    """
+
+    # Get neccessary tables from DB,
+    # later this should probably be done through SQL comamnds and joins rather than in DataFrames
+
+    # Get ground images from database, in this case THEMIS
+    # hardcoded for now until we can finalize our databases
+    db_uri = '{}://{}:{}@{}:{}/{}'.format('postgresql',
```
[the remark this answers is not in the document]
Kelvinrr: This one is too particular, doesn't use the config file since it's a completely different database. The project's engine object should be imported for the second connection, though.

autocnet/matcher/cross_instrument_matcher.py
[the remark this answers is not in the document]
jlaura: `ALTER TABLE <name> ADD COLUMN <columnname> <columntype>` - Let's just add it.

autocnet/matcher/cross_instrument_matcher.py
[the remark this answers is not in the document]
acpaquette: Could setup two configs and that could be used for either of the connections

autocnet/io/db/model.py
jessemapel: The blame says this commit https://github.com/USGS-Astrogeology/autocnet/commit/4262c9ef496ff45e046988441e2e1b467241f06f

autocnet/matcher/cross_instrument_matcher.py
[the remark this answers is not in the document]
jessemapel: Could we specify in the config file that we want to use different databases for different datasets?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+        # lazily iterate for now
+        for i,row in precise_matches.iterrows():
+            try:
+                res = pvl.loads(campt(from_=row["path"],
+                                      longitude=p.x,latitude=p.y,
+                                      type="ground", allowoutside=False))
+            except ProcessError as e:
+                continue
+
+            ctx_line, ctx_sample = res["GroundPoint"]["Line"], res["GroundPoint"]["Sample"]
```
jessemapel: You probably need to convert isis pixels to CSM pixels here, if the ISIS pixel is (line ,sample), then the CSM pixel is (line-0.5, sample-0.5)

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+            scaled_ctx_line = (ctx_arr.shape[0]-ctx_line)*ctx_to_themis_scale
+            scaled_ctx_sample = ctx_sample*ctx_to_themis_scale
+
+            ctx_arr = imresize(ctx_arr, ctx_to_themis_scale)[::-1]
+
+            # list of matching results in the format:
+            # [measure_index, x_offset, y_offset, offset_magnitude]
+            match_results = []
+            for k,m in measures.iterrows():
+                themis_arr = GeoDataset(m["path"]).read_array()
```
jessemapel: This is going to get called on each image for each measure in the image. Is there concerns about calling this a whole bunch? Could it be called once on each image and then re-used?

Kelvinrr:
Not really worried about testing until it becomes more generalized. No point test something that we are confident is gonna change a lot in the coming days.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: I also need the footprint information, at that point, I'd just be rewriting merge.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: Originally just had a GeoDataset lookup, not sure why I changed it back.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: what's the difference between point and measure attrs?

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: I'm not using CSM at all here, so does it matter?

autocnet/matcher/cross_instrument_matcher.py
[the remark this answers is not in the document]
Kelvinrr: there should be a solution that doesn't cause the config to become too bloated. Not sure what that looks like at the moment.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+                                          dbconfig["username"],
+                                          dbconfig["password"],
+                                          dbconfig["host"],
+                                          dbconfig["port"],
+                                          dbconfig["name"])
+
+    engine = create_engine(db_uri, poolclass=pool.NullPool,
+                           isolation_level="AUTOCOMMIT")
+
+    ctx_images = gpd.GeoDataFrame.from_postgis("select * from images", engine, geom_col="footprint_latlon")
```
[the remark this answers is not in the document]
Kelvinrr: Geopandas is easier to work with especially with doing point - footprint intersections and while I still need to do merges.

autocnet/matcher/cross_instrument_matcher.py
jlaura: Same, let's merge and parallelize the refactor on this?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+from autocnet.matcher.subpixel import iterative_phase
+from autocnet import engine
+from gdal import ogr
+from geoalchemy2 import Geometry, WKTElement
+from geoalchemy2.shape import to_shape
+from knoten import csm
+from plio.io.io_controlnetwork import from_isis, to_isis
+from plio.io.io_gdal import GeoDataset
+from pysis.exceptions import ProcessError
+from pysis.isis import campt
```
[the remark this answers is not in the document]
Kelvinrr: Totally just copy pasted a bunch of imports.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+    db_uri = '{}://{}:{}@{}:{}/{}'.format('postgresql',
+                                          'jay',
+                                          'abcde',
+                                          'smalls',
+                                          '8083',
+                                          'mars')
+    engine = create_engine(db_uri, poolclass=pool.NullPool,
+                           isolation_level="AUTOCOMMIT")
+
+    themis_images = gpd.GeoDataFrame.from_postgis("select * from themis_ir", engine, geom_col="footprint_latlon")
```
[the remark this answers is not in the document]
Kelvinrr: yeah, currently it's not a lot of images.

autocnet/matcher/cross_instrument_matcher.py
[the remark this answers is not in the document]
Kelvinrr: we already agreed to that in a different comment. Jesse seems to be asking why load to a data frame with from_postgis.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+        possible_matches = ctx_images.iloc[possible_matches_index]
+        precise_matches = possible_matches[possible_matches.intersects(p)]
+
+        if precise_matches.empty:
+            continue
+
+        # lazily iterate for now
+        for i,row in precise_matches.iterrows():
+            try:
+                res = pvl.loads(campt(from_=row["path"],
```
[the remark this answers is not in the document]
Kelvinrr: you have a comment on the line

Kelvinrr:
Just gonna merge as is, the database concerns need to be addressed but it seems more of a question of what to do with the database(s) when working across instruments and are out of the scope of this PR.

autocnet/matcher/cross_instrument_matcher.py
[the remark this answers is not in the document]
Kelvinrr: Still need to be populated. Not really sure what the bigger picture purpose of this table is supposed to be. Right now it only exists for this. But it was made to hold all themis_ir stuff? Is it like a master repository of themis IR images?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -0,0 +1,255 @@
...
+                        LargeBinary, String, UniqueConstraint, create_engine,
+                        event, orm, pool)
+from sqlalchemy.ext.declarative import declarative_base
+
+import geoalchemy2
+import geopandas as gpd
+import plio
+import pvl
+import pyproj
+import pysis
```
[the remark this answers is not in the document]
Kelvinrr: don't we still need CTX support?

PR #303, opened 2019-06-06
Abstracts out geoalchemy to single place

jlaura:
This PR does a few things related to the DB. I want to also run the demo.ipynb notebook with these changes in a clean DB to see I have broken anything that is not being well tested at the moment. Changes are:

- Remove GeoAlchemy2 from all locations except the models
- Add setters/getters to the models to support shapely<->geoalchemy2 (to support above)
- Adds more db tests

### DIFF
--- a/autocnet/graph/edge.py
+++ b/autocnet/graph/edge.py
@@ -883,6 +883,8 @@ class NetworkEdge(Edge):
                         row_val = int(row_val)
                     elif isinstance(row_val, (np.float,)):
                         row_val = float(row_val)
+                    # This should be uncommented if the matches
+                    # df is refactored to be a geodataframe
                     #elif isinstance(row_val, WKBElement):
                     #    continue
                     mapping[index] = row_val
--- a/autocnet/graph/edge.py
+++ b/autocnet/graph/edge.py
@@ -2,7 +2,6 @@ from functools import wraps, singledispatch
 import warnings
 from collections import defaultdict, MutableMapping, Counter
 
-from geoalchemy2.elements import WKBElement
 import numpy as np
 import pandas as pd
 import networkx as nx
@@ -884,8 +883,8 @@ class NetworkEdge(Edge):
                         row_val = int(row_val)
                     elif isinstance(row_val, (np.float,)):
                         row_val = float(row_val)
-                    elif isinstance(row_val, WKBElement):
-                        continue
+                    #elif isinstance(row_val, WKBElement):
+                    #    continue
                     mapping[index] = row_val
                 to_db_update.append(mapping)
             else:
--- a/autocnet/graph/node.py
+++ b/autocnet/graph/node.py
@@ -4,7 +4,6 @@ import os
 import warnings
 
 from csmapi import csmapi
-import geoalchemy2
 import numpy as np
 import pandas as pd
 from plio.io.io_gdal import GeoDataset
@@ -650,11 +649,8 @@ class NetworkNode(Node):
             boundary = generate_boundary(self.geodata.raster_size[::-1])  # yx to xy
             footprint_latlon = generate_latlon_footprint(self.camera, boundary)
             footprint_latlon.FlattenTo2D()
-            footprint_latlon = footprint_latlon.ExportToWkt()
-            footprint_latlon = geoalchemy2.elements.WKTElement(footprint_latlon,
-                                                               srid=config['spatial']['srid'])
         else:
-            footprint_latlon = geoalchemy2.shape.to_shape(res.footprint_latlon)
+            footprint_latlon = res.footprint_latlon
         return footprint_latlon
 
     @property
--- a/autocnet/io/db/model.py
+++ b/autocnet/io/db/model.py
@@ -14,7 +14,7 @@ from sqlalchemy_utils import database_exists, create_database
 from sqlalchemy.types import TypeDecorator
 
 from geoalchemy2 import Geometry
-from geoalchemy2.shape import to_shape
+from geoalchemy2.shape import from_shape, to_shape
 
 from autocnet import engine, Session, config
 
@@ -151,7 +151,7 @@ class Matches(BaseMixin, Base):
     destination_idx = Column(Integer, nullable=False)
     lat = Column(Float)
     lon = Column(Float)
-    geom = Column(Geometry('POINT', dimension=2, srid=srid, spatial_index=True))
+    _geom = Column("geom", Geometry('POINT', dimension=2, srid=srid, spatial_index=True))
     source_x = Column(Float)
     source_y = Column(Float)
     destination_x = Column(Float)
@@ -161,6 +161,13 @@ class Matches(BaseMixin, Base):
     original_destination_x = Column(Float)
     original_destination_y = Column(Float)
 
+    @property
+    def geom(self):
+        return to_shape(self._geom)
+
+    @geom.setter
+    def geom(self, geom):
+        self._geom = from_shape(geom, srid=srid)
 
 class Cameras(BaseMixin, Base):
     __tablename__ = 'cameras'
@@ -176,7 +183,7 @@ class Images(BaseMixin, Base):
     path = Column(String)
     serial = Column(String, unique=True)
     active = Column(Boolean)
-    footprint_latlon = Column(Geometry('MultiPolygon', srid=srid, dimension=2, spatial_index=True))
+    _footprint_latlon = Column("footprint_latlon", Geometry('MultiPolygon', srid=srid, dimension=2, spatial_index=True))
     footprint_bodyfixed = Column(Geometry('MULTIPOLYGON', dimension=2))
     #footprint_bodyfixed = Column(Geometry('POLYGON',dimension=3))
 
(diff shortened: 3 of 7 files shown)


autocnet/graph/edge.py
```diff
@@ -884,8 +883,8 @@ def matches(self, v):
                         row_val = int(row_val)
                     elif isinstance(row_val, (np.float,)):
                         row_val = float(row_val)
-                    elif isinstance(row_val, WKBElement):
-                        continue
+                    #elif isinstance(row_val, WKBElement):
+                    #    continue
```
jlaura: Intentionally commented out. Manual check on the matches looked good, but perhaps matches should be a geodata frame (and then this needs to be in).

autocnet/graph/node.py
```diff
@@ -650,11 +649,8 @@ def footprint(self):
             boundary = generate_boundary(self.geodata.raster_size[::-1])  # yx to xy
             footprint_latlon = generate_latlon_footprint(self.camera, boundary)
             footprint_latlon.FlattenTo2D()
-            footprint_latlon = footprint_latlon.ExportToWkt()
-            footprint_latlon = geoalchemy2.elements.WKTElement(footprint_latlon,
-                                                               srid=config['spatial']['srid'])
         else:
-            footprint_latlon = geoalchemy2.shape.to_shape(res.footprint_latlon)
+            footprint_latlon = res.footprint_latlon
```
jlaura: All the geoalchemy stuff is in the model now

autocnet/io/db/model.py
```diff
@@ -12,10 +12,13 @@
 from sqlalchemy.orm import relationship, backref
 from sqlalchemy_utils import database_exists, create_database
 from sqlalchemy.types import TypeDecorator
+from sqlalchemy.ext.hybrid import hybrid_property
```
jlaura: standard properties do not work when they define a DB column, so Sqlalchemy provides this decorator.

autocnet/io/db/model.py
```diff
@@ -12,10 +12,13 @@
...
 
 from geoalchemy2 import Geometry
-from geoalchemy2.shape import to_shape
+from geoalchemy2.shape import from_shape, to_shape
 
+import osgeo
+import shapely
```
jlaura: This is because knotted returns on our poly and we need to convert it into shapely (for conversion).

autocnet/io/db/model.py
```diff
@@ -149,7 +154,7 @@ class Matches(BaseMixin, Base):
     destination_idx = Column(Integer, nullable=False)
     lat = Column(Float)
     lon = Column(Float)
-    geom = Column(Geometry('POINT', dimension=2, srid=srid, spatial_index=True))
+    _geom = Column("geom", Geometry('POINT', dimension=2, srid=srid, spatial_index=True))
```
jlaura: This is the syntax that needs changing.

jlaura: for the properties

autocnet/spatial/overlap.py
```diff
@@ -1,14 +1,13 @@
 import warnings
 from autocnet import config
 from autocnet.cg import cg as compgeom
-from autocnet.io.db.model import Images, Measures, MeasureType, Overlay, Points, PointType
+from autocnet.io.db.model import Images, Measures, Overlay, Points
```
jlaura: Types are abstracted now into the model.py file.

autocnet/spatial/overlap.py
```diff
@@ -66,9 +65,8 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
         source = cg.node[source]['data']
         source_camera = source.camera
         for v in valid:
-            geom = f'srid={srid};Point({v[0]} {v[1]})'
-            point = Points(geom=geom,
-                           pointtype=PointType(2)) # Would be 3 or 4 for ground
+            point = Points(geom=shapely.geometry.Point(*v),
+                           pointtype=2) # Would be 3 or 4 for ground
```
jlaura: More instances of syntax like this can be refactored from the code base. I did not find them all.

jlaura:
@all this is ready for review.

jessemapel:
This looks good to me, but I am very weak when it comes to databases, so I would feel better if someone else checked off on this.

jlaura:
@jessemapel :+1:

@Kelvinrr @acpaquette Can you also take a look at this to make sure I am not introducing something unintended?

autocnet/io/db/model.py
```diff
@@ -159,6 +164,17 @@ class Matches(BaseMixin, Base):
     original_destination_x = Column(Float)
     original_destination_y = Column(Float)
 
+    @hybrid_property
+    def geom(self):
+        try:
+            return to_shape(self._geom)
+        except:
+            return self._geom
```
acpaquette: Do we want to support wtk style strings and geometry objects or just pick one of the two?

autocnet/io/db/model.py
```diff
@@ -226,6 +265,28 @@ class Points(BaseMixin, Base):
...
+
+    @geom.setter
+    def geom(self, geom):
+        if geom:
+            self._geom = from_shape(geom, srid=srid)
+
+    @hybrid_property
+    def pointtype(self):
+        return self._pointtype
```
acpaquette: Do we have to define this due to the weirdness between the column name and trying to use it as a property?

autocnet/io/db/model.py
acpaquette: Any immediate thoughts or just something to ruminate? Seems like a similar issue we were running into in ALE

autocnet/io/db/model.py
acpaquette: Knotted returns?

acpaquette:
@jlaura Made some general comments and asked a few questions. Overall looks decent :+1:

autocnet/io/db/model.py
jlaura: knoten autocorrected.

autocnet/io/db/model.py
jlaura: I think that this issue might be slightly different. Here we just need to make sure that the variable name that is a class property  is different from the instance property. This is not necessarily a public/private issue, but more about how the table description is parsed into the schema.

autocnet/io/db/model.py
jlaura: exactly. https://docs.sqlalchemy.org/en/13/orm/extensions/hybrid.html

autocnet/io/db/model.py
jlaura: I put this in because, when testing, I was seeing that the properties were sometimes being called with something that was not WKBElement or WKTElement. I think that this is on initial instantiation, but I am not 100% sure.

jlaura:
@acpaquette @jessemapel  Is this good to merge?

autocnet/io/db/model.py
```diff
@@ -159,6 +164,17 @@ class Matches(BaseMixin, Base):
     original_destination_x = Column(Float)
     original_destination_y = Column(Float)
 
+    @hybrid_property
```
Kelvinrr: Sorry, can you elaborate as to why these are hybrid properties? My understanding is this can have both self and cls passed in but what specifically is it trying to accomplish here?

autocnet/io/db/model.py
```diff
@@ -54,7 +57,9 @@ def __init__(self, enumtype, *args, **kwargs):
         self._enumtype = enumtype
 
     def process_bind_param(self, value, dialect):
-        return value.value
+        if hasattr(value, 'value'):
+            value = value.value
```
Kelvinrr: In what context would this not have the attr value? is it handling number vs unit problems or similar type problem?

autocnet/io/db/model.py
jlaura: If we use just `@property` then it is not possible to declare this as a class variable (above). The hybrid is allowing us to use properties while also declaring the table schema above. This is my understanding. I can report that using a standard property fails.

autocnet/io/db/model.py
jlaura: If it is none at table instantiation.

autocnet/io/db/model.py
jlaura: These are good docs: https://docs.sqlalchemy.org/en/13/orm/mapped_sql_expr.html

jlaura: And here for the why: https://docs.sqlalchemy.org/en/13/orm/mapped_attributes.html#using-descriptors-and-hybrids

jlaura: So the error I was seeing that led me to these docs was with using a standard property with a `_`private attribute in the model and then trying to execute a `session.Query(model.attr = 'foo'`. We need to the `.attr` to be both a standard property and still work inside the `Query` obj.

autocnet/graph/edge.py
jessemapel: I'd like to see a comment in line about when this code should be uncommented. Without that, it's unclear why this was left in but commented out.

autocnet/io/db/model.py
jessemapel: How does this impact packaging these into JSON messages for cluster dispatch? Will these need to be converted to WKT strings and then back?

autocnet/io/db/model.py
```diff
@@ -174,7 +190,7 @@ class Images(BaseMixin, Base):
     path = Column(String)
     serial = Column(String, unique=True)
     active = Column(Boolean)
-    footprint_latlon = Column(Geometry('MultiPolygon', srid=srid, dimension=2, spatial_index=True))
+    _footprint_latlon = Column("footprint_latlon", Geometry('MultiPolygon', srid=srid, dimension=2, spatial_index=True))
```
jessemapel: Why the new column name?

jessemapel: Is this being set elsewhere?

jlaura: Because of the setter below and the hybrid_property items.

autocnet/io/db/model.py
jlaura: This is going to be coming out as a shapely geom. For serialization, I add the following to the JsonEncoder:

```        
if isinstance(obj,  shapely.geometry.base.BaseGeometry):
    return obj.wkt
```

jlaura: PR inbound with these changes since we are live commenting.

jessemapel: Where is out encoder in this project and has that change already been made?

jlaura: @jessemapel updated with tests

jlaura: See model.py for the encoder change

Issue #305, opened 2019-06-06
Concurrent ComputeFundamentalMatrix fails


jessemapel:
When using `cng.apply(ComputeFundamentalMatrix, ...)`, there are failures due to concurrent updates in the `validate_points` trigger.

Sample output log for a failed job

```
/home/jmapel/.conda/envs/autocnet/lib/python3.6/site-packages/sqlalchemy/sql/functions.py:68: SAWarning: The GenericFunction 'array_agg' is already registered and is going to be overriden.
  "is going to be overriden.".format(identifier))
Traceback (most recent call last):
  File "/home/jmapel/.conda/envs/autocnet/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1244, in _execute_context
    cursor, statement, parameters, context
  File "/home/jmapel/.conda/envs/autocnet/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 550, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.InternalError_: tuple concurrently updated


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/jmapel/.conda/envs/autocnet/bin/acn_submit", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/work/users/jmapel/autocnet/bin/acn_submit", line 11, in <module>
    from autocnet.io.db import connection
  File "/work/users/jmapel/autocnet/autocnet/__init__.py", line 52, in <module>
    import autocnet.graph
  File "/work/users/jmapel/autocnet/autocnet/graph/__init__.py", line 1, in <module>
    from . import edge
  File "/work/users/jmapel/autocnet/autocnet/graph/edge.py", line 15, in <module>
    from autocnet.graph.node import Node
  File "/work/users/jmapel/autocnet/autocnet/graph/node.py", line 20, in <module>
    from autocnet.io.db.model import Images, Keypoints, Matches, Cameras,  Base, Overlay, Edges, Costs, Points, Measures
  File "/work/users/jmapel/autocnet/autocnet/io/db/model.py", line 277, in <module>
    Keypoints.__table__])
  File "/home/jmapel/.conda/envs/autocnet/lib/python3.6/site-packages/sqlalchemy/sql/schema.py", line 4287, in create_all
    ddl.SchemaGenerator, self, checkfirst=checkfirst, tables=tables
[...]

PR #306, opened 2019-06-06
Fixes 305

jlaura:
The two other files are me deprecating some tests. I was working on this when the issue came in and did not stash. I can pull those out if folks are worried. The tests are already turned off in dev.


acpaquette:
Should the deprecated integration tests be in this?

jessemapel:
@jlaura you should remove the deprecated tests, the actual change here looks good though

jlaura:
Once CI passes I am going to merged based on the review, comments on #305 and the fix for the other issue in #308.

PR #307, opened 2019-06-06
Added warning when json file cannot be written out

jessemapel:
This prevents failure when the JSON file cannot be written out. If there already exists a file, things can continue just fine, but if no file exists and one cannot be written, then the camera will throw an exception itself.

### DIFF
--- a/autocnet/graph/node.py
+++ b/autocnet/graph/node.py
@@ -620,8 +620,8 @@ class NetworkNode(Node):
         try:
             with open(isdpath, 'w') as f:
                 json.dump(response, f)
-        except:
-            warnings.warn('Failed to write JSON ISD for image {}'.format(self['image_path']))
+        except Exception as e:
+            warnings.warn('Failed to write JSON ISD for image {}.\n{}'.format(self['image_path'], e))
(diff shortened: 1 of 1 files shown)


autocnet/graph/node.py
```diff
@@ -617,13 +617,16 @@ def create_camera(self):
...
             return
         isdpath = os.path.splitext(self['image_path'])[0] + '.json'
-        with open(isdpath, 'w') as f:
-            json.dump(response, f)
-            isd = csmapi.Isd(self['image_path'])
+        try:
+            with open(isdpath, 'w') as f:
+                json.dump(response, f)
+        except:
+            warnings.warn('Failed to write JSON ISD for image {}'.format(self['image_path']))
```
acpaquette: Would it be worth making this a little more detailed? The usual error would be a problem with permissions but it would be nice to just know that rather than having to search for it

Kelvinrr: Capturing the Exception and adding it to the warning string would be good enough I imagine.

acpaquette: :+1:

PR #315, opened 2019-06-11
Added height calculation from DEM instead of using a hardcoded value.

kaitlyndlee:


autocnet/spatial/overlap.py
```diff
@@ -72,8 +69,14 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
             point = Points(geom=shapely.geometry.Point(*v),
                            pointtype=2) # Would be 3 or 4 for ground
 
+            # Calculate the height, the distance (in meters) above or 
+            # below the aeroid (meters above or below the BCBF spheroid).
+            gd = GeoDataset(dem_tiff)
```
acpaquette: Not sure how much it matters in terms of speed, but this could be moved outside of this for loop to only generate the geodata once.

autocnet/spatial/overlap.py
```diff
@@ -47,7 +43,8 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
     semi_minor = config['spatial']['semiminor_rad']
     ecef = pyproj.Proj(proj='geocent', a=semi_major, b=semi_minor)
     lla = pyproj.Proj(proj='latlon', a=semi_major, b=semi_minor)
-
+    dem_tiff = config['spatial']['dem_tiff']
```
jlaura: This is going to fail if we do not have a 'spatial' key. Consider:

```python
spatial = config.get('spatial', {})
dem_tiff = spatial.get('dem_tiff')
if dem_tiff is None:
    raise or return or warn or something
```

autocnet/spatial/overlap.py
```diff
@@ -72,8 +69,14 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
             point = Points(geom=shapely.geometry.Point(*v),
                            pointtype=2) # Would be 3 or 4 for ground
 
+            # Calculate the height, the distance (in meters) above or 
+            # below the aeroid (meters above or below the BCBF spheroid).
+            gd = GeoDataset(dem_tiff)
+            px, py = gd.latlon_to_pixel(v[1], v[0])
```
jlaura: Does it make sense to be more verbose here with `lon = v[1]`, `lat = v[0]`, lines and then to pass lon/lat in?

autocnet/spatial/overlap.py
```diff
@@ -72,8 +69,14 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
...
+            height = gd.read_array(1, [px, py, 1, 1])[0][0]
+
             # Get the BCEF coordinate from the lon, lat
-            x, y, z = pyproj.transform(lla, ecef, v[0], v[1], height)  # -3000 working well in elysium, need aeroid
+            x, y, z = pyproj.transform(lla, ecef, v[0], v[1], height)
```
jlaura: The above is v[1], v[0] and then this is v[0], v[1]. What is the correct ordering? Does it differ between these two calls. If so we should comment well.

notebooks/server/Demo.ipynb
```diff
@@ -125,9 +125,7 @@
     "from autocnet.spatial.overlap import place_points_in_overlaps\n",
     "\n",
     "# Place points\n",
-    "place_points_in_overlaps(ncg, height=-3000) # This value is good for elysium, but needs to be more granularly parameterizable\n",
-    "# A bad height value results in very poor results... The height is height above (below) the sphere (the aeroid).\n",
-    "# To generalize this, we would spawn a new cluster job for each geomety and pull a height dynamically from from reference"
+    "place_points_in_overlaps(ncg)"
```
jlaura: Can you check this back out and revert to get the Demo out of the PR?

jlaura:
Total side comment - in the demo, the elevation that is working is -3000. Is the SRS of the aeroid.tif in the same longitude domain as the images DB (-180, 180) I believe. I wonder if the aeroid isn't (0,360).

autocnet/spatial/overlap.py
jessemapel: Similarly, how does this work with testing?

config/sample.yml
```diff
@@ -45,6 +45,8 @@ spatial:
     semimajor_rad: 3396190  # in meters
     semiminor_rad: 3376200  # in meters
     proj4_str: '+proj:longlat +a:3396190 +b:3376200 +no_defs'
+    # Tiff used to generate height values for projection
```
jessemapel: We should have documentation of what all of these config values are. Something like conda-forge's [meta.yml](https://github.com/conda-forge/staged-recipes/blob/master/recipes/example/meta.yaml) documentation.

jessemapel: This is a potential issue, not work to be done under this PR

autocnet/spatial/overlap.py
kaitlyndlee: Would we want to add this check for the other values being pulled from the config, like srid?

autocnet/spatial/overlap.py
kaitlyndlee: I am actually a little confused about this. I was under the impression that v[0] was lat and v[1] was lon because of the transform method call, but calling latlon_to_pixel with v[0], v[1] order did not work and v[1], v[0] did, even though the plio documentation for the method: https://github.com/USGS-Astrogeology/plio/blob/10ef4a8245e53dcd2c413c85b824b1f513b5fc76/plio/io/io_gdal.py#L467 has order lat lon

kaitlyndlee:
Do we want the ability to pass in a height? I could then take the cube out of the config file and just have it be passed in as well?

config/sample.yml
jlaura: This is done in PR #318

jessemapel:
After talking to Jay, ignore what I said about height. We don't need it for now.

autocnet/spatial/overlap.py
```diff
@@ -68,12 +66,18 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
...
                            pointtype=2) # Would be 3 or 4 for ground
 
+            # Calculate the height, the distance (in meters) above or 
+            # below the aeroid (meters above or below the BCBF spheroid).
+            px, py = gd.latlon_to_pixel(v[1], v[0])
+            height = gd.read_array(1, [px, py, 1, 1])[0][0]
+
             # Get the BCEF coordinate from the lon, lat
-            x, y, z = pyproj.transform(lla, ecef, v[0], v[1], height)  # -3000 working well in elysium, need aeroid
+            x, y, z = pyproj.transform(lla, ecef, v[0], v[1], height)
```
jessemapel: The height parameter should be removed as it's unused now

autocnet/spatial/overlap.py
```diff
@@ -31,10 +31,6 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
     reference : int
                 the i.d. of a reference node to use when placing points. If not
                 speficied, this is the node with the lowest id
-
-    height : numeric
-             The distance (in meters) above or below the aeroid (meters above or
-             below the BCBF spheroid).
     """
```
jessemapel: There should be a note in the doc string that the DEM specified in the config file will be used to compute point elevations.

autocnet/spatial/overlap.py
kaitlyndlee: I am initializing height above with the GeoDataset read_array method. The parameter itself is removed

autocnet/spatial/overlap.py
kaitlyndlee: I have been thinking about the config file. Is this the only place that the DEM will be used? Does it make more sense to have the DEM passed in as a param instead?

autocnet/spatial/overlap.py
jessemapel: Yep my bad!

autocnet/spatial/overlap.py
jessemapel: The DEM is going to be used anywhere we have to do image to ground, so it makes sense to specify it in the config. Right now, the other location we're talking about is footprint generation.

jessemapel:
Once CI passes I think this is good to go. Then I'll update #297

PR #316, opened 2019-06-11
Db Fixes #309

### LINKED ISSUE #309: Write function to run overlap SQL queries
It's really annoying having to run these by hand, we should just have functions that can run them.

jlaura:
Fixes #309

Tested locally on the 'deleted' db since we do not have good test coverage for executing these items via the DB.


autocnet/graph/network.py
```diff
@@ -1297,11 +1298,14 @@ def __init__(self, *args, **kwargs):
         for s, d, e in self.edges(data='data'):
             e.parent = self
 
+        # Execute the computation to compute overlapping geometries
+        self._execute_sql(compute_overlaps_sql)
```
jlaura: A generic SQL submission func now exists on the NCG.

autocnet/graph/network.py
```diff
@@ -1320,6 +1324,18 @@ def empty_queues(self):
...
+        """
+        Execute a raw SQL string in the database currently specified
+        by the AutoCNet config file.
+
+        Use this method with caution as you can easily do things like
+        truncate a table.
+        """
+        conn = engine.connect()
+        conn.execute(sql)
+        conn.close()
```
jlaura: Pretty boring here. I need to update the doc string with parameters.

autocnet/graph/network.py
```diff
@@ -1421,10 +1437,10 @@ def generate_vrts(self, **kwargs):
         For the nodes in the graph, genreate a GDAL compliant vrt file.
         This is just a dispatcher to the knoten generate_vrt file.
         """
-        for i, n in self.nodes(data='data'):
+        for _, n in self.nodes(data='data'):
```
jlaura: Just cleanup for an unused variable in the linter.

autocnet/graph/network.py
```diff
@@ -1459,6 +1475,23 @@ def to_isis(self, path, flistpath=None,         sql = """
         cnet.to_isis(path, df, self.serials())
         cnet.write_filelist(self.files, path=flistpath)
 
+    @classmethod
```
jlaura: This is an NCG from_filelist implementation. The initial if/else block is not great. The two other ways I see to implemented this are (1) using multiple dispatch on the file list or (2) using a try/except. It felt like this version was the most readable. If the filets is an utterable, just pass. Else pass to plio to open the file.

autocnet/graph/network.py
```diff
@@ -1515,6 +1548,6 @@ def from_database(cls, query_string='SELECT * FROM public.images'):
                 adjacency[spath].append(dpath)
         session.close()
         # Add nodes that do not overlap any images
-        obj = cls.from_adjacency(adjacency, node_id_map=adjacency_lookup, config=config)
+        obj = cls(adjacency, node_id_map=adjacency_lookup, config=config)
```
jlaura: As I dug through the networkX docs I could not find any reference to `from_adjacency`. Turns out the default constructor is `from_adjacency` so all good there.

autocnet/spatial/overlap.py
```diff
@@ -103,3 +103,22 @@ def place_points_in_overlaps(cg, size_threshold=0.0007, reference=None, height=0
...
+ SELECT images.id, intersectiongeom.geom AS geom
+		FROM images, intersectiongeom
+		WHERE images.footprint_latlon is NOT NULL AND
+		ST_INTERSECTS(intersectiongeom.geom, images.footprint_latlon) AND
+		ST_AREA(ST_INTERSECTION(intersectiongeom.geom, images.footprint_latlon)) > 0.000001
+)
+INSERT INTO overlay(intersections, geom) SELECT row.intersections, row.geom FROM 
+(SELECT iid.geom, array_agg(iid.id) AS intersections
+  FROM iid GROUP BY iid.geom) AS row WHERE array_length(intersections, 1) > 1;
+"""
```
jlaura: ... 👎 , but I still get a 🏅 for coming up with whatever this is.

jlaura:
@all I restarted the build and all passing. Failures were on the conda side installing the dependencies. Looks like a network issue was to blame.

autocnet/graph/network.py
```diff
@@ -1459,6 +1475,23 @@ def to_isis(self, path, flistpath=None,         sql = """
...
+        elif os.path.exists(filelist):
+            filelist = io_utils.file_to_list(filelist)
+        else:
+            warning.warn('Unable to parse the passed filelist')
+
+        for f in filelist:
+            # Create the nodes in the graph. Really, this is creating the
+            # images in the DB
+            image_name = os.path.basename(f)
+            NetworkNode(image_path=f, image_name=image_name)
```
Kelvinrr: NetworkNode init handles footprint generation and overlap computation?

autocnet/graph/network.py
```diff
@@ -1459,6 +1475,23 @@ def to_isis(self, path, flistpath=None,         sql = """
...
+        
+        return cls.from_database()
```
Kelvinrr: 👍

autocnet/spatial/overlap.py
Kelvinrr: Any way to use the query from the sql folder instead of duplicating?

autocnet/graph/network.py
jlaura: yes, the properties are lazy loaded and if they do not exist they are computed

autocnet/spatial/overlap.py
jlaura: I should remove those queries. This one supersedes that one and it is cleaner to have in one place. Any issue with me removing those?

Kelvinrr: nah just remove them

jlaura: Done.

jlaura:
@Kelvinrr I think that this should be good to go. I do not see CI processing again on the deleted file (which is maybe a little concerning...not that the sql was being tested).

PR #317, opened 2019-06-11
Jigsaw Reingest

acpaquette:
Initial addition of the jigsaw reingest. Tests incoming

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1458,6 +1458,40 @@ WHERE points.active = True AND measures.active=TRUE AND measures.jigreject=FALSE
         cnet.to_isis(path, df, self.serials())
         cnet.write_filelist(self.files, path=flistpath)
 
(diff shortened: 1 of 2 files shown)


acpaquette:
I'm running into weird issues while trying to test this.
It almost feels like Schrodinger's database where a columns true value is only known when observed. :grimacing: I'm going to PR what I have in case anyone knows what's happening

jlaura:
Here is the fix: https://github.com/jlaura/autocnet/commit/f43854e2185725f5910121782f407b42a269da62

The issue is actually Schrodinger's Session. The test has a session object and the method made use of a different session object. 

The linked solution is one potential solution. The thing that we need to decide is who is responsible for the session. My preference would be to have the session passed in to these non-method items.

jlaura:
This also adds an __repr__ to the measures so that I could debug what was going on.

PR #319, opened 2019-06-12
Specified lat and lon to increase readability

kaitlyndlee:


thareUSGS:
+1 to be very specific for lon, lat or lat, lon order.  Insert scream here: https://macwright.org/lonlat/

acpaquette:
@kaitlyndlee Some conflicts need to be resolved before this can get merged. Probably just happened from a different PR getting merged in

PR #320, opened 2019-06-13
control point type propagation on to_isis

Kelvinrr:
Goes with an update to plio


autocnet/matcher/subpixel.py
```diff
@@ -237,42 +237,56 @@ def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=251, reduction=11, conver
...
+        # Apply the shift to d_search and compute the new correspondence location
+        dx += (shift_x + dxr)
+        dy += (shift_y + dyr)
+
+        # Break if the solution has converged
+        size -= reduction
+        dist = np.linalg.norm([dsample-dx, dline-dy])
+        if size <1:
+            return None, None, None
+        if abs(shift_x) <= convergence_threshold and\
```
Kelvinrr: seems like the only way to simulate do while in python?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -168,21 +168,27 @@ def themis_ground_to_ctx_matcher(cnet):
                     match_results.append(ex)
                     continue
 
-                if ret is not None:
+                if ret is not None and None not in ret:
```
jlaura: This is legit. If self is not the soul and soul is not in the self.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -168,21 +168,27 @@ def themis_ground_to_ctx_matcher(cnet):
...
                     x,y,metrics = ret
                 else:
                     match_results.append("Failed to Converge")
                     continue
-
-                dist = np.linalg.norm([x-dx, -1*(y-dy)])
-                match_results.append([k, x-dx, -1*(y-dy), dist])
+                try:
+                  dist = np.linalg.norm([x-dx, -1*(y-dy)])
```
jlaura: Under what case is the try necessary?

autocnet/matcher/subpixel.py
jlaura: The alternative would be to do this in the `while` above and prepopulating the vals with infs. Having said that, this looks fine to me too.

jlaura:
One comments. Let's merge this and plio concurrently.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: I think it was from the two return None originally. Should be removed.

PR #321, opened 2019-06-13
Adds a clear method to reset the DB

jlaura:


jlaura:
Since we do not have integration testing at this point, this was tested in a notebook on the deleteme database. I tested passing no args, 'images', and ['cameras', 'edges'].

jlaura:
@Kelvinrr :+1:

PR #325, opened 2019-06-17
Valid geom trigger in the db

jlaura:


acpaquette:
@jlaura Can we remove the RMS column in the measures table in this PR as well?

jlaura:
Can we put that in a different PR? Then we can warn everyone to drop the column as soon as that PR is merged.

PR #330, opened 2019-06-17
Empty footprint

jessemapel:
Avoids strange errors later on about None type objects not having the required attributes for `from_shape`


jlaura:
This looks good. Once we have passing CI and a test we can merge.

jessemapel:
NetworkNode doesn't have any tests at the moment, so I'm going to start writing tests for it.

jlaura:
I was totaling mean just on the model. Knowing that the network node is not tested.

PR #331, opened 2019-06-17
Adding isis ground_to_image/image_to_ground variants in

Kelvinrr:
Until CSM is more matured I think this will still be useful, especially for cross instrument stuff coming up where we do not have Themis/HiRISE drivers for CSM.


autocnet/spatial/overlap.py
```diff
@@ -180,30 +189,39 @@ def place_points_in_overlap(nodes, geom, dem=None,
...
-                                       line=sic.line,
+
+        if cam_type == "csm":
+            # Calculate the height, the distance (in meters) above or
+            # below the aeroid (meters above or below the BCBF spheroid).
+            if dem is None:
+                height = 0
+            else:
+                px, py = dem.latlon_to_pixel(lat, lon)
+                height = dem.read_array(1, [px, py, 1, 1])[0][0]
```
jlaura: Single question - does this capability already exist in the code from @kaitlyndlee? If so, can we use that? If not, I think we can merge this.

Kelvinrr: can you point me to the location of the code so I can check it out? Not sure what code you're talking about.

jlaura: nope, because it was in this file....jfail

autocnet/spatial/isis.py
```diff
@@ -0,0 +1,95 @@
...
+        point in the x direction. Either a sample or a longitude value
+        depending on the point_type flag
+
+    y : float
+        point in the y direction. Either a line or a latitude value
+        depending on the point_type flag
+
+    point_type : str
+                 whether x,y are in image space (sample, line) or
+                 ground space (longitude, lattiude)
```
jessemapel: It would be nice to have the options for this parameter documented. It looks like they are just the ISIS options of `IMAGE` for `(sample, line)` and `GROUND` for `(lat, lon)`.

autocnet/spatial/overlap.py
```diff
@@ -82,7 +85,7 @@ def place_points_in_overlaps(cg, size_threshold=0.0007,
 
 def cluster_place_points_in_overlaps(size_threshold=0.0007,
                                      iterative_phase_kwargs={'size':71},
-                                     walltime='00:10:00'):
+                                     walltime='00:10:00', cam_type="csm"):
```
jessemapel: I think this is fine for right now, but it would be nice to make this a config option.

jessemapel: We would probably want to make it consistent across the code base.

jessemapel: New param needs docs.

autocnet/spatial/overlap.py
```diff
@@ -135,7 +139,7 @@ def cluster_place_points_in_overlaps(size_threshold=0.0007,
     submitter.submit(array='1-{}'.format(job_counter))
     return job_counter
 
-def place_points_in_overlap(nodes, geom, dem=None,
+def place_points_in_overlap(nodes, geom, dem=None, cam_type="csm",
```
jessemapel: Same as above, needs docs for new param

autocnet/spatial/overlap.py
Kelvinrr: DorDnbAU4AARTc3

jlaura: I ❤️ fail whale!

autocnet/spatial/overlap.py
Kelvinrr: agreed, I would say for now we can do that in a different PR.

Kelvinrr: Probably skip on docs for now since this will probably change within a week.

jlaura: :+1:

jlaura:
Once CI finishes this should merge.

PR #332, opened 2019-06-18
Changes the func that was making N/S point more dense

jlaura:
This is a short term solution that removes a hard coded limiter on the number of points. This should get further refactored. I'll open an issue.

### DIFF
--- a/autocnet/cg/cg.py
+++ b/autocnet/cg/cg.py
@@ -366,8 +366,6 @@ def distribute_points_in_geom(geom):
     elif ns==True:
         # Class, north/south poly, multi-point
         nspts = int(round(long, 1) * 10)
-        if nspts >= 5:
-            nspts = int(nspts/(nspts/3))
         ewpts = max(int(round(short, 1) * 5), 1)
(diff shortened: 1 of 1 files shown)


autocnet/cg/cg.py
```diff
@@ -366,8 +366,6 @@ def distribute_points_in_geom(geom):
     elif ns==True:
         # Class, north/south poly, multi-point
         nspts = int(round(long, 1) * 10)
-        if nspts >= 5:
-            nspts = int(nspts/(nspts/3))
```
Kelvinrr: What do these removed lines suppose to be doing and how did it make it less dense?

jlaura: It was a method to remove the number of along track points if a geometry was going to have more than 5. Removing this means that we take the long distance (in degrees) multiple it by 10 and then place that many points. This is ghetto, but works at the equator (ish). Working in degrees is most janky part of this.

Kelvinrr:
Merging this so it can be used right away. In the near future, we need to control the density through parameters.

PR #334, opened 2019-06-20
🚧 Initial expansion on the point/measure tables 🚧

jlaura:


autocnet/io/db/model.py
```diff
@@ -258,19 +258,46 @@ class PointType(enum.IntEnum):
...
+    datetime = Column(DateTime)
+    editLock = Column(Boolean)
+    ignore = Column(Boolean)
+    jigsawrejected = Column(Boolean)
+    referenceindex = Column(Integer)
+    # Our own, is this point live field
     active = Column(Boolean, default=True)
+    
+    # Provenance
+    apriori_surfacepointsource = Column(Integer)
```
jessemapel: If we map point types from enumerations (ints) to strings, we should map these enumerations too

jlaura:
@acpaquette Is your notebook based off of this?

acpaquette:
@jlaura No, those two notebooks should work with just autocnet dev and plio master with some potential local changes :grimacing:

jlaura:
@all this needs to wait until we have a decision about BCBF vs. lat/lon.

jlaura:
I am going to leave this until #347 and #349 are merged. Then this can get updated, reviewed, and merged, to be able to round trip a full cnet.

jessemapel:
@jlaura Is this still a thing we want to do? Should it just be closed?

PR #339, opened 2019-06-21
Fixed typo getting dem from config

jessemapel:
Just a minor typo fix.


jlaura:
The alternative, where we get the `spatial` key and then get the `dem` is not needed here because we are wrapped in a try/accept anyway.

PR #341, opened 2019-06-21
i don't like it

jlaura:


autocnet/io/db/model.py
```diff
@@ -263,14 +264,25 @@ class Points(BaseMixin, Base):
     identifier = Column(String, unique=True)
     _geom = Column("geom", Geometry('POINT', srid=srid, dimension=2, spatial_index=True))
     active = Column(Boolean, default=True)
-    apriorix = Column(Float)
-    aprioriy = Column(Float)
-    aprioriz = Column(Float)
+    _apriorix = Column("apriorix", Float, default=0)
+    _aprioriy = Column("aprioriy", Float, default=0)
+    _aprioriz = Column("aprioriz", Float, default=0)
```
Kelvinrr: What's the reasoning behind making this private with setters/getters?

jlaura: something (geom or these) should take precedence. This is maybe something to talk about in a few. If we store two piece of identical information, which is the 'correct' one? Can the lat/lon geom be created from an xyz geom? Does the DB only store public XYZ geoms? Or only Lat/Lon geoms because of what a GIS wants to consume. I have been back and forth on this this morning a half dozen times.

PR #342, opened 2019-06-22
Remote

jlaura:
This PR demonstrates how to get the bin scripts working properly with a remote jupyterhub kernel. We set enough other env variables that it is not longer concerning to set one more.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1298,9 +1298,6 @@ class NetworkCandidateGraph(CandidateGraph):
         for s, d, e in self.edges(data='data'):
             e.parent = self
 
-        # Execute the computation to compute overlapping geometries
-        self._execute_sql(compute_overlaps_sql)
-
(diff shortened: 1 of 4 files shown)


jessemapel:
This doesn't look like it actually does anything with the `bin` directory set in the config

jlaura:
@jessemapel I need to blackout the first file. The second adds the bin and then render the ipynb to see how the path gets patched in.

jlaura:
Fixed the erroneous include. This file is updated in another PR.

jessemapel:
Manually updating the PATH in the notebook is not a good solution in my opinion. using something like [shutil.which](https://docs.python.org/dev/library/shutil.html#shutil.which) to check if the script is already in the path and if it's not checking the bin config variable inside the library makes more sense.

Right now, there is no reason to even have the bin in the config, it can just be a variable in the notebook.

PR #344, opened 2019-06-23
Exposes point funcs

jlaura:
This PR exposes parameters to allow users to control the number of points in a given direction. The parameters are exposed as funcs where the defaults are the kwarg lambdas.

The doc string describes the logic of the funcs that can be passed. I also ported the unit test tests over to pytest and added a few more. I am see ~91% coverage on the cg stuff locally.

### DIFF
--- a/autocnet/cg/cg.py
+++ b/autocnet/cg/cg.py
@@ -238,28 +238,8 @@ def nearest(pt, search):
     """
     return np.argmin(np.sum((search - pt)**2, axis=1))
 
-def compute_slope(a, b):
-    if b[1] - a[1] == 0:
-        # Horizontal
-        return 0
-    elif b[0]-a[0] == 0:
-        # Vertical
-        return 
-    else:
-        return (b[1] - a[1]) / (b[0]-a[0])
-
 def create_points_along_line(p1, p2, npts):
-
-    m = compute_slope(p1, p2)
-    # The +2 is because the start/stop are later stripped from the list
-    if m is not None:
-        b = (p1[1] - p1[0] * m)
-        nodes = np.linspace(p1[0], p2[0], num=npts+2)
-        points = [(x, m*x+b) for x in nodes[1:-1]]
-    else:
-        nodes = np.linspace(p1[1], p2[1], num=npts+2)
-        points = [(p1[0], y) for y in nodes[1:-1]]
-    return points
+    return np.linspace(p1, p2, npts+2)[1:-1]
 
 def xy_in_polygon(x,y, geom):
     """
--- a/autocnet/cg/cg.py
+++ b/autocnet/cg/cg.py
@@ -350,10 +350,25 @@ def distribute_points_in_geom(geom,
     is made, the algorithm places points in the geometry and returns
     a list of valid (intersecting) points.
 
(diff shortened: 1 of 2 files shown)


autocnet/cg/cg.py
```diff
@@ -238,6 +238,52 @@ def nearest(pt, search):
...
+
+    m = compute_slope(p1, p2)
+    # The +2 is because the start/stop are later stripped from the list
+    if m is not None:
+        b = (p1[1] - p1[0] * m)
+        nodes = np.linspace(p1[0], p2[0], num=npts+2)
+        points = [(x, m*x+b) for x in nodes[1:-1]]
+    else:
+        nodes = np.linspace(p1[1], p2[1], num=npts+2)
+        points = [(p1[0], y) for y in nodes[1:-1]]
```
jessemapel: These two functions can just be a call to `np.linspace(p1, p2, npts)` It will produce a sequence of points from p1 to p2.

jlaura: linspace supports 2d - sweet! updated.

jlaura:
@jessemapel @ladoramkershner Tests all passing locally. I need to make one more modification to spatial.py in order to expose this capability higher up in the API. Right now, this is very low level - only the cg.py file has been touched.

jlaura:
Scratch the above - that change needs to be in another PR. This is going to get too big otherwise.

autocnet/cg/cg.py
```diff
@@ -238,6 +238,32 @@ def nearest(pt, search):
     """
     return np.argmin(np.sum((search - pt)**2, axis=1))
 
+def create_points_along_line(p1, p2, npts):
+    return np.linspace(p1, p2, npts+2)[1:-1]
```
Kelvinrr: Does this function really need to exist? why not an explicit numpy call?

autocnet/cg/cg.py
```diff
@@ -238,6 +238,32 @@ def nearest(pt, search):
...
+    y : Number
+        The y coordinate
+
+    Returns
+    -------
+     : bool
+       True if the point is contained within the geom.
+    """
+    pt = Point(x, y)
+    if geom.contains(pt):
```
Kelvinrr: `return geom.contains(pt)` seems cleaner, or maybe use `Point(x,y).contains(pt)` and just skip the seperate function all together?

autocnet/cg/cg.py
jlaura: I pushed it in because we would be making this call to np 3 times with different args and then clipping the ends each time. I am happy to pop this and replace the inline calls to this with the np calls.

autocnet/cg/cg.py
jlaura: :+1:

jlaura: Updated in the latest commit on the PR.

autocnet/cg/cg.py
Kelvinrr: I think as long as the `+2` and `[1:-1]` are documented here it would be worth staying as a separate func.

jlaura: docs added

PR #345, opened 2019-06-25
Fully exposes distribution of points in geom

jlaura:
I am going to put comments inline.


autocnet/io/db/model.py
```diff
@@ -35,6 +36,13 @@ def create(cls, session, **kw):
         session.commit()
         return obj
 
+    @staticmethod
+    def bulkadd(iterable):
+        session = Session()
+        session.add_all(iterable)
+        session.commit()
+        session.close()
+
```
jlaura: As above with having funcs doing multiple things. Testing the overlap func with point adding meant having to have the func create and add valid points which was a pain when all I wanted to test was proper kwarg passing.

autocnet/io/db/model.py
```diff
@@ -248,6 +256,23 @@ def geom(self):
...
+        size_threshold : Number
+                        area >= this arg are returned
+        """
+        session = Session()
+        res = session.query(self).\
+                filter(sqlalchemy.func.ST_Area(self.geom) >= size_threshold).\
+                filter(sqlalchemy.func.array_length(self.intersections, 1) > 1)
+        session.close()
+        return res
+
```
jlaura: This was duplicated twice in the overlap code meaning the DB was tightly coupled with that code. Doing it this way should allow for baby autocnet to make a geopandas call in the future to then use the identical func.

autocnet/io/db/tests/test_model.py
```diff
@@ -115,8 +115,7 @@ def test_points_exists(tables):
     {'id':1, 'pointtype':2},
     {'pointtype':2, 'identifier':'123abc'},
     {'pointtype':3, 'geom':Point(0,0)},
-    {'pointtype':2, 'geom':Point(1,1), 'active':True},
-    {'pointtype':2, 'rms':0.001}
+    {'pointtype':2, 'geom':Point(1,1), 'active':True}
```
jlaura: no rms in the db, so no rms test

autocnet/spatial/overlap.py
```diff
@@ -7,7 +7,7 @@
 import sqlalchemy
 from plio.io.io_gdal import GeoDataset
 
-from autocnet import config, Session, engine
+from autocnet import config, dem
```
jlaura: now DEM needs to come in from the root namespace

autocnet/spatial/overlap.py
```diff
@@ -48,40 +49,29 @@ def place_points_in_overlaps(cg, size_threshold=0.0007,
 
     Parameters
     ----------
-    cg : CandiateGraph object
-         that is used to access sensor information
+    nodes : dict-link
+            A dict like object with a shared key with the intersection
+            field of the database Overlay table and a cg node object
+            as the value. This could be a NetworkCandidateGraph or some
+            other dict-like object.
```
jlaura: accepting any dict-like (which then ncg is) makes testing this much more straightforward. I would argue this is more pythonic too since was are taking advantage of duck typing.

autocnet/spatial/overlap.py
```diff
@@ -48,40 +49,29 @@ def place_points_in_overlaps(cg, size_threshold=0.0007,
...
-        dem = config['spatial']['dem']
-        gd = GeoDataset(dem)
-    else:
-        gd = None
-
-    # TODO: This should be a passable query where we can subset.
-    for o in session.query(Overlay).\
-             filter(sqlalchemy.func.ST_Area(Overlay.geom) >= size_threshold).\
-             filter(sqlalchemy.func.array_length(Overlay.intersections, 1) > 1):
+    for o in Overlay.overlapping_larger_than(size_threshold):
```
jlaura: this whole func gets pretty basic with all the 'extras' refactored out.

autocnet/spatial/overlap.py
```diff
@@ -183,7 +166,7 @@ def place_points_in_overlap(nodes, geom, dem=None, cam_type="csm",
     ecef = pyproj.Proj(proj='geocent', a=semi_major, b=semi_minor)
     lla = pyproj.Proj(proj='latlon', a=semi_major, b=semi_minor)
 
-    valid = compgeom.distribute_points_in_geom(geom)
+    valid = compgeom.distribute_points_in_geom(geom, **distribute_points_kwargs)
```
jlaura: all this PR to expose (and test) this....

autocnet/spatial/tests/test_overlap_points.py
```diff
@@ -58,3 +58,26 @@ def test_place_points_in_overlap(point_distributer, phase_matcher):
...
+
+
+
+@patch('autocnet.io.db.model.Overlay.overlapping_larger_than', return_value=[MockOverlap()]*3)
+@patch('autocnet.dem', return_value=None)
+@patch('autocnet.io.db.model.Points.bulkadd')
+@pytest.mark.parametrize("iterativekwargs, distributekwargs",[
+    ({'iterative_phase_kwargs':{'size':15}},
+     {'distribute_points_kwargs':{'ewpts_func':lambda:True}})
+])
```
jlaura: patches for the items we don't want to actually call. What I really want to test is that the kwargs are being passed through properly since this func is just a dispatcher now to a bunch of other pieces.

autocnet/spatial/tests/test_overlap_points.py
```diff
@@ -58,3 +58,26 @@ def test_place_points_in_overlap(point_distributer, phase_matcher):
...
+def test_place_points_in_overlaps(overlapper, dem, adder,iterativekwargs, distributekwargs):
+    nodes = [{'id':0}, {'id':1}]
+    with patch('autocnet.spatial.overlap.place_points_in_overlap', return_value=[1,2,3]) as ppio:
+        place_points_in_overlaps(nodes, iterative_phase_kwargs=iterativekwargs,distribute_points_kwargs=distributekwargs)
+        ppio.assert_called_with(nodes,
+                                Polygon([(0,0),(0,5),(5,5),(5,0),(0,0)]),
+                                dem=None,
+                                iterative_phase_kwargs=iterativekwargs,
+                                distribute_points_kwargs=distributekwargs)
```
jlaura: Don't care what happens in the ppio black box, because it is tested elsewhere. All I want to test is that the call args are moving properly.

autocnet/__init__.py
```diff
@@ -29,6 +31,16 @@
...
+if 'dem' in config['spatial']:
+    dem = config['spatial']['dem']
+    try:
+        dem = GeoDataset(dem)
+    except:
+        warnings.warn(f'Unable to load the dem: {dem}')
+        dem = None
+else:
+    dem = None
+
```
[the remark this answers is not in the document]
jessemapel: Should we have a separate file that is purely managing the config? We have stuff like setting default SRIDs in the model.

autocnet/__init__.py
```diff
@@ -42,8 +54,13 @@
                     isolation_level="AUTOCOMMIT")                   
     Session = orm.session.sessionmaker(bind=engine)
 except: 
-    Session = None
-    engine = None
+    def sessionwarn():
+        raise RuntimeError('This call requires a database connection.')
+    
+    Session = sessionwarn
+    engine = sessionwarn
```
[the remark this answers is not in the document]
jessemapel: Should we warn right away if no session/engine is available and then error if something tries to use it? This looks like it will only error when something tries to use the session/engine.

autocnet/__init__.py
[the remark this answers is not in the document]
jlaura: I think that that makes sense as a further enhancement. What we really want to do is refactor any direct calls inside of autocnet to the config to a single place. This is how Flask handles the `application` object and the interface is pretty nice.

autocnet/__init__.py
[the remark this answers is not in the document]
jlaura: This will only warn when someone tries to use it. For baby autocnet, we have no reason to warn/raise which is why I put this here. The above is problematic though: `session = Session()` will raise. `engine.url` will fail with a more obtuse error about a func not having a member. 100% open to thoughts here. Keep `engine` as `=None` and let standard python do its thing, etc.

jlaura:
@jessemapel Updated for the submission stuff. Perhaps @ladoramkershner or Ethan can test this on the demo notebook with cluster submission since the bin scripts are not being CI tested in any way. This has the potential to be breaking I I screw something up in the message passing and debugging would be a pain down the line.

bin/acn_overlaps
```diff
@@ -37,15 +37,11 @@ def main(msg, config):
...
 
     print('Placing points in overlap', id)
-    points = place_points_in_overlap(nodes, overlap.geom, gd, msg["cam_type"],
-                                     msg['iterative_phase_kwargs'])
+    points = place_points_in_overlap(nodes, overlap.geom, dem, msg["cam_type"],
+                                     msg['iterative_phase_kwargs'],
+                                     msg['distribute_points_kwargs'])
     session.add_all(points)
     session.commit()
     session.close()
```
jessemapel: This should to be updated to use the new wrapped method to add the points.

jlaura: What wrapped method are you referring to? This looks like the correct point since we know the nodes and geoms.

jessemapel: `Points.bulkadd(points)` instead of manually adding to the session. Then, the session can be closed immediately after line 39.

jlaura:
+1 I was looking at the func call above.

bin/acn_overlaps
```diff
@@ -37,18 +37,14 @@ def main(msg, config):
...
-        gd = GeoDataset(dem)
-    else:
-        gd = None
 
     print('Placing points in overlap', id)
-    points = place_points_in_overlap(nodes, overlap.geom, gd, msg["cam_type"],
-                                     msg['iterative_phase_kwargs'])
-    session.add_all(points)
-    session.commit()
-    session.close()
```
jessemapel: The session doesn't get closed now! It should be closed after all of the nodes are collected.

jessemapel:
I think this is good to be manually tested now

jlaura:
Good catch on the session. I am looking at the code base now and we are inconsistent in closing sessions and who has that responsibility.

autocnet/io/db/model.py
ladoramkershner: I do not understand why this function (overlapping_larger_than) is repeated exactly, within the same class, one right after the other. Is this a python thing I just don't know about.

jlaura:
@ladoramkershner @jessemapel Just fixed for more merge conflicts. Is this being held merging because of CSM or for another reason? Just asking so that I can stop fixing merge conflicts with newer PRs and get this closed off.

PR #346, opened 2019-06-25
Bug fixes in Demo notebook and for Demo notebook

ladoramkershner:
Errors that probably have been seen previously and fixed locally. Mostly just adding ['data'] for source and destination images and fixing typos

Took out height parameter from ng.place_points_in_overlaps()

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1534,7 +1534,7 @@ WHERE points.active = True AND measures.active=TRUE AND measures.jigreject=FALSE
         elif os.path.exists(filelist):
             filelist = io_utils.file_to_list(filelist)
         else:
-            warning.warn('Unable to parse the passed filelist')
+            warnings.warn('Unable to parse the passed filelist')
 
         for f in filelist:
             # Create the nodes in the graph. Really, this is creating the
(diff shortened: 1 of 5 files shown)


Kelvinrr:
If no ones has objections I'll merge this once/if the CI passes.

jlaura:
@Kelvinrr I am not sure how CI is/will pass with all of the additional `['data']` lines. Is this to fix something that I broke in my last PR or something else?

Kelvinrr:
No idea where or when it happened, been using ISIS so I never ran into it. the test will definitely fail because the test assumes that the data key doesn't have to be there so the problem seems to be the code was written as some point assuming a node data object but place_points_in_overlaps is passing the dictionary instead.

jlaura:
Scratch my comment because I was just referencing a not merged alteration (the other open PR). This PR should probably merge first to keep everyone rolling. Then I can deal with the merge issues and get the other PR updated. Does that sound reasonable?

Kelvinrr:
you talking about #345? Yeah it sounds reasonable, I am trying to merge this one ASAP. without it no one can really move forward.

jlaura:
@Kelvinrr yup

PR #347, opened 2019-06-25
Change Points to use XYZ in the DB and automatically update the lat, lon geom

jessemapel:
Updated points Table to store points in XYZ and provide a lat/lon representation.

This is going to force people to re-create their databases so they get the correct table and triggers.

### DIFF
--- a/autocnet/io/db/model.py
+++ b/autocnet/io/db/model.py
@@ -278,10 +278,8 @@ class Points(BaseMixin, Base):
 
     @geom.setter
     def geom(self, geom):
-        if geom:
-            self._geom = from_shape(geom, srid=latitudinal_srid)
-        else:
-            self._geom = geom
+        raise TypeError("The geom column for Points cannot be set." \
+                        " Set the adjusted column to update the geom.")
 
     @hybrid_property
     def apriori(self):
--- a/autocnet/io/db/model.py
+++ b/autocnet/io/db/model.py
@@ -264,8 +264,8 @@ class Points(BaseMixin, Base):
     identifier = Column(String, unique=True)
     _geom = Column("geom", Geometry('POINT', srid=latitudinal_srid, dimension=2, spatial_index=True))
     active = Column(Boolean, default=True)
-    apriori = Column(Geometry('POINT', srid=rectangular_srid, dimension=3))
-    adjusted = Column(Geometry('POINT', srid=rectangular_srid, dimension=3))
+    _apriori = Column("apriori", Geometry('POINTZ', srid=rectangular_srid, dimension=3, spatial_index=False))
+    _adjusted = Column("adjusted", Geometry('POINTZ', srid=rectangular_srid, dimension=3, spatial_index=False))
     measures = relationship('Measures')
     rms = Column(Float)
 
@@ -280,6 +280,36 @@ class Points(BaseMixin, Base):
     def geom(self, geom):
         if geom:
             self._geom = from_shape(geom, srid=latitudinal_srid)
+        else:
+            self._geom = geom
+
+    @hybrid_property
+    def apriori(self):
+        try:
+            return to_shape(self._apriori)
+        except:
+            return self._apriori
+
+    @apriori.setter
+    def apriori(self, apriori):
+        if apriori:
+            self._apriori = from_shape(apriori, srid=rectangular_srid)
+        else:
+            self._apriori = apriori
+
+    @hybrid_property
+    def adjusted(self):
+        try:
+            return to_shape(self._adjusted)
+        except:
+            return self._adjusted
+
+    @adjusted.setter
+    def adjusted(self, adjusted):
+        if adjusted:
+            self._adjusted = from_shape(adjusted, srid=rectangular_srid)
+        else:
+            self._adjusted = adjusted
 
     @hybrid_property
     def pointtype(self):
--- a/autocnet/io/db/model.py
+++ b/autocnet/io/db/model.py
@@ -264,12 +264,8 @@ class Points(BaseMixin, Base):
     identifier = Column(String, unique=True)
     _geom = Column("geom", Geometry('POINT', srid=latitudinal_srid, dimension=2, spatial_index=True))
     active = Column(Boolean, default=True)
-    apriorix = Column(Float)
-    aprioriy = Column(Float)
-    aprioriz = Column(Float)
-    adjustedx = Column(Float)
-    adjustedy = Column(Float)
-    adjustedz = Column(Float)
+    apriori = Column(Geometry('POINT', srid=rectangular_srid, dimension=3))
+    adjusted = Column(Geometry('POINT', srid=rectangular_srid, dimension=3))
     measures = relationship('Measures')
     rms = Column(Float)
 
@@ -334,7 +330,7 @@ class Measures(BaseMixin, Base):
         self._measuretype = v
 
 if Session:
-    from autocnet.io.db.triggers import valid_point_function, valid_point_trigger, valid_geom_function, valid_geom_trigger
+    from autocnet.io.db.triggers import valid_point_function, valid_point_trigger, update_point_function, update_point_trigger, valid_geom_function, valid_geom_trigger
     # Create the database
     if not database_exists(engine.url):
         create_database(engine.url, template='template_postgis')  # This is a hardcode to the local template
@@ -343,6 +339,8 @@ if Session:
         # based on the point count.
         event.listen(Base.metadata, 'before_create', valid_point_function)
         event.listen(Measures.__table__, 'after_create', valid_point_trigger)
+        event.listen(Base.metadata, 'before_create', update_point_function)
+        event.listen(Images.__table__, 'after_create', update_point_trigger)
         event.listen(Base.metadata, 'before_create', valid_geom_function)
         event.listen(Images.__table__, 'after_create', valid_geom_trigger)
 
--- a/autocnet/io/db/model.py
+++ b/autocnet/io/db/model.py
@@ -24,8 +24,9 @@ from autocnet import engine, Session, config
 Base = declarative_base()
 
 # Default to mars if no config is set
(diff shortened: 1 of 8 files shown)


autocnet/io/db/model.py
```diff
@@ -24,8 +24,9 @@
 Base = declarative_base()
 
 # Default to mars if no config is set
-spatial = config.get('spatial', {'srid': 949900})
-srid = spatial['srid']
+spatial = config.get('spatial', {'latitudinal_srid': 949900, 'rectangular_srid': 949980})
+latitudinal_srid = spatial['latitudinal_srid']
+rectangular_srid = spatial['rectangular_srid']
```
jessemapel: This is a bad place to default these, can I move this to the config.py file?

autocnet/io/db/model.py
```diff
@@ -281,8 +278,36 @@ def geom(self):
...
+            return to_shape(self._apriori)
+        except:
+            return self._apriori
+
+    @apriori.setter
+    def apriori(self, apriori):
+        if apriori:
+            self._apriori = from_shape(apriori, srid=rectangular_srid)
+        else:
+            self._apriori = apriori
```
jessemapel: All of the other geometry setters do not have the else block like these two. Without it, you cannot set them to None/Null. Should I add similar else blocks to things like the Overlay geom setter on line 248?

autocnet/io/db/model.py
```diff
@@ -281,8 +278,36 @@ def geom(self):
 
     @geom.setter
     def geom(self, geom):
-        if geom:
-            self._geom = from_shape(geom, srid=srid)
+        raise TypeError("The geom column for Points cannot be set." \
+                        " Set the adjusted column to update the geom.")
```
jessemapel: This is the best way I can think of to prevent people from directly modifying the geometry column.

autocnet/io/db/tests/test_model.py
```diff
@@ -23,7 +23,9 @@ def session(tables, request):
     def cleanup():
         session.rollback()  # Necessary because some tests intentionally fail
         for t in reversed(tables):
-            session.execute(f'TRUNCATE TABLE {t} CASCADE')
+            # Skip the srid table
+            if t != 'spatial_ref_sys':
+                session.execute(f'TRUNCATE TABLE {t} CASCADE')
```
jessemapel: If this truncates the `spatial_ref_sys` table, then we lose all of our srids after the first test! Are there other tables we don't want to truncate?

autocnet/io/db/triggers.py
```diff
@@ -57,4 +59,31 @@
   ON measures
   FOR EACH ROW
 EXECUTE PROCEDURE validate_points();
-""")
\ No newline at end of file
+""")
+
+latitudinal_srid = config['spatial']['latitudinal_srid']
```
jessemapel: This is potentially unsafe. What should happen if this doesn't exist in the config?

autocnet/io/db/model.py
jlaura: yes - this should go into the config. We can also consider allowing a full 'grid' obj where if the srid/string is not in the DB we add it. Something down the road.

autocnet/io/db/model.py
jlaura: :+1: a general question here - are the points the only table operating in rectangular space? It looks like the footprints above are still in latitudinal space? I think that this is totally fine, just checking.

autocnet/io/db/model.py
jlaura: Yes.

jlaura: But this could be in a different PR to keep this mergeable in shorter order.

autocnet/io/db/tests/test_model.py
jlaura: 🤦‍♂  - good catch.

jlaura: Is this working without the reverse? The reverse was extra insurance on top of the CASCADE, so likely not needed. Just curious.

autocnet/io/db/tests/test_model.py
```diff
@@ -114,15 +116,39 @@ def test_points_exists(tables):
 @pytest.mark.parametrize("data", [
     {'id':1, 'pointtype':2},
     {'pointtype':2, 'identifier':'123abc'},
-    {'pointtype':3, 'geom':Point(0,0)},
-    {'pointtype':2, 'geom':Point(1,1), 'active':True},
+    {'pointtype':3, 'apriori':Point(0,0,0)},
+    {'pointtype':3, 'adjusted':Point(0,0,0)},
```
jlaura: :+1: This means we can stop the merges soon! sweet.

autocnet/io/db/tests/test_model.py
```diff
@@ -114,15 +116,39 @@ def test_points_exists(tables):
...
     assert p == resp
 
+@pytest.mark.parametrize("data, expected", [
+    ({'pointtype':3, 'adjusted':Point(0,-1,0)}, Point(-90, 0)),
+    ({'pointtype':3}, None)
+])
+def test_create_point_geom(session, data, expected):
+    p = model.Points.create(session, **data)
+    resp = session.query(model.Points).filter(model.Points.id == p.id).first()
+    print(resp.geom)
```
jlaura: remove?

autocnet/io/db/tests/test_model.py
```diff
@@ -114,15 +116,39 @@ def test_points_exists(tables):
...
     resp = session.query(model.Points).filter(model.Points.id == p.id).first()
     assert p == resp
 
+@pytest.mark.parametrize("data, expected", [
+    ({'pointtype':3, 'adjusted':Point(0,-1,0)}, Point(-90, 0)),
+    ({'pointtype':3}, None)
+])
+def test_create_point_geom(session, data, expected):
+    p = model.Points.create(session, **data)
+    resp = session.query(model.Points).filter(model.Points.id == p.id).first()
```
jlaura: we should switch to using `one()` if and only if we expect a single result. Then this will error if we ever have double entries that we are not expecting.

autocnet/io/db/tests/test_model.py
```diff
@@ -114,15 +116,39 @@ def test_points_exists(tables):
...
+    ({'pointtype':3, 'adjusted':Point(0,-1,0)}, None, None),
+    ({'pointtype':3, 'adjusted':Point(0,-1,0)}, Point(0,1,0), Point(90, 0)),
+    ({'pointtype':3}, Point(0,-1,0), Point(-90, 0))
+])
+def test_update_point_geom(session, data, new_adjusted, expected):
+    p = model.Points.create(session, **data)
+    p.adjusted = new_adjusted
+    session.commit()
+    resp = session.query(model.Points).filter(model.Points.id == p.id).first()
+    print(resp.geom)
```
jlaura: remove?

autocnet/io/db/tests/test_model.py
```diff
@@ -114,15 +116,39 @@ def test_points_exists(tables):
...
+    assert resp.geom == expected
+
```
jlaura: `session.close()` - though I have no idea how this works in the tests when the full DB is than truncated.

autocnet/io/db/triggers.py
jlaura: Leave for now? Then we need to centrally locate in a place that does checking.

config/sample.yml
```diff
@@ -55,19 +55,23 @@ redis:
...
+    # The spatial reference identifier for rectangular geometries in the database.
+    rectangular_srid: 949980
     # The radii of the body. If the body is a sphere, these values should match.
     # Units are in meters
     semimajor_rad: 3396190
     semiminor_rad: 3376200
-    # A compliant proj4 string. Checkout http://spatialreference.org for examples 
+    # A compliant proj4 string. Checkout http://spatialreference.org for examples
     # for planetary bodies.
     proj4_str: '+proj:longlat +a:3396190 +b:3376200 +no_defs'
```
jlaura: Should we have the latitudinal_srid and rectangular_srid defined in here?

config/test_config.yml
```diff
@@ -41,7 +41,8 @@ redis:
 
 ### Spatial Reference Setup ###
 spatial:
-    srid: 4326  # 949900; this is a custom SRID for Mars. For testing we use WGS84.
+    latitudinal_srid: 4326  # For testing we use WGS84.
+    rectangular_srid: 4978
```
jlaura: is 4978 terrestrial ECEF? If so, :+1:. If not, where is it? (Just curious.)

config/test_config.yml
```diff
@@ -41,7 +41,8 @@ redis:
...
     semimajor_rad: 3396190  # in meters
     semiminor_rad: 3376200  # in meters
     proj4_str: '+proj:longlat +a:3396190 +b:3376200 +no_defs'
```
jlaura: The best part of this file is that folks in NYC are also somewhere in PA (scary!) since we just squished the earth a bit. 😆

autocnet/io/db/model.py
jessemapel: Yes the points are the only one that has anything in rectangular.

config/test_config.yml
jessemapel: I am 90% sure it is, I'm checking with Trent though

config/sample.yml
jessemapel: In the proj4 string?

jlaura: should we have 2 proj4_strs? One for latitudinal and one for rectangular.

jessemapel: Where does this actually get used though.

jlaura: haha, nowhere. We should remove it.

jessemapel:
I've tested this now, and everything is still working with the empty points fix from #349

jessemapel:
Also it looks like travis passed, github just didn't get the signal.

PR #349, opened 2019-06-26
Added `from_cnet` code

Kelvinrr:
Mostly code from @jlaura and @acpaquette but with some bugs ironed out. Tests inbound, shouldn't be merged without them.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -17,6 +17,9 @@ import shapely.geometry
 import shapely.wkt as swkt
 import shapely.ops
 
+import pyproj
+
+from plio.io.io_controlnetwork import from_isis
 from plio.io import io_hdf, io_json
 from plio.utils import utils as io_utils
 from plio.io.io_gdal import GeoDataset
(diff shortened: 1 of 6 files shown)


autocnet/graph/network.py
```diff
@@ -1632,3 +1635,83 @@ def clear_db(self, tables=None):
...
+                newcols.append('pointtype')
+            elif i == 5:
+                newcols.append('pointignore')
+            elif i == 6:
+                newcols.append('pointjigsawRejected')
+            elif i == 25:
+                newcols.append('measuretype')
+            else:
+                newcols.append(c)
+        cnet.columns = newcols
```
jlaura: Should these renames happen here or should they happen in plio?

autocnet/graph/network.py
```diff
@@ -1632,3 +1635,83 @@ def clear_db(self, tables=None):
...
+
+        cnetpoints = cnet.groupby('id')
+        points = []
+
+        for id, cnetpoint in cnetpoints:
+            def get_measures(row):
+                session = Session()
+                res = session.query(Images).filter(Images.serial == row.serialnumber).one()
+                session.close()
```
jlaura: Since this is happening in one apply, it is probably better to pass the session than to pay the session creation overhead so many times in rapid succession. Especially since the commit is not going to happen until later.

autocnet/graph/network.py
```diff
@@ -1632,3 +1635,83 @@ def clear_db(self, tables=None):
...
+                           adjustedz=row.adjustedZ,
+                           pointtype=row.pointtype)
+
+            point.measures = list(measures)
+            points.append(point)
+
+        session = Session()
+        session.add_all(points)
+        session.commit()
+        session.close()
```
jlaura: This should probably use the new `Point.bulkadd(points)` call. Then no session management in the add here.

autocnet/graph/network.py
```diff
@@ -1632,3 +1635,83 @@ def clear_db(self, tables=None):
...
+
+    @classmethod
+    def from_cnet(cls, cnet, filelist):
+        """
+
+        """
+        networkobj = cls.from_filelist(filelist)
+        networkobj.place_points_from_cnet(cnet)
```
jlaura: Does this assume that the images are already in the DB? If so, does it make sense to execute the overlay code before the obj is returned?

autocnet/graph/node.py
```diff
@@ -663,8 +664,10 @@ def footprint(self):
                 except Exception as e:
                     warnings.warn('Unable to get the Geodata from dem.\n{}'.format(e))
                     geodata = 0.0
+
                 footprint_latlon = generate_latlon_footprint(self.camera, boundary, dem=geodata)
                 footprint_latlon.FlattenTo2D()
+                return footprint_latlon
```
jlaura: bug fix?

autocnet/graph/network.py
jessemapel: These already got renamed in https://github.com/USGS-Astrogeology/plio/pull/122, it's just not merged yet.

autocnet/graph/node.py
Kelvinrr: yeah, without this you get no nodes and no edges unless your images are footprint inited.

autocnet/graph/network.py
Kelvinrr: It will fail since there will be no images to grab info from, similar to place point in overlaps

autocnet/graph/network.py
Kelvinrr: couldn't find evidence of such a method existing in autocnet.io.db.model.Points?

jlaura: Never mind. The change is in an open PR.

autocnet/graph/network.py
```diff
@@ -1631,3 +1634,76 @@ def clear_db(self, tables=None):
...
+            x,y,z= row.adjustedX, row.adjustedY, row.adjustedZ
+            lon, lat, alt = pyproj.transform(ecef, lla, x, y, z)
+
+            point = Points(identifier=id,
+                           active=not row.pointignore, # active = ~ignored
+                           apriori= shapely.geometry.Point(float(row.aprioriX), float(row.aprioriY), float(row.aprioriZ)),
+                           adjusted= shapely.geometry.Point(float(row.adjustedX),float(row.adjustedY),float(row.adjustedZ)),
+                           pointtype=float(row.pointtype))
+
+            point.measures = list(measures)
```
jlaura: Once #345 is merged this should be refactored. This way the session instantiation is right next to the use of the session. Right now, if this apply takes a long time, the session could be lost. Not something to touch in this PR.

jlaura:
I have to log off for a while, but as far as I am concerned, this is good to go.

PR #354, opened 2019-06-28
updates to fundamental matrix

jlaura:


autocnet/graph/edge.py
```diff
@@ -750,11 +750,12 @@ def _from_db(self, table_obj):
 
     @property
     def masks(self):
-        res = Session().query(Edges.masks).\
+        session = Session()
+        res = session.query(Edges.masks).\
```
jlaura: Every change in edge.py should be to remove `Session()` and replace with making the session into a variable and then closing it.

autocnet/graph/node.py
```diff
@@ -508,6 +508,7 @@ def __init__(self, *args, parent=None, **kwargs):
 
         # For now, just use the PATH to determine if the node/image is in the DB
         res = session.query(Images).filter(Images.path == kwargs['image_path']).first()
+        session.close()
```
jlaura: Every change in here is session related as well.

bin/acn_submit
```diff
@@ -8,7 +8,6 @@ import sys
 from redis import StrictRedis
 import yaml
 
-from autocnet.io.db import connection
```
jlaura: connection was never used.

jessemapel:
I am okay with the coverage decrease here. On a side note, I wonder if there's a way to check if there are any dangling sessions after the tests run.

jlaura:
`SELECT sum(numbackends) FROM pg_stat_database;`

jessemapel:
@jlaura We'd expect just 1 from that for the actual query

jlaura:
I think so. We almost want a teardown test after every DB test that runs that query and errors if > 1. This assumes too that the DB is also releasing the session (and not setting it to idle) as soon as the client disconnect.

I think we need an issue on this to do some experimentation in the future.

Kelvinrr:
Would it possible to create scoped sessions that auto-close when they are out of scope? That way we don't have all these close calls polluting the code base.

jlaura:
@Kelvinrr Yes. I think that that is the next iteration on this. Here is the example from the docs:

```
### another way (but again *not the only way*) to do it ###

from contextlib import contextmanager

@contextmanager
def session_scope():
    """Provide a transactional scope around a series of operations."""
    session = Session()
    try:
        yield session
        session.commit()
    except:
        session.rollback()
        raise
    finally:
        session.close()


def run_my_program():
    with session_scope() as session:
        ThingOne().go(session)
        ThingTwo().go(session)
```

I think putting this into the root namespace makes sense. Then we would not import session, but session_scope and go from there.

PR #357, opened 2019-06-30
refactors db to use context manager

jlaura:


jlaura:
Apologies that this is a bunch of files. We had the DB stuff scatter all around, so the refactor to use a context was more involved than I originally anticipated. Everything should be pushed into one place now and the `__init__` should be cleaner.

autocnet/__init__.py
```diff
@@ -41,27 +42,16 @@
...
-except: 
-    def sessionwarn():
-        raise RuntimeError('This call requires a database connection.')
-    
-    Session = sessionwarn
-    engine = sessionwarn
-
+# Funcs/Objs for the root namespace
+from autocnet.io.db.connection import session_scope, engine
+from autocnet.graph.network import CandidateGraph, NetworkCandidateGraph
```
jlaura: Pulling all of the DB stuff out into a separate file so that the __init__ is not cluttered. Then pushing the NCG, and DB items into the root namespace.

autocnet/__init__.py
```diff
@@ -41,27 +42,16 @@
...
 
+# Check to see if the desired DB exists. If not, create it.
+from autocnet.io.db.connection import Session as _Session
+from autocnet.io.db.createdb import createdb as _createdb
+_createdb(_Session, engine)
```
jlaura: No reason for the DB creation to be inside the models. This pushes the creation into a separate file and then calls it from here when the library is imported.

autocnet/graph/edge.py
```diff
@@ -10,7 +10,7 @@
 from shapely.geometry import Point
 import sqlalchemy
 
-from autocnet import Session, engine
+from autocnet import session_scope, engine
```
jlaura: All db access is not via the `session_scope` context. No more manual session creation and closing necessary.

autocnet/graph/edge.py
```diff
@@ -741,21 +741,20 @@ def _from_db(self, table_obj):
         table_obj : object
                     The declared table class (from db.model)
         """
-        session = Session()
-        res = session.query(table_obj).\
-               filter(table_obj.source == self.source['node_id']).\
-               filter(table_obj.destination == self.destination['node_id'])
-        session.close()
+        with session_scope() as session:
```
jlaura: This is what all of the calls looks like.

autocnet/graph/edge.py
```diff
@@ -731,31 +731,13 @@ def __init__(self, *args, **kwargs):
...
-        table_obj : object
-                    The declared table class (from db.model)
-        """
-        session = Session()
-        res = session.query(table_obj).\
-               filter(table_obj.source == self.source['node_id']).\
-               filter(table_obj.destination == self.destination['node_id'])
-        session.close()
-        return res
-
```
jlaura: This has to go because the `res` object is not accessible outside of the session scope. This is because the res is lazy loaded and we want the session to go out of scope as quickly as is reasonable. Therefore, wrapping the query in a func with a context is not possible because the return is immediately out of scope.

autocnet/graph/edge.py
```diff
@@ -731,31 +731,13 @@ def __init__(self, *args, **kwargs):
...
-        res = session.query(Edges.masks).\
-                                        filter(Edges.source == self.source['node_id']).\
-                                        filter(Edges.destination == self.destination['node_id']).\
-                                        first()
-        session.close()
+        with session_scope() as session:
+            res = session.query(Edges.masks).\
+                                            filter(Edges.source == self.source['node_id']).\
+                                            filter(Edges.destination == self.destination['node_id']).\
+                                            first()
```
jlaura: This is the result of the above removal - the property needs to make the query it needs to get a return.

autocnet/graph/edge.py
```diff
@@ -815,53 +794,49 @@ def costs(self):
...
-                    if np.isnan(v):
-                        v = None
-                    costs[k] = v
-                cost = Costs(match_id=idx, _cost=costs)
-                to_db_add.append(cost)
-        if to_db_add:
-            session.bulk_save_objects(to_db_add)
-        session.commit()
-        session.close()
+        with session_scope() as session:
```
jlaura: This is all 'changed' because the call is happening inside the context.

autocnet/graph/edge.py
```diff
@@ -815,53 +794,49 @@ def costs(self):
...
-                      Matches.destination == self.destination['node_id'])
-        odf = pd.read_sql(qf.statement, q.session.bind).set_index('id')
-        df = pd.DataFrame(odf.values, index=odf.index.values, columns=odf.columns.values)
-        df.index.name = 'id'
-        # Explicit close to get the session cleaned up
-        session.close()
-        return DbDataFrame(df,
-                           parent=self,
-                           name='matches')
+        with session_scope() as session:
```
jlaura: As above - the scope of change is much smaller than the coloring indicates.

autocnet/graph/network.py
```diff
@@ -1590,19 +1589,19 @@ def from_database(cls, query_string='SELECT * FROM public.images'):
...
-        for r in res:
-            sid, spath, did, dpath = r
-
-            adjacency_lookup[spath] = sid
-            adjacency_lookup[dpath] = did
-            if spath != dpath:
-                adjacency[spath].append(dpath)
-        session.close()
+
+        with session_scope() as session:
```
jlaura: Small change - just using the context.

autocnet/graph/network.py
```diff
@@ -1681,7 +1672,8 @@ def get_measures(row):
                          linesigma=float(row.linesigma),
                          samplesigma=float(row.samplesigma))
 
-            measures = cnetpoint.apply(get_measures, axis=1)
+            with session_scope() as session:
+                measures = cnetpoint.apply(get_measures, args=(session,), axis=1)
```
jlaura: Passing the session into the apply to avoid lots of session creation.

autocnet/graph/network.py
```diff
@@ -1695,9 +1687,8 @@ def get_measures(row):
 
             point.measures = list(measures)
             points.append(point)
-        session.add_all(points)
-        session.commit()
-        session.close()
+
+        Points.bulkadd(points)
```
jlaura: Uses the new bulkadd syntax.

autocnet/graph/node.py
```diff
@@ -600,9 +589,10 @@ def nkeypoints(self):
         """
         Get the number of keypoints from the database
         """
-        res = self._from_db(Keypoints)
-        nkps = res.nkeypoints
-        return nkps
+        with session_scope() as session:
```
jlaura: This is another from_db refactor

autocnet/graph/tests/test_network_graph.py
```diff
@@ -39,16 +38,14 @@ def cnet():
 def test_creation():
     ncg = NetworkCandidateGraph()
 
-
 @pytest.mark.parametrize("image_data, expected_npoints", [({'id':1, 'serial': 'BRUH'}, 1)])
 def test_place_points_from_cnet(session, cnet, image_data, expected_npoints):
-    model.Images.create(session, **image_data)
+    model.Images.create(**image_data)
```
jlaura: The tests get a little weird with the scoped session, but I didn't want to have the tests all the sudden using the internal _Session API. I say weird because, the `.create` method on the obj uses the scoped session internally. Then the test uses a second scoped session to do the query for asserting. The `create` method no longer returns anything, because the `obj` that is created inside the session is disconnected once the create returns. Therefore, we can't do anything with the return any way. I would argue this is cleaner - if the obj is created, it is just created and then it is the DB responsibility to manage the lifecycle. If we then want to access the obj, we need to query for it.

autocnet/io/db/connection.py
```diff
@@ -2,38 +2,45 @@
...
+
+except: 
+    def sessionwarn():
+        raise RuntimeError('This call requires a database connection.')
+    
+    Session = sessionwarn
+    engine = sessionwarn
+
+
+@contextmanager
```
jlaura: The magic context

autocnet/io/db/createdb.py
```diff
@@ -0,0 +1,36 @@
...
+        event.listen(Images.__table__, 'after_create', valid_geom_trigger)
+
+    Base.metadata.bind = engine
+    # If the table does not exist, this will create it. This is used in case a
+    # user has manually dropped a table so that the project is not wrecked.
+    Base.metadata.create_all(tables=[Overlay.__table__,
+                                    Edges.__table__, Costs.__table__, Matches.__table__,
+                                    Cameras.__table__, Points.__table__,
+                                    Measures.__table__, Images.__table__,
+                                    Keypoints.__table__])
```
jlaura: I refactored this out of the `__init__`.

autocnet/io/db/model.py
```diff
@@ -31,18 +31,25 @@
 
 class BaseMixin(object):
     @classmethod
-    def create(cls, session, **kw):
+    def create(cls, **kw):
+        from autocnet import session_scope
```
jlaura: imports here to avoid circular imports.

autocnet/io/db/tests/test_model.py
```diff
@@ -80,21 +64,23 @@ def test_images_exists(tables):
...
     ])
 def test_create_images(session, data):
-    i = model.Images.create(session, **data)
-    resp = session.query(model.Images).filter(model.Images.id==i.id).first()
-    assert i == resp
+    model.Images.create(**data)
+    with session() as session:
+        resp = session.query(model.Images).first()
+        for k, v in data.items():
+            assert getattr(resp, k) == v
```
jlaura: We can't do an obj to resp comparison any more, so we have to do attribute comparisons. This is because the obj that is originally inserted is out of scope by the time the query is made.

autocnet/graph/edge.py
```diff
@@ -870,8 +845,8 @@ def matches(self, v):
         df = pd.DataFrame(v)
         df.index.name = v.index.name
         # Get the query obj
-        session = Session()
-        q = session.query(Matches)
+        with session_scope() as session:
+            q = session.query(Matches)
```
jessemapel: `q` needs to be accessed here because the session will be gone by the time it is accessed.

autocnet/graph/network.py
jessemapel: This will still create a new session every iteration, You need to define the session scope outside of the for loop above.

autocnet/io/db/model.py
jessemapel: How is this a circular import? The session_scope doesn't use the model.

bin/acn_overlaps
```diff
@@ -25,27 +25,25 @@ except:
...
+
     
-    print('Placing points in overlap', id)
     points = place_points_in_overlap(nodes, overlap.geom, dem, msg["cam_type"],
                                      msg['iterative_phase_kwargs'],
                                      msg['distribute_points_kwargs'])
-    session.add_all(points)
-    session.commit()
-    session.close()
+    Points.bulkadd(points)
```
jessemapel: This needs to be in a session context now.

conftest.py
```diff
@@ -145,22 +145,21 @@ def tables():
...
-        for t in reversed(tables):
-            # Skip the srid table
-            if t != 'spatial_ref_sys':
-                session.execute(f'TRUNCATE TABLE {t} CASCADE')
-            # Reset the autoincrementing
-            if t in ['Images', 'Cameras', 'Matches', 'Measures']:
-                session.execute(f'ALTER SEQUENCE {t}_id_seq RESTART WITH 1')
-        session.commit()
+        with session_scope() as session:
+            #session.rollback()  # Necessary because some tests intentionally fail
```
jessemapel: why is this no longer needed? Because the context already does it in case of an exception?

autocnet/io/db/model.py
jlaura: If we include the autocnet import at the root then components inside of autocnet also import the model. We have to disconnect either autocnet from the model or the model from autocnet. This approach is the latter.

bin/acn_overlaps
jlaura: It happens inside of the bulkadd method

conftest.py
jlaura: yup - the context handles all of the rollback capability.

autocnet/io/db/model.py
```diff
@@ -268,11 +275,11 @@ def overlapping_larger_than(cls, size_threshold):
...
-        session = Session()
-        res = session.query(cls).\
-                filter(sqlalchemy.func.ST_Area(cls.geom) >= size_threshold).\
-                filter(sqlalchemy.func.array_length(cls.intersections, 1) > 1)
-        session.close()
+        from autocnet import session_scope
+        with session_scope() as session:
+            res = session.query(cls).\
+                    filter(sqlalchemy.func.ST_Area(cls.geom) >= size_threshold).\
+                    filter(sqlalchemy.func.array_length(cls.intersections, 1) > 1)
```
jessemapel: This needs to convert the response into actual overlaps before it returns. Right now it returns the response which cannot lazy load outside of the scope.

bin/acn_overlaps
```diff
@@ -25,27 +25,25 @@ except:
...
+        geom = overlap.geom
+        nodes = []
+        
+        for id in overlap.intersections:
+            res = session.query(Images).filter(Images.id == id).first()
+            nodes.append(NetworkNode(node_id=id, image_path=res.path))
+
     
-    print('Placing points in overlap', id)
     points = place_points_in_overlap(nodes, overlap.geom, dem, msg["cam_type"],
```
Ethands123: overlap.geom needs to be changed to geom

bin/acn_overlaps
```diff
@@ -25,27 +25,25 @@ except:
...
                                      msg['iterative_phase_kwargs'],
                                      msg['distribute_points_kwargs'])
-    session.add_all(points)
-    session.commit()
-    session.close()
+    Points.bulkadd(points)
     
     print('Placing points in overlap', id)
     points = place_points_in_overlap(nodes, overlap.geom, dem, msg["cam_type"],
```
Ethands123: This should also be removed. We don't want to be adding points twice

autocnet/graph/edge.py
jlaura: In the current version PRed in this should be working as intended since the entire block occurs within the context.

PR #358, opened 2019-07-01
some fixes for demo.ipynb

Kelvinrr:
some minor bugs squashed

### DIFF
--- a/autocnet/spatial/overlap.py
+++ b/autocnet/spatial/overlap.py
@@ -66,11 +66,11 @@ def place_points_in_overlaps(nodes, size_threshold=0.0007,
         overlaps = o.intersections
         if overlaps == None:
             continue
-        overlapnodes = [nodes[id] for id in overlaps]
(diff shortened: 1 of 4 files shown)


jlaura:
Once this passes - looks good here.

autocnet/spatial/overlap.py
```diff
@@ -66,11 +66,11 @@ def place_points_in_overlaps(nodes, size_threshold=0.0007,
         overlaps = o.intersections
         if overlaps == None:
             continue
-        overlapnodes = [nodes[id] for id in overlaps]
+        overlapnodes = [nodes[id]["data"] for id in overlaps]
```
Kelvinrr: probably not a bug per se, but this is the difference between `place_points_on_overlaps(ncg.nodes)` and  `place_points_in_overlaps(ncg.nodes(data="data"))`

jlaura: :+1: Good call here. I bet this is the source of the differences that we all are seeing. I have been in the habit of using the latter. Good find.

PR #365, opened 2019-07-03
Refactored iterative phase call from place_points_in_overlap

kaitlyndlee:
I also removed the DEM parameter since it is part of the config and accessible in place_points_in_overlap by importing it instead of passing it in from the calling functions.

### DIFF
--- a/autocnet/spatial/overlap.py
+++ b/autocnet/spatial/overlap.py
@@ -170,9 +170,9 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
                        adjusted=geom,
                        pointtype=2) # Would be 3 or 4 for ground
         
+        gnd = csmapi.EcefCoord(x, y, z)
         for node in nodes:
             if cam_type == "csm":
-                gnd = csmapi.EcefCoord(x, y, z)
                 image_coord = node.camera.groundToImage(gnd)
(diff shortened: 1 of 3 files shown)


autocnet/spatial/overlap.py
```diff
@@ -185,42 +160,28 @@ def place_points_in_overlap(nodes, geom, dem=dem, cam_type="csm",
...
-                                       measuretype=3))
-
-
-        for i, dest in enumerate(nodes):
+        
+        for node in nodes:
             if cam_type == "csm":
-                dic = dest.camera.groundToImage(gnd)
-                dline, dsample = dic.line, dic.samp
+                gnd = csmapi.EcefCoord(x, y, z)
```
jessemapel: You can actually create this up by where the x,y,z are computed so that's only created once. This will re-created the same thing every iteration.

autocnet/spatial/tests/test_overlap_points.py
```diff
@@ -6,9 +6,9 @@
 import csmapi
 
 
-@patch('autocnet.spatial.overlap.iterative_phase', return_value=(0, 1, 2))
+# @patch('autocnet.spatial.overlap.iterative_phase', return_value=(0, 1, 2))
```
jessemapel: you should just remove this instead of commenting it out

PR #367, opened 2019-07-05
Serialize

jlaura:
This adds function serialization for the cluster computation methods.  Usage is:

* no imports: just use a lambda
* with imports: for example, math.ceil. Here you want to pass a function in. For example:

```python
# This block is used to compute the overlapping polygon components and then place points into them.
from autocnet.spatial.overlap import cluster_place_points_in_overlaps

def ns(x):
    from math import ceil
    return ceil(round(x,1)*20)

# Place points
distribute_points_kwargs = {'nspts_func':ns}
cluster_place_points_in_overlaps(distribute_points_kwargs=distribute_points_kwargs)
```

### DIFF
--- a/autocnet/__init__.py
+++ b/autocnet/__init__.py
@@ -23,17 +23,19 @@ except DistributionNotFound:
 else:
     __version__ = _dist.version
 
+# Defaults
+dem = None
+
 from config_parser import parse_config
 
 config = parse_config()
 
 if config:
     dem = config['spatial']['dem']
-    if dem:
-        try:
-            dem = GeoDataset(dem)
-        except:
-            dem = None
+    try:
+        dem = GeoDataset(dem)
+    except:
+        dem = None
 
     db_uri = '{}://{}:{}@{}:{}/{}'.format(config['database']['type'],
                                             config['database']['username'],
@@ -47,7 +49,6 @@ if config:
                     isolation_level="AUTOCOMMIT")                   
     Session = orm.session.sessionmaker(bind=engine)
 else:
-    dem = None
     def sessionwarn():
         raise RuntimeError('This call requires a database connection.')
     Session = sessionwarn
--- a/autocnet/__init__.py
+++ b/autocnet/__init__.py
@@ -35,6 +35,7 @@ if config:
     try:
         dem = GeoDataset(dem)
     except:
+        warnings.warn(f'Unable to load the desired DEM: {dem}.')
         dem = None
 
     db_uri = '{}://{}:{}@{}:{}/{}'.format(config['database']['type'],
--- a/autocnet/__init__.py
+++ b/autocnet/__init__.py
@@ -23,25 +23,18 @@ except DistributionNotFound:
 else:
     __version__ = _dist.version
 
-#Load the config file and setup a global DB session factory
-try:
-    with open(os.environ['autocnet_config'], 'r') as f:
-        config = yaml.safe_load(f)
-except:
-    warnings.warn('No autocnet_config environment variable set. Defaulting to an empty configuration.')
-    config = {}
+from config_parser import parse_config
 
-if 'dem' in config['spatial']:
+config = parse_config()
+
+if config:
     dem = config['spatial']['dem']
-    try:
-        dem = GeoDataset(dem)
-    except:
-        warnings.warn(f'Unable to load the dem: {dem}')
-        dem = None
-else:
-    dem = None
+    if dem:
+        try:
+            dem = GeoDataset(dem)
+        except:
+            dem = None
 
-try:
     db_uri = '{}://{}:{}@{}:{}/{}'.format(config['database']['type'],
                                             config['database']['username'],
                                             config['database']['password'],
@@ -53,14 +46,15 @@ try:
                     connect_args={"application_name":"AutoCNet_{}".format(hostname)},
                     isolation_level="AUTOCOMMIT")                   
     Session = orm.session.sessionmaker(bind=engine)
-except: 
+else:
+    dem = None
     def sessionwarn():
         raise RuntimeError('This call requires a database connection.')
-    
     Session = sessionwarn
     engine = sessionwarn
 
-
+# Patch the candidate graph into the root namespace
+from autocnet.graph.network import CandidateGraph, NetworkCandidateGraph
 
 import autocnet.examples
 import autocnet.camera
@@ -71,9 +65,6 @@ import autocnet.matcher
 import autocnet.transformation
 import autocnet.utils
 import autocnet.spatial
-
-# Patch the candidate graph into the root namespace
-from autocnet.graph.network import CandidateGraph
 
 def get_data(filename):
     packagdir = autocnet.__path__[0]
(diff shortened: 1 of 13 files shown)


autocnet/io/db/triggers.py
```diff
@@ -61,8 +61,11 @@
 EXECUTE PROCEDURE validate_points();
 """)
 
-latitudinal_srid = config['spatial']['latitudinal_srid']
-
+try:
+  latitudinal_srid = config['spatial']['latitudinal_srid']
+except:
+  latitudinal_srid = None
+
```
ladoramkershner: Isn't latitudinal_srid needed in autocnet operations? Wouldn't we want it to error early, not provide a None value, which may be incorrectly accepted by functions down the line?

autocnet/io/db/model.py
```diff
@@ -383,7 +366,7 @@ def measuretype(self, v):
             v = MeasureType(v)
         self._measuretype = v
 
-if Session:
+if isinstance(Session, sqlalchemy.orm.sessionmaker):
     from autocnet.io.db.triggers import valid_point_function, valid_point_trigger, update_point_function, update_point_trigger, valid_geom_function, valid_geom_trigger
```
ladoramkershner: Why the specification? Does the variable 'Session' exist in the code as something other than an sqlalchemy.orm.sessionmaker?

autocnet/utils/serializers.py
```diff
@@ -0,0 +1,37 @@
...
+import datetime
+import json
+
+import dill
+import numpy as np
+import shapely
+
+class JsonEncoder(json.JSONEncoder):
+    def default(self, obj):
+        if isinstance(obj, np.ndarray):
```
ladoramkershner: JsonEncoder already exist under model.py, why duplicate? If it does not make sense to have it in model.py anymore (because the scope of its use is increasing) should we move it to a more intuitive spot?

autocnet/__init__.py
```diff
@@ -23,25 +23,18 @@
...
+if config:
     dem = config['spatial']['dem']
-    try:
-        dem = GeoDataset(dem)
-    except:
-        warnings.warn(f'Unable to load the dem: {dem}')
-        dem = None
-else:
-    dem = None
+    if dem:
```
acpaquette: Shouldn't need this if dem check here

autocnet/__init__.py
```diff
@@ -53,14 +46,15 @@
                     connect_args={"application_name":"AutoCNet_{}".format(hostname)},
                     isolation_level="AUTOCOMMIT")                   
     Session = orm.session.sessionmaker(bind=engine)
-except: 
+else:
+    dem = None
```
acpaquette: Could just make this the default outside of the if else block but it isn't a huge concern

autocnet/utils/serializers.py
acpaquette: It's being moved out of model.py in this same PR

autocnet/io/db/triggers.py
jlaura: Yes. But, what about running this without a db or without tests? I think the solution here is to have the check happen in the config parser. Then this can safely default to None.

autocnet/io/db/model.py
jlaura: Because if, in the `__init__` the Session is set to the sessionwarn function, then the if statement will fail. We can not '`if <function>`. We can though, check to see if the session is either an instance of the sessionmaker or something else.

autocnet/spatial/overlap.py
```diff
@@ -85,23 +85,35 @@ def cluster_place_points_in_overlaps(size_threshold=0.0007,
...
+    ids = session.query(Overlay.id).all()
+    session.close()
+    for i, id in enumerate(ids):
+        msg = {'id' : id[0],
                'distribute_points_kwargs' : distribute_points_kwargs,
                'walltime' : walltime,
                'cam_type': cam_type}
         rqueue.rpush(queuename, json.dumps(msg, cls=JsonEncoder))
-    job_counter = len([*overlaps]) + 1
+        if (i > 0) and i % 10000 == 0:
```
acpaquette: I thought this was something we wanted to put in plurmy?

jlaura: crud. It is. This was just supposed to be a local change for now. I'll revert.

acpaquette:
Works locally, other than the stuff that should be in plurmy this looks good to merge

config_parser.py
```diff
@@ -0,0 +1,32 @@
+import os
+import yaml
+
+def parse_config(name='autocnet_config'):
+    if name not in os.environ.keys():
+        return {}
+    
+    filepath = os.environ[name]
+    if not os.path.exists(filepath):
+        return {}
```
Kelvinrr: If we are going for an exception on variable set but malformed file, we should probably raise some exception if the env var is set but the file doesn't exist.

autocnet/__init__.py
```diff
@@ -23,25 +23,20 @@
...
+
+config = parse_config()
 
-if 'dem' in config['spatial']:
+if config:
     dem = config['spatial']['dem']
     try:
         dem = GeoDataset(dem)
     except:
-        warnings.warn(f'Unable to load the dem: {dem}')
```
Kelvinrr: Should probably keep the warning

autocnet/__init__.py
```diff
@@ -53,14 +48,14 @@
                     connect_args={"application_name":"AutoCNet_{}".format(hostname)},
                     isolation_level="AUTOCOMMIT")                   
     Session = orm.session.sessionmaker(bind=engine)
-except: 
+else:
     def sessionwarn():
         raise RuntimeError('This call requires a database connection.')
-    
     Session = sessionwarn
```
ladoramkershner: This error that specifically warns about the session is raised if the config does not exist. Config is responsible for more than just the session setup, should the warning be more general?

autocnet/__init__.py
jlaura: I decided to throw if the config_parser and then also warn here. This way, we can force a DEM. Then, if the user specifies either None or a bad path, they get a warning that they can chose to ignore if intentional.

autocnet/__init__.py
jlaura: Updated with something more verbose to assist a user in debugging.

autocnet/matcher/subpixel.py
```diff
@@ -549,5 +550,5 @@ def cluster_subpixel_register_points(iterative_phase_kwargs={'size': 71},
                  time=walltime,
                  partition=config['cluster']['queue'],
                  output=config['cluster']['cluster_log_dir']+'/slurm-%A_%a.out')
-    submitter.submit(array='1-{}'.format(job_counter))
+    submitter.submit(array='1-{}'.format(job_counter), chunksize=chunksize)
     return job_counter
```
ladoramkershner: Would array='1-{}%{}'.format(job_counter, chunksize), not work with out making any changes to plurmy?

jlaura: I do not believe so. That would like the number of concurrent jobs, but not allow for the submission of more jobs than the current job limit.

ladoramkershner: Ohhhh, so the cluster will fail if you submit over X jobs, even if you limit the concurrent jobs to below that limit? The whole issue makes a lot more sense..

jlaura: yup.

PR #373, opened 2019-07-15
Serialize

jlaura:


autocnet/spatial/overlap.py
```diff
@@ -69,7 +69,8 @@ def cluster_place_points_in_overlaps(size_threshold=0.0007,
                                      distribute_points_kwargs={},
                                      walltime='00:10:00',
                                      chunksize=1000,
-                                     cam_type="csm"):
+                                     cam_type="csm",
+                                     query_string='SELECT overlay.id FROM overlay LEFT JOIN points ON ST_INTERSECTS(overlay.geom, points.geom) WHERE points.id IS NULL;'):
```
jlaura: This query string allows placing points into a subset of the overlaps. The default is now to place points into overlaps that do not already contain points.

autocnet/spatial/overlap.py
```diff
@@ -96,10 +99,10 @@ def cluster_place_points_in_overlaps(size_threshold=0.0007,
     queuename = config['redis']['processing_queue']
     past = 0
     session = Session()
-    ids = session.query(Overlay.id).all()
+    ids = [i[0] for i in session.execute(query_string)]
```
jlaura: Change from the ORM to the Core.

bin/acn_load_images
```diff
@@ -0,0 +1,126 @@
...
+    Images.bulkadd(images)
+
+if __name__ == '__main__':
+    conf = config['redis']
+    queue = StrictRedis(host=conf['host'], port=conf['port'], db=0)
+    
+    msg = pop_computetime_push(queue,
+                               conf['processing_queue'],
+                               conf['working_queue'])
+    main(msg, config)
```
jlaura: This is a submission script that will bulk add images into a database. This was used to get all of the CTX images into the DB.

bin/acn_overlaps
```diff
@@ -24,20 +24,19 @@ except:
 
 def main(msg, config):
     session = Session()
-    id = msg['id']
-    res = session.query(Overlay).filter(Overlay.id == msg['id'])
-    if res is None:
+    oid = msg['id']
+    overlap = session.query(Overlay).filter(Overlay.id == oid).one()
+    if overlap is None:
```
jlaura: Cleanup to make sure overlay entries are unique.

autocnet/io/db/connection.py
```diff
@@ -12,10 +12,11 @@
 
 class Parent:
     def __init__(self, config):
-        self.session, _ = new_connection(config)
+        Session, _ = new_connection(config)
+        self.session = Session()
         self.session.begin()
```
ladoramkershner: Below you just specified that new_connection is called with dbconfig, but that change is not reflected here. (Unless you replaced config with config['database'] somewhere before this call)

jlaura: See line 502 in node.py (above).

ladoramkershner: :+1:

PR #375, opened 2019-07-16
Cnet generation

Kelvinrr:


autocnet/io/db/tests/test_model.py
```diff
@@ -127,18 +127,19 @@ def test_create_point(session, data):
...
     resp = session.query(model.Points).filter(model.Points.id == p.id).first()
+
     assert resp.geom == expected
 
 @pytest.mark.parametrize("data, new_adjusted, expected", [
-    ({'pointtype':3, 'adjusted':Point(0,-1,0)}, None, None),
-    ({'pointtype':3, 'adjusted':Point(0,-1,0)}, Point(0,1,0), Point(90, 0)),
-    ({'pointtype':3}, Point(0,-1,0), Point(-90, 0))
+    ({'pointtype':3, 'adjusted':Point(0,-100000,0)}, None, None),
+    ({'pointtype':3, 'adjusted':Point(0,-100000,0)}, Point(0,100000,0), Point(90, 0)),
```
Kelvinrr: pyproj seems to return `inf` if the point isn't close to the surface.

autocnet/io/db/model.py
```diff
@@ -367,7 +374,8 @@ def measuretype(self, v):
         self._measuretype = v
 
 if isinstance(Session, sqlalchemy.orm.sessionmaker):
-    from autocnet.io.db.triggers import valid_point_function, valid_point_trigger, update_point_function, update_point_trigger, valid_geom_function, valid_geom_trigger
+    from autocnet.io.db.triggers import valid_point_function, valid_point_trigger, update_point_function, valid_geom_function, valid_geom_trigger
```
jlaura: `update_point_trigger` is removed but not the function. How come?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -11,159 +11,183 @@
 import pandas as pd
 import scipy
 from matplotlib import pyplot as plt
-from skimage.transform import resize
+from scipy.misc import imresize
```
jlaura: scipy version needs to be pinned in the environment.yml for this to work I believe

Kelvinrr: it should be pinned

autocnet/io/db/model.py
Kelvinrr: `update_point_trigger` was removed from `triggers.py` if that's what you mean.

jlaura: I mean, we should also remove the function from being added to the DB if we aren't going to use the trigger.

Kelvinrr: I see what you mean now, should be removed now.

PR #379, opened 2019-07-23
add warning message for empty redis queue

ladoramkershner:


autocnet/io/db/redis_queue.py
```diff
@@ -27,8 +28,15 @@ def pop_computetime_push(queue, inqueue, outqueue):
     msg : dict
           The message from the processing queue.
     """
+
+    # Check if the redis queue is empty
+    msg = queue.rpop(inqueue)
+    if msg is None:
+        warnings.warn('Expected to process a cluster job, but the message queue is empty.')
+        sys.exit()
```
jlaura: This should not sys.exit inside of this function. It should return None. Then the caller (the bin scripts in our case) should check for None and behave in some way.

jlaura:
:+1: I think this just needs to get added to the other bin scripts as well and then we can merge. :+1:

PR #381, opened 2019-07-26
lower requested memory to 1Gb

ladoramkershner:


config/sample.yml
```diff
@@ -12,8 +12,8 @@ cluster:
     tmp_scratch_dir: '/scratch/jlaura'
 
     # The amount of RAM (in MB) to request for jobs
-    extractor_memory: 8192
-    processing_memory: 8192
+    extractor_memory: 1000
+    processing_memory: 1000
```
jlaura: Can this be 1024. That is base 2.

PR #385, opened 2019-08-16
leaky session fix, populating new columns in DB

ladoramkershner:
I AM A LIAR :smiling_imp: 

These are just fast changes, I would still love to add in a column that tracks the calls (and their kwargs) that massage the line and samples values of measures. That is just a little harder for me to figure out how to do.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1570,11 +1570,15 @@ WHERE points.active = True AND measures.active=TRUE AND measures.jigreject=FALSE
         if clear_db:
             cls.clear_db()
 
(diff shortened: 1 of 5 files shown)


autocnet/graph/network.py
```diff
@@ -1570,11 +1570,15 @@ def from_filelist(cls, filelist, clear_db=False):
         if clear_db:
             cls.clear_db()
 
+        cnt=1
+        total=len(filelist)
         for f in filelist:
```
jlaura: for cnt, f in enumerate(filelist):
    #now do the things

PR #386, opened 2019-08-19
Minor updates to fix bugs

jlaura:
Comments inline

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1474,7 +1474,7 @@ WHERE points.active = True AND measures.active=TRUE AND measures.jigreject=FALSE
         """
         df = pd.read_sql(sql, engine)
         df.rename(columns={'imageid':'image_index','id':'point_id', 'pointtype' : 'type',
-            'sample':'x', 'line':'y', 'serial': 'serialnumber'}, inplace=True)
+            'sample':'x', 'line':'y', 'serial': 'serialnumber', 'measuretype':'measure_type'}, inplace=True)
 
         #create columns in the dataframe; zeros ensure plio (/protobuf) will
         #ignore unless populated with alternate values
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1448,10 +1448,28 @@ class NetworkCandidateGraph(CandidateGraph):
             n.generate_vrt(**kwargs)
 
     def to_isis(self, path, flistpath=None,sql = """
-SELECT points.id, measures.serial, points.pointtype, points.apriori, points.adjusted,
-measures.sample, measures.line, measures.measuretype, measures.imageid
-FROM measures INNER JOIN points ON measures.pointid = points.id
-WHERE points.active = True AND measures.active=TRUE AND measures.jigreject=FALSE;
+SELECT points.id,
+        measures.serial,
+        points.pointtype, 
+        points.apriori, 
+        points.adjusted,
+        measures.sample, 
+        measures.line, 
+        measures.measuretype,
+        measures.imageid
+FROM measures 
+INNER JOIN points ON measures.pointid = points.id
+WHERE 
+    points.active = True AND 
+    measures.active=TRUE AND 
+    measures.jigreject=FALSE AND 
+    measures.imageid NOT IN
+        (SELECT measures.imageid
+        FROM measures 
+        INNER JOIN points ON measures.pointid = points.id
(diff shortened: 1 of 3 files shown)


autocnet/graph/network.py
```diff
@@ -1448,10 +1448,28 @@ def generate_vrts(self, **kwargs):
...
+    points.active = True AND 
+    measures.active=TRUE AND 
+    measures.jigreject=FALSE AND 
+    measures.imageid NOT IN
+        (SELECT measures.imageid
+        FROM measures 
+        INNER JOIN points ON measures.pointid = points.id
+        WHERE measures.active = true and measures.jigreject = false AND points.active = True
+        GROUP BY measures.imageid
+        HAVING COUNT(DISTINCT measures.pointid)  < 3);
```
jlaura: This change to the SQL query now auto drops images that do not have at least 3 measures.

autocnet/graph/network.py
```diff
@@ -1474,7 +1492,7 @@ def to_isis(self, path, flistpath=None,sql = """
         """
         df = pd.read_sql(sql, engine)
         df.rename(columns={'imageid':'image_index','id':'point_id', 'pointtype' : 'type',
-            'sample':'x', 'line':'y', 'serial': 'serialnumber'}, inplace=True)
+            'sample':'x', 'line':'y', 'serial': 'serialnumber', 'measuretype':'measure_type'}, inplace=True)
```
jlaura: I had to make this change for plio. Can you confirm?

autocnet/graph/network.py
```diff
@@ -1504,8 +1522,14 @@ def to_isis(self, path, flistpath=None,sql = """
             flistpath = os.path.splitext(path)[0] + '.lis'
         target = config['spatial'].get('target', None)
 
+        ids = df['image_index'].unique()
+        fpaths = [self.nodes[i]['data']['image_path'] for i in ids]
+        for f in self.files:
+            if f not in fpaths:
+                warnings.warn(f'{f} in candidate graph but not in output network.')
```
jlaura: If images get dropped we want to warn so that the user is not surprised having lest images.

autocnet/spatial/overlap.py
```diff
@@ -70,7 +70,7 @@ def cluster_place_points_in_overlaps(size_threshold=0.0007,
                                      walltime='00:10:00',
                                      chunksize=1000,
                                      cam_type="csm",
-                                     query_string='SELECT overlay.id FROM overlay LEFT JOIN points ON ST_INTERSECTS(overlay.geom, points.geom) WHERE points.id IS NULL;'):
+                                     query_string='SELECT overlay.id FROM overlay LEFT JOIN points ON ST_INTERSECTS(overlay.geom, points.geom) WHERE points.id IS NULL AND ST_AREA(overlay.geom) >= {};'):
```
jlaura: The overlap call was never using the area threshold, so this adds that back in.

autocnet/spatial/overlap.py
```diff
@@ -99,7 +99,7 @@ def cluster_place_points_in_overlaps(size_threshold=0.0007,
     queuename = config['redis']['processing_queue']
     past = 0
     session = Session()
-    ids = [i[0] for i in session.execute(query_string)]
+    ids = [i[0] for i in session.execute(query_string.format(size_threshold))]
```
jlaura: Here is where the size_threshold gets injected.

autocnet/graph/network.py
ladoramkershner: Could you clarify what you want me to confirm? I did not have to make this change for autocnet to work (if that is what you mean)

jlaura: That is what I mean...crud....I need to get my plio into the correct version. Ahhh!!!

ladoramkershner: I have not worked very much with plio, do you know what alterations you may have made that makes it so this is necessary? And if you have an idea, is it something that is going to be implemented moving forward?

Additionally, would you like me to test this using the little 13 image area to see if it works with my version of plio?

ladoramkershner: I am now getting this error on my build and had to put in this line of code. 

However, Jesse just put up a PR on plio to do some name unmangling. I am looking at it, but I am fairly unfamiliar with protobuff and plio, so I am not sure if that PR will make this portion of this PR unnecessary.

PR #389, opened 2019-08-23
Fixing naming conventions on jay's PR for plio communication

ladoramkershner:


autocnet/graph/network.py
```diff
@@ -1447,11 +1447,49 @@ def generate_vrts(self, **kwargs):
...
+    points.active = True AND
+    measures.active=TRUE AND
+    measures.jigreject=FALSE AND
+    measures.imageid NOT IN
+        (SELECT measures.imageid
+        FROM measures
+        INNER JOIN points ON measures.pointid = points.id
+        WHERE measures.active = true and measures.jigreject = false AND points.active = True
+        GROUP BY measures.imageid
+        HAVING COUNT(DISTINCT measures.pointid)  < 3);
```
jlaura: I think that we have a few options to avoid having to do 2 calls.

1. Rename the DB columns in the model - I actually think I like this the best
2. Use <columnname> AS <remapped_name> - I dislike this because it is a hard implicit coupling between the DB <-> autocnet <-> plio. The caller that wanted to make a change would need to know about all these pieces.

autocnet/graph/network.py
```diff
@@ -1472,9 +1510,18 @@ def to_isis(self, path, flistpath=None,sql = """
...
 
         """
-        df = pd.read_sql(sql, engine)
-        df.rename(columns={'imageid':'image_index','id':'point_id', 'pointtype' : 'type',
-            'sample':'x', 'line':'y', 'serial': 'serialnumber'}, inplace=True)
+        #since measures and points tables contain some of the same attributes
+        #read them in seperately and rename
+        points_df = pd.read_sql(points_sql, engine)
+        points_df.rename(columns={'pointtype' : 'pointType', 'jigreject' : 'pointJigsawRejected'}, inplace=True)
+        points_df['pointIgnore'] = ~points_df['active']
```
jlaura: why are these being computed?

PR #390, opened 2019-08-26
Update network.py

jlaura:
Removes cameras from `from_remote_database`.


ladoramkershner:
This is still under construction, isn't it?

Kelvinrr:
It seems Lauren and I were able to get the whole thing running? We had to manually comment this out so I think it is worth merging as is?

PR #399, opened 2019-10-02
ensuring that that source image is consistent for all measure

ladoramkershner:
… subpixel registers in a given overlap. Also a small clean up of a redundant function return.


autocnet/matcher/subpixel.py
```diff
@@ -230,9 +230,6 @@ def subpixel_template(sx, sy, dx, dy, s_img, d_img, image_size=(251, 251), templ
 
     return dx, dy, metrics
 
-    x_shift, y_shift, strength = naive_template.pattern_match(template, image, **kwargs)
-    return x_shift, y_shift, strength
-
```
ladoramkershner: Redundant return

PR #406, opened 2020-01-07
New logic for laying down ground points

ladoramkershner:
This is not well tested! It is being used for a CTX sprint aimed at improving the ground control functionality of autocnet.


jlaura:
Closing this after #407 was merged.

PR #409, opened 2020-01-28
General fix to autocnet

acpaquette:
Gets all test for autocnet passing!


autocnet/graph/edge.py
```diff
@@ -859,9 +859,7 @@ def matches(self):
         df.index.name = 'id'
         # Explicit close to get the session cleaned up
         session.close()
-        return DbDataFrame(df,
-                           parent=self,
-                           name='matches')
+        return DbDataFrame(df,  parent=self, name='matches')
```
Kelvinrr: 10/10 change :+1:

PR #411, opened 2020-01-30
Updates to hirise change detection 

Kelvinrr:


autocnet/cg/change_detection.py
```diff
@@ -11,56 +10,63 @@
 from autocnet.matcher.cpu_extractor import extract_features
 
 def image_diff(arr1, arr2):
-    diff = arr1-arr2
-    diff[np.isnan(diff)] = 0
+     arr1 = arr1.astype("float32")
+     arr2 = arr2.astype("float32")
+     arr1[arr1 == 0] = np.nan
+     arr2[arr2 == 0] = np.nan
```
acpaquette: Could this blow out parts of the image that aren't ISIS nulls?

Kelvinrr: We will need to add nodata support to bytescale. That seems to be the crux of the issue but this doesn't cause problems in the example images we are using.

acpaquette:
:+1: Good to merge

Issue #402, opened 2019-11-23
PyProj Defaults to -180to180 domain, but we need 0to360 sometimes


jlaura:
When we define a lat/lon/alt projection, for example [here](https://github.com/USGS-Astrogeology/autocnet/blob/dev/autocnet/spatial/overlap.py#L158) the default behavior is to default to -180to180. This causes issues when we use footprints that have been generated using ISIS that have a 0to360 domain. The issue is only inside of the DB and subsequent visualization as the control/bundle process stays in image space.

From the Proj docs, it looks like we want to be able to specify a `+lon_warp=180` parameter. Here is a [link to the docs](https://proj.org/usage/projections.html#longitude-wrapping). 

The quick and dirty fix would be to modify the signature of  [`place_points_in_overlap`](https://github.com/USGS-Astrogeology/autocnet/blob/dev/autocnet/spatial/overlap.py#L126). I am not a fan of this approach. A cleaner solution, I think, would be to centralize the transformation definitions ([here](https://github.com/USGS-Astrogeology/autocnet/blob/d8a1c2088f81434ab13892e4932f51514c17dc1b/autocnet/spatial/overlap.py), [here](https://github.com/USGS-Astrogeology/autocnet/blob/65bb30d6d9a1d5a6d0f82da4e36cff62fd510d60/autocnet/graph/edge.py), and [here](https://github.com/USGS-Astrogeology/autocnet/blob/d14b5c6a66bd91cd559514f666ece4a82f79033f/autocnet/graph/network.py)) into one centralized location where the config file can be used to define the domain. Options in the config could be `domain:180` or `domain:360`.

thareUSGS:
It seems a cleaner fix might be update ISIS3 to **optionally** support the generation of footprints and other label information using a Longitude Domain of 180. UPC is heading to support -180 to 180 as the default so it would help there also. I assume this change would need to made in caminfo and/or footprintinit.

jessemapel:
@thareUSGS We would probably need to make such a change at a the level of the footprint objects in ISIS. It would not be terribly hard, but it would likely require code changes in every application and object that works with footprints.

thareUSGS:
Yes probably not worth it. 

One option I have heard would be to support an **additional** (again optional) section with -180 to 180 information. Sounds lame, but at least the default would maintained and the user could have access to the optional -180 to 180 Longitude sections (including stats like Min/Max Longitude and an additional footprint). 

[...]

PR #416, opened 2020-02-28
Changing point projection domain from +/-180 to 0/360

ladoramkershner:
pyproj.Proj class defaults to the -180 to 180 domain. Therefore, when pyproj.transform is called to preform the transformation using those classes the output is in the 180 domain. Instead of chasing down all instances of pyproj and inserting the lon_wrap=180 to set the domain to 360, I consolidated all pyproj calls to instead go through the already implemented reproject() function in autocnet and updated only the pyproj calls within reproject() with the lon_wrap keyword.

This pull request assumes that the images coming in are in the 360 domain!! A method should later be implemented where the projection domain for the points is determined by the domain of the input image (see issue #402 for current discussion on that topic)

### DIFF
--- a/autocnet/graph/edge.py
+++ b/autocnet/graph/edge.py
@@ -5,7 +5,6 @@ from collections import defaultdict, MutableMapping, Counter
 import numpy as np
 import pandas as pd
 import networkx as nx
-# import pyproj
 from scipy.spatial.distance import cdist
 from shapely.geometry import Point
 import sqlalchemy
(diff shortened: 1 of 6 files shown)


jlaura:
@ladoramkershner Looks good here. Can we remove the commented out lines completely? I am good to merge after that.

ladoramkershner:
All good to go! I wanted to leave that code in place until it passed the CI, just in case my changes caused everything to explode. Thank you for the reminder to remove it!

PR #417, opened 2020-03-02
Dev

jlaura:


autocnet/matcher/subpixel.py
```diff
@@ -375,6 +374,69 @@ def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=251, reduction=11, conver
...
+        destination.ignore = True # Unable to phase match
+        return
+
+    new_template_x, new_template_y, template_metric = subpixel_template(source.sample,
+                                                            source.line,
+                                                            new_phase_x,
+                                                            new_phase_y,
+                                                            source_node.geodata,
+                                                            destination_node.geodata,
+                                                            **subpixel_template_kwargs)
```
Kelvinrr: From the AM convo, swap subpixel and phase.

autocnet/matcher/subpixel.py
```diff
@@ -375,6 +374,69 @@ def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=251, reduction=11, conver
             break
     return dx, dy, metrics
 
+def subpixel_register_measure(measureid, iterative_phase_kwargs={}, subpixel_template_kwargs={},
+                            cost_func=lambda x,y: 1/x**2 * y, threshold=0.005):
```
Kelvinrr: Do we care about Autoreg's gradient support? 
These params seem different from the ones I saw locally on your box? Is this fully up to date 
Why is the default threshold super low? I believe ISIS default is 0.7. 
Do we want to have a Z-score threshold as well?

Kelvinrr:
Just some stuff for now, not sure how much more this is expected to change as the changes to make our subpixel closer to ISIS's autoreg isn't ready yet.

autocnet/matcher/subpixel.py
jlaura: This is for a different func. A few thoughts on the autoreg:

- I didn't do gradients because we haven't used those yet. I would say an add on to this at a later time might be appropriate.
- The 0.2 you saw on my box was for the masking for center of gravity. The 0.7 default is the threshold for the correlation coefficient.
- I didn't do a Z-score yet. We could, to identify bland areas, but I would add later.

autocnet/matcher/subpixel.py
jlaura: :+1:

jlaura: I made the swap. A follow on PR to this should separate the phase/template from subpixel_register_point and subpixel_register_measure. I think we can also maybe go to a single submission script that has arguments for point/measure and template/phase.

autocnet/matcher/subpixel.py
```diff
@@ -7,7 +7,7 @@
 from plurmy import Slurm
 
 from autocnet import Session, config
-from autocnet.matcher import naive_template
+from autocnet.matcher.naive_template import pattern_match, pattern_match_autoreg
```
Kelvinrr: pattern_match_autoreg doesn't seem to exist?

autocnet/matcher/subpixel.py
```diff
@@ -212,6 +212,7 @@ def subpixel_template(sx, sy, dx, dy, s_img, d_img, image_size=(251, 251), templ
     See Also
     --------
     autocnet.matcher.naive_template.pattern_match : for the kwargs that can be passed to the matcher
+    autocnet.matcher.naive_template.pattern_match_autoreg : for the jwargs that can be passed to the autoreg style matcher
```
Kelvinrr: reference to autoreg left in

autocnet/matcher/subpixel.py
```diff
@@ -165,7 +165,7 @@ def subpixel_phase(template, search, **kwargs):
     (y_shift, x_shift), error, diffphase = register_translation(search, template, **kwargs)
     return x_shift, y_shift, (error, diffphase)
 
-def subpixel_template(sx, sy, dx, dy, s_img, d_img, image_size=(251, 251), template_size=(51,51), **kwargs):
+def subpixel_template(sx, sy, dx, dy, s_img, d_img, image_size=(251, 251), template_size=(51,51), method=pattern_match_autoreg, **kwargs):
```
Kelvinrr: same here

autocnet/matcher/subpixel.py
```diff
@@ -571,3 +634,46 @@ def cluster_subpixel_register_points(iterative_phase_kwargs={'size': 251},
                  output=config['cluster']['cluster_log_dir']+f'/autocnet.subpixel_register-%j')
     submitter.submit(array='1-{}'.format(job_counter), chunksize=chunksize, exclude=exclude)
     return job_counter
+
+def cluster_subpixel_register_measures(iterative_phase_kwargs={'size': 251},
```
ladoramkershner: I know we were talking about how phase works best on very small areas, I haven't worked with it much so I don't know what constitutes a small area.. but should we set this default size lower?

jlaura: @ladoramkershner :+1: we should open a new PR to make it smaller.

autocnet/spatial/overlap.py
```diff
@@ -170,6 +175,39 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
...
+            # The CSM conversion makes the LLA/ECEF conversion explicit
+            x, y, z = reproject([lon, lat, height],
+                                 semi_major, semi_minor,
+                                 'latlon', 'geocent')
+            gnd = csmapi.EcefCoord(x, y, z)
+            image_coord = node.camera.groundToImage(gnd)
+            sample, line = image_coord.samp, image_coord.line
+
+        # Extract ORB features in a sub-image around the desired point
+        size = 71
```
Kelvinrr: Might as well make this a parameter or optimally have it computed based on point density.

jlaura:
The latest push is failing because the patches on the `test_place_points_in_overlap` function are not being called. Anyone have any thoughts as to why? Full paths to the funcs are being provided in the decorators. Either I am misreading the docs pretty badly or the defined MagicMocks are interacting with the patch. Tried the patches as contexts as well, but same problem.

autocnet/spatial/tests/test_overlap_points.py
```diff
@@ -1,13 +1,20 @@
...
 from autocnet.graph.node import Node
 import csmapi
 
+MockKeypoints = namedtuple('Keypoints', ['x', 'y'])
+mockkeypoints = MockKeypoints(0,0)
 
 @patch('autocnet.cg.cg.distribute_points_in_geom', return_value=[(0, 0), (5, 5), (10, 10)])
-def test_place_points_in_overlap(point_distributer):
+@patch('autocnet.matcher.subpixel.clip_roi', return_value=np.zeros((3,3)))
+@patch('autocnet.matcher.cpu_extractor.extract_most_interesting', return_value=mockkeypoints)
```
jessemapel: These should be 'autocnet.spatial.overlap.cliproi` and `autocnet.spatial.overlap.extract_most_interesting` because you import them using a 'from ... import ...` statement.

https://docs.python.org/3/library/unittest.mock.html#where-to-patch

autocnet/spatial/overlap.py
```diff
@@ -170,6 +175,39 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
...
+        image, _, _ = clip_roi(node.geodata, sample, line, size_x=size, size_y=size)
+        interesting = extract_most_interesting(image, sample, line)
```
Kelvinrr: the signature is `extract_most_interesting(image, extractor_method='orb', extractor_parameters={'nfeatures':10})`, so this call should always fail.

jlaura:
The issue is line 206 in overlap.py - need to support both CSM and ISIS sensors there as well.

PR #419, opened 2020-03-04
Added ignore property to NetworkNodes

jessemapel:
Added ignore property to Nodes and added a db trigger to ignore all measures on an image when that image's node is ignored.

This does not ignore edges in "little autocnet"

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -627,8 +627,9 @@ class CandidateGraph(nx.Graph):
         if options[on] == self.edges_iter:
             obj = 2
         for elem in options[on](data=True):
-            if hasattr(elem, 'ignore') and getattr(elemen, 'ignore'):
-                res.append(function(elem[obj], *args, **kwargs))
+            if hasattr(elem[obj], 'ignore') and getattr(elem[obj], 'ignore'):
+                continue
+            res.append(function(elem[obj], *args, **kwargs))
 
         if out:
             out = res
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -585,7 +585,8 @@ class CandidateGraph(nx.Graph):
     def apply(self, function, on='edge', out=None, args=(), **kwargs):
         """
         Applys a function to every node or edge, returns collected return
-        values.
+        values. If applying to functions to nodes, then all ignored nodes
+        will be skipped.
 
         TODO: Merge with apply_func_to_edges?
 
@@ -626,7 +627,8 @@ class CandidateGraph(nx.Graph):
         if options[on] == self.edges_iter:
             obj = 2
         for elem in options[on](data=True):
-            res.append(function(elem[obj], *args, **kwargs))
+            if hasattr(elem, 'ignore') and getattr(elemen, 'ignore'):
+                res.append(function(elem[obj], *args, **kwargs))
 
         if out:
             out = res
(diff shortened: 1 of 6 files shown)


autocnet/io/db/model.py
```diff
@@ -131,7 +131,7 @@ class Edges(BaseMixin, Base):
     destination = Column(Integer)
     ring = Column(ArrayType())
     fundamental = Column(ArrayType())
-    ignore = Column(Boolean)
+    ignore = Column(Boolean, default=False)
```
jessemapel: For existing databases, if this value is null does it get turned into False by the model?

autocnet/io/db/model.py
```diff
@@ -389,6 +389,8 @@ def measuretype(self, v):
         event.listen(Measures.__table__, 'after_create', valid_point_trigger)
         event.listen(Base.metadata, 'before_create', valid_geom_function)
         event.listen(Images.__table__, 'after_create', valid_geom_trigger)
+        event.listen(Base.metadata, 'before_create', ignore_image_function)
+        event.listen(Images.__table__, 'after_create', ignore_image_trigger)
```
jessemapel: The if here doesn't make sense for all of the triggers. The first works on points, the second on images, and the third on images and measures

autocnet/io/db/model.py
jlaura: Does an edge ever need to be ignored or are we safe to just ignore nodes?

autocnet/io/db/model.py
jlaura: I believe that the if was in because without it the triggers attempt to register whenever model.py is imported. This then throws errors about pre-existing triggers. That is my recollection.

autocnet/io/db/model.py
jessemapel: If we ignore a node we likely also want to ignore all of the adjacent edges too. I'm not sure of a great way to do this in the non-network case because the graph doesn't seem accessible by the nodes.

autocnet/graph/network.py
```diff
@@ -585,7 +585,8 @@ def apply_func_to_edges(self, function, nodes=[], *args, **kwargs):
     def apply(self, function, on='edge', out=None, args=(), **kwargs):
         """
         Applys a function to every node or edge, returns collected return
-        values.
+        values. If applying to functions to nodes, then all ignored nodes
+        will be skipped.
```
jlaura: a function?

autocnet/graph/network.py
```diff
@@ -626,6 +627,8 @@ def apply(self, function, on='edge', out=None, args=(), **kwargs):
         if options[on] == self.edges_iter:
             obj = 2
         for elem in options[on](data=True):
+            if hasattr(elem[obj], 'ignore') and getattr(elem[obj], 'ignore'):
```
jlaura: what about just `getattr(elem[obj], 'ignore', False)`? or to not break an old project where 'ignore' doesn't exist yet, default to True?

autocnet/graph/node.py
```diff
@@ -695,6 +699,26 @@ def measures(self):
...
+        res = self._from_db(Images, key='id')
+        return res.ignore
+
+    @ignore.setter
+    def ignore(self, ignore):
+        """
+        Sets the ignore flag in the Images table
+        """
+        session = Session()
+        res = session.query(Images).filter(getattr(Images,'id') == self['node_id']).first()
```
jlaura: we should probably use `.one()` instead of `.first()`? one will error out if 2+ entries are found and this is a nice additional consistency check to have, maybe?

PR #421, opened 2020-03-11
:construction:  Algorithm in, working on tests 

jlaura:
Center of gravity based subpixel matcher. Working on tests now.

### DIFF
--- a/autocnet/matcher/naive_template.py
+++ b/autocnet/matcher/naive_template.py
@@ -5,9 +5,45 @@ from scipy.ndimage.interpolation import zoom
 
 
 def pattern_match_autoreg(template, image, subpixel_size=3, max_scaler=0.2, func=cv2.TM_CCORR_NORMED):
+    """
+    Call an arbitrary pattern matcher using a subpixel approach where a center of gravity using
+    the correlation coefficients are used for subpixel alignment.
+
+    Parameters
+    ----------
+    template : ndarray
+               The input search template used to 'query' the destination
+               image
+
+    image : ndarray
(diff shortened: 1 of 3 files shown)


jlaura:
Failing as expected while I work on the tests. Wanted to get the logic up because I would like get the warping stuff @Kelvinrr and this into the code base ASAP, but cleanly,  since they both look quite promising.

autocnet/matcher/naive_template.py
```diff
@@ -1,7 +1,54 @@
+from math import floor
 import cv2
+import numpy as np
 from scipy.ndimage.interpolation import zoom
 
 
+def pattern_match_autoreg(template, image, subpixel_size=3, max_scaler=0.2, func=cv2.TM_CCORR_NORMED):
```
jessemapel: Needs a doc string

autocnet/matcher/naive_template.py
```diff
@@ -1,7 +1,54 @@
...
+    lower = upper - 1
+    # x, y are the location of the upper left hand corner of the template in the image
+    area = result[y-lower:y+upper,
+                  x-lower:x+upper]
+
+    if area.shape != (subpixel_size+2, subpixel_size+2):
+        print("Max correlation is too close to the boundary.")
+        return None, None, 0
+        
+    # Find the max on the edges, scale just like autoreg (but why?)
```
jessemapel: I'd imagine this is trying to eliminate the effects of the window clipping the peak. I am not 100% sure why you would want to threshold by less than the max on the edge though.

PR #422, opened 2020-03-11
Various bug fixes

jessemapel:
This fixes a few things.

 * A bug where NetworkNodes tried to set the ignore column in the images table before they were added to the table.
 * A parameter got re-named but the serial point placing code didn't get updated
 * If the new "most interesting point" code doesn't find a feature it causes the entire run to fail

### DIFF
--- a/autocnet/spatial/overlap.py
+++ b/autocnet/spatial/overlap.py
@@ -218,24 +218,32 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
             image_coord = csmapi.ImageCoord(newline, newsample)
             pcoord = node.camera.imageToGround(image_coord)
             # Get the BCEF coordinate from the lon, lat
-            lon, lat, _ = reproject([pcoord.x, pcoord.y, pcoord.z], semi_major, semi_minor,
-                            'geocent', 'latlon')
+            updated_lon, updated_lat, _ = reproject([pcoord.x, pcoord.y, pcoord.z],
+                                                    semi_major, semi_minor, 'geocent', 'latlon')
 
             # Get the new DEM height
             if dem is None:
-                height = 0
+                updated_height = 0
             else:
-                px, py = dem.latlon_to_pixel(lat, lon)
-                height = dem.read_array(1, [px, py, 1, 1])[0][0]
+                px, py = dem.latlon_to_pixel(updated_lat, updated_lon)
+                updated_height = dem.read_array(1, [px, py, 1, 1])[0][0]
 
 
             # Get the BCEF coordinate from the lon, lat
-            x, y, z = reproject([lon, lat, height], semi_major, semi_minor,
-                                'latlon', 'geocent')
+            x, y, z = reproject([updated_lon, updated_lat, updated_height],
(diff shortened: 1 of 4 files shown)


jessemapel:
I've got some other changes that I'm also testing out, so don't merge this yet.

ladoramkershner:
Two of these are the same bugs I found while working with the code. I found one more, would you like to put it in this PR to get it all out of the way and not have to deal with merge conflicts down the road?

jessemapel:
@ladoramkershner That sounds good to me

autocnet/spatial/overlap.py
```diff
@@ -204,7 +208,7 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
 
         # Get the updated lat/lon from the feature in the node
         if cam_type == "isis":
-            p = isis.point_info(node["image_path"], newsample, newline, pointtype="image")
+            p = isis.point_info(node["image_path"], newsample, newline, point_type="image")
             x, y, z = p["GroundPoint"]["BodyFixedCoordinate"].value
```
ladoramkershner: x, y, and z are returned in km and everything else is working in meters. Easy Fix, could do here?

jessemapel:
@ladoramkershner I'm going to add a test case for some of these, so don't merge just yet.

bin/acn_overlaps
```diff
@@ -39,7 +39,7 @@ def main(msg, config):
 
     print('Placing points in overlap', oid)
     points = place_points_in_overlap(nodes, geom, msg["cam_type"],
-                                     msg['distribute_points_kwargs'])
+                                     distribute_points_kwargs=msg['distribute_points_kwargs'])
```
jlaura: Is this a weird new line thing going on or is this just git acting weird?

ladoramkershner: Additionally, could you pass cam_type in specifically?

bin/acn_overlaps
```diff
@@ -39,7 +39,7 @@ def main(msg, config):
 
     print('Placing points in overlap', oid)
```
ladoramkershner: I noticed that having this print statement higher is helpful if something goes wrong in the node generation loop

ladoramkershner:
I actually have one more edit that could get added to this PR if you want, there are no merge conflicts, so I could do this later.

config_parser.py is in the root directory (autocnet), not the root of the code (autocnet/autocnet), which means the setup.py call does not copy it to the relevant area. After that is moved the import of config_parser in __init__.py (line 29) is changed to `from autocnet.config_parser import parse_config`

jessemapel:
@ladoramkershner Wanna post that as issue in case we don't get to it? That sounds like it's been around for a while.

bin/acn_overlaps
ladoramkershner: It is just git acting weird. I viewed the file (using the options on this file preview) and it looks totally fine there

ladoramkershner:
Tests are passing, everything looks good, and I will do a separate issue and PR to move the config_parser. So I am going to merge!

PR #427, opened 2020-03-13
Updates to get docs rendering across the library.

jlaura:


jlaura:
The first two files are .py. Changes are whitespace and to remove unused imports that were causing doc build failures. Everything else are .rst files that map the docs.

PR #431, opened 2020-03-14
Fixes #426

### LINKED ISSUE #426: Interesting feature search can move poitn off image
In spatial/overlap.py -> place_points_in_overlap()

The following piece of code
```
image, _, _ = clip_roi(node.geodata, sample, line, size_x=size, size_y=size)
interesting = extract_most_interesting(image)
newsample = sample + (interesting.x - size)
newline = line + (interesting.y - size)
```

can result in negative newline or newsample. I am not sure if this is because the orb detector is finding its most interesting feature off of the image, or if the conversion from interesting to newline/newsample is flawed. However, this process is all operating on a single image, so it seems like this shouldn't even be possible if the code were doing what we think it is doing.

jlaura:
See #426 for the related issue.

- Test updates to the subpixel_template, subpixel_phase, iterative_phase, and place_points_in_overlaps methods that make me much more comfortable that the code is behaving how we believe it is.
- Added an ROI object to (1) store the offsets that we need in subpixel shifting and overlap computation, (2) to make the returns cleaners, (3) to provide a place to implement warping, and (4) to, I think, make this more testable.

### DIFF
--- a/autocnet/matcher/subpixel.py
+++ b/autocnet/matcher/subpixel.py
@@ -179,7 +179,7 @@ def subpixel_phase(sx, sy, dx, dy,
 
         s_size = s_image.shape
         d_size = d_template.shape
-        updated_size_x = int(min(s_size[1], d_size[1])) # Why is this /2?
+        updated_size_x = int(min(s_size[1], d_size[1]))
         updated_size_y = int(min(s_size[0], d_size[0]))
         
         # Have to subtract 1 from even entries or else the round up that
@@ -326,13 +326,13 @@ def subpixel_ciratefi(sx, sy, dx, dy, s_img, d_img, search_size=251, template_si
                                 size_x=search_size, size_y=search_size)
     template = t_roi.clip()
     search = s_roi.clip()
-    
+
     if template is None or search is None:
         return None, None, None
 
     x_offset, y_offset, strength = ciratefi.ciratefi(template, search, **kwargs)
-    dx += (x_offset + dxr)
-    dy += (y_offset + dyr)
+    dx += (x_offset + t_roi.axr)
+    dy += (y_offset + t_roi.ayr)
     return dx, dy, strength
 
 def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=(251, 251), reduction=11, convergence_threshold=1.0, max_dist=50, **kwargs):
--- a/autocnet/matcher/subpixel.py
+++ b/autocnet/matcher/subpixel.py
@@ -11,6 +11,7 @@ from autocnet.matcher.naive_template import pattern_match, pattern_match_autoreg
 from autocnet.matcher import ciratefi
 from autocnet.io.db.model import Measures, Points, Images, JsonEncoder
 from autocnet.graph.node import NetworkNode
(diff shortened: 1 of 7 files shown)


jlaura:
@Kelvinrr I made changes to the API to use the ROI object everywhere except for the cross_instrument_matcher. Would you mind reviewing this PR and, if it looks good I'll also update the CIM?

Since the CIM isn't tested I don't feel comfortable refactoring anything in there!

jessemapel:
@jlaura Is the plan to merge this and then let Kelvin put in a separate PR with the CIM?

Kelvinrr:
@jlaura I would say probably wait on refactoring CIM since I am currently making changes in there. I can probably just add that to my TODO list before making a PR for my current changes if this get merged first.

jlaura:
This - assuming it looks good, followed by integration into CIM in another PR I think makes sense.

jlaura:
Pls - checkout whether I caught even most of the cases in the tests for ROI extraction, subpixel shifts, and overlap placement. The latter two use ROI so are impacted.

autocnet/matcher/subpixel.py
```diff
@@ -160,10 +167,48 @@ def subpixel_phase(template, search, **kwargs):
...
+    d_roi = roi.Roi(d_img, dx, dy, size_x=image_size[0], size_y=image_size[1])
+
+    s_image = s_roi.clip()
+    d_template = d_roi.clip()
+
+    if s_image.shape != d_template.shape:
+
+        s_size = s_image.shape
+        d_size = d_template.shape
+        updated_size_x = int(min(s_size[1], d_size[1])) # Why is this /2?
```
Kelvinrr: What is the significance of the comment?

jlaura: that should be popped. I'll update.

autocnet/transformation/roi.py
```diff
@@ -0,0 +1,153 @@
...
+
+    @bottom_y.setter
+    def bottom_y(self, y):
+        self._bottom_y = int(y)
+
+    @property
+    def origin_roi(self):
+        _ = self.image_extent
+        return ((self.right_x - self.left_x) / 2,
+                (self.bottom_y - self.top_y) / 2)
```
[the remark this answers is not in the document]
jlaura: I think I  can/should completely remove this method. I ended up not needing it for the overlap stuff.

autocnet/transformation/roi.py
```diff
@@ -0,0 +1,153 @@
...
+            self.left_x = 0
+        if self._y - self.size_y < 0:
+            self.top_y = 0
+        if self._x + self.size_x > raster_size[0]:
+            self.right_x = raster_size[0]
+        if self._y + self.size_y > raster_size[1]:
+            self.bottom_y = raster_size[1]
+
+        return [self.left_x, self.right_x,
+                self.top_y, self.bottom_y]
```
jlaura: Same here as the comment above from @Kelvinrr. I struggled here with whether to return and not set or to set. I am thinking just return, because since this is now stateful, a caller could update the size_x and size_y and then the self.*_x/y in here will be out of sync.

PR #433, opened 2020-03-18
minor bug fixes and weight value changes

ladoramkershner:
Changes weights so that they are null before subpixel registration and after subpixel registration, reference measures have a weight of one while non-reference measures have a weight=cost func.

Additionally there were a few minor bugs I found while running the code. If both phase and template matching pass, the new adjusted line sample should be the phase results (since my last PR #432).


ladoramkershner:
I am not sure if I should add tests for confirming the weight values

ladoramkershner:
Closing in favor of #434

PR #435, opened 2020-03-18
Adds geom_match to point registration 

Kelvinrr:
Tested on a few points manually with python plots, results matched Themis matches. Ran on the cluster but can't verify since qnet is not loading over the VPN.  

I still need to incorporate Jay's clip_roi changes.


autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -265,101 +93,58 @@ def generate_ground_points(ground_db_config, nspts_func=lambda x: int(round(x,1)
...
-                adjusted = True
-                break
-            except Exception as e:
-                continue
-        if not adjusted:
-            raise("This is some garbage")
+        linessamples = isis.point_info(ground_mosaic.file_name, p.x, p.y, 'ground')
+        sample = linessamples.get('Sample')
+        line = linessamples.get('Line')
+        size = 200
```
jlaura: Has this just worked at this size? No reason to parametrize?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -265,101 +93,58 @@ def generate_ground_points(ground_db_config, nspts_func=lambda x: int(round(x,1)
...
+        image = roi.Roi(ground_mosaic, sample, line, size_x=size, size_y=size, dtype="uint64").clip()
+        interesting = extract_most_interesting(bytescale(image),  extractor_parameters={'nfeatures':30})
```
jlaura: Same question with parametrizing the extractor on ORB.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -265,101 +93,58 @@ def generate_ground_points(ground_db_config, nspts_func=lambda x: int(round(x,1)
...
 
-        for k, record in res.iterrows():
-            record["pointid"] = i
-            records.append(record)
-            coord_list.append(p)
+        # kps are in the image space with upper left origin, so convert to
+        # center origin and then convert back into full image space
+        newsample = sample + (interesting.x - size)
+        newline = line + (interesting.y - size)
```
jlaura: With the way clip_roi was refactored, I think this should be:
```
        left_x, _, top_y, _ = image.image_extent
        newsample = left_x + interesting.x
        newline = top_y + interesting.y
```

because the image might be smaller than what this caller expects if we are near an image boundary. Please check this logic though!

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -400,22 +186,27 @@ def propagate_point(lon, lat, pointid, paths, lines, samples, resolutions, verbo
...
-          print("Winning CORR: ", match_results[3], "Themis Pixel shift: ", match_results[4])
-          print("Themis Image: ", match_results[6], "CTX image:", match_results[7])
+          print("Full results: ", best_results)
+          print("Winning CORR: ", best_results[3], "Themis Pixel shift: ", best_results[4])
+          print("Themis Image: ", best_results[6], "CTX image:", best_results[7])
           print("Themis S,L: ", f"{sx},{sy}", "CTX S,L: ", f"{sample},{line}")
 
         # hardcoded for now
-        if best_results[3] < 0.7:
+        if best_results[3] == None or best_results[3] < 0.7:
```
jlaura: For the 2 questions above, adding the `#hardcoded for now` comment is totally cool.

autocnet/matcher/subpixel.py
```diff
@@ -407,9 +417,132 @@ def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=(251, 251), reduction=11,
...
+    # read_array not getting correct type by default
+    isis2np_types = {
+            "UnsignedByte" : "uint8",
+            "SignedWord" : "int16",
+            "Real" : "float64"
+    }
+
+    base_pixels = list(map(int, [match_points[0][0], match_points[0][1], size_x*2, size_y*2]))
+    base_type = isis2np_types[pvl.load(base_cube.file_name)["IsisCube"]["Core"]["Pixels"]["Type"]]
+    base_arr = base_cube.read_array(pixels=base_pixels, dtype=base_type)
```
jlaura: so even passing the dtype arg to read_array, you aren't getting back what you expect? This is a plio issue, good to fix here in the short term, but we should get an issue open on plio.

Kelvinrr: Explicit dtype pass works fine, but I believe Plio has logic to determine the correct dtype if not explicitly mentioned which results in the wrong type (e.g. uint image being read as float) and it messes everything downstream. I'll open an issue.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: Old copy-pasted code from original "find interesting" implementation, should try to somehow incorporate code but putting if putting in comments works, for now, I can do this in separate PR to get this in the code base first.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: I definitely should incorporate those changes. Not sure if in this PR or not?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -400,22 +186,27 @@ def propagate_point(lon, lat, pointid, paths, lines, samples, resolutions, verbo
...
             continue
 
-        pointpvl = isis.point_info(paths[0], x=lon, y=lat, point_type="ground")
+        if dem is None:
+            height = 0
+        else:
+            px, py = dem.latlon_to_pixel(lat, lon)
+            height = dem.read_array(1, [px, py, 1, 1])[0][0]
```
jessemapel: Is this height in meters or kilometers? It needs to be in the same units as the pyproj transoformation or else it your final x, y, z will be off.

autocnet/matcher/subpixel.py
```diff
@@ -407,9 +417,132 @@ def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=(251, 251), reduction=11,
...
+        raise Exception(f"Window: {base_stopx} > {match_size[0]}, center: {bcenter_x},{bcenter_y}")
+    if base_startx < 0:
+        raise Exception(f"Window: {base_startx} < 0, center: {bcenter_x},{bcenter_y}")
+    if base_stopy > match_size[1]:
+        raise Exception(f"Window: {base_stopy} > {match_size[1]}, center: {bcenter_x},{bcenter_y} ")
+    if base_starty < 0:
+        raise Exception(f"Window: {base_starty} < 0, center: {bcenter_x},{bcenter_y}")
+
+    mlat, mlon = spatial.isis.image_to_ground(base_cube.file_name, bcenter_x, bcenter_y)
+    center_x, center_y = spatial.isis.ground_to_image(base_cube.file_name, mlon, mlat)
```
jessemapel: Why is this here? I would expect center_x, center_y and bcenter_x, bcenter_y to be nearly identical.

autocnet/matcher/subpixel.py
```diff
@@ -407,9 +417,132 @@ def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=(251, 251), reduction=11,
...
+                    (base_startx,base_stopy),
+                    (base_stopx,base_stopy),
+                    (base_stopx,base_starty)]
+
+    cube_points = []
+    for x,y in match_points:
+        lat, lon = spatial.isis.image_to_ground(base_cube.file_name, x, y)
+        cube_points.append(spatial.isis.ground_to_image(input_cube.file_name, lon, lat)[::-1])
+    for x,y in cube_points:
+        if x < 0 or y < 0:
```
jessemapel: You should also check `if x > image_size[0] or y > image_size[1]`

autocnet/matcher/subpixel.py
```diff
@@ -516,29 +648,15 @@ def subpixel_register_point(pointid, iterative_phase_kwargs={}, subpixel_templat
...
 
-        new_phase_x, new_phase_y, phase_metrics = iterative_phase(source.sample,
-                                                                source.line,
-                                                                new_template_x,
-                                                                new_template_y,
-                                                                source_node.geodata,
-                                                                destination_node.geodata,
-                                                                **iterative_phase_kwargs)
-        if new_phase_x == None:
+        if new_x == None:
```
jessemapel: probably should check new_x and new_y just to be sure

autocnet/spatial/overlap.py
```diff
@@ -213,8 +213,12 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
...
         if cam_type == "isis":
             p = isis.point_info(node["image_path"], newsample, newline, point_type="image")
-            x, y, z = p["GroundPoint"]["BodyFixedCoordinate"].value
-            if p["GroundPoint"]["BodyFixedCoordinate"].units.lower() == "km":
+            try:
+                x, y, z = p["BodyFixedCoordinate"].value
+            except:
+                x,y,x = ["BodyFixedCoordinate"]
+
+            if p["BodyFixedCoordinate"].units.lower() == "km":
```
jessemapel: This will error if the keyword doesn't have units for some reason. I saw this come up a few times but couldn't reproduce. I would use getattr with an empty string default.

autocnet/matcher/subpixel.py
Kelvinrr: these centers are in full image space and represent the location of the feature trying to be matched. Following ISIS's code, the source feature is reprojected into the destination image rather than having their locations be passed in.

autocnet/spatial/overlap.py
Kelvinrr: I've been meaning to normalize this in point_info since it is obnoxious having everyone downstream check if it has units attr or is implied. I'll put in the check here for now and fix this in another PR.

autocnet/matcher/cross_instrument_matcher.py
Kelvinrr: DEM is in meters, not sure exactly what pyproj wants them in but from just visualizing the results, meters seems correct or else presumably the x,y,z would be much more off than subpixel.

autocnet/matcher/subpixel.py
jessemapel: This is re-projecting into the same image though. I'd expect:

center_x, center_y = spatial.isis.ground_to_image(**input_cube**.file_name, mlon, mlat)

for that

Kelvinrr: Oof, that is a bug from when I changed params. This is causing distances to be wrong.

autocnet/matcher/subpixel.py
jlaura: :+1:

autocnet/matcher/cross_instrument_matcher.py
jlaura: :+1:

jlaura:
This looks good to merge here. Any objections?

autocnet/spatial/overlap.py
```diff
@@ -213,8 +213,12 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
...
         if cam_type == "isis":
             p = isis.point_info(node["image_path"], newsample, newline, point_type="image")
-            x, y, z = p["GroundPoint"]["BodyFixedCoordinate"].value
-            if p["GroundPoint"]["BodyFixedCoordinate"].units.lower() == "km":
+            try:
+                x, y, z = p["BodyFixedCoordinate"].value
+            except:
+                x,y,x = ["BodyFixedCoordinate"]
+
+            if gitattr(p["BodyFixedCoordinate"], "units", "None").lower() == "km":
```
jessemapel: Typo, should be getattr

PR #436, opened 2020-03-18
Adds additional debug prints to point registering

jlaura:
@ladoramkershner Can you pull this locally and give it a test? I am struggling to reproduce, but this should make it so that the code throws and we can then grep into the logs to see what is going on instead of 🍀 searching.

### DIFF
--- a/autocnet/matcher/subpixel.py
+++ b/autocnet/matcher/subpixel.py
@@ -385,10 +385,10 @@ def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=(251, 251), reduction=11,
     dline = dy
 
     while True:
-        try:
-            shifted_dx, shifted_dy, metrics = subpixel_phase(sx, sy, dx, dy, s_img, d_img, image_size=size, **kwargs)
-        except:
-            return None, None, None
+        #try:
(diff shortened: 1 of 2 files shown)


autocnet/matcher/subpixel.py
```diff
@@ -385,10 +385,10 @@ def iterative_phase(sx, sy, dx, dy, s_img, d_img, size=(251, 251), reduction=11,
...
 
     while True:
-        try:
-            shifted_dx, shifted_dy, metrics = subpixel_phase(sx, sy, dx, dy, s_img, d_img, image_size=size, **kwargs)
-        except:
-            return None, None, None
+        #try:
+        shifted_dx, shifted_dy, metrics = subpixel_phase(sx, sy, dx, dy, s_img, d_img, image_size=size, **kwargs)
+        #except:
+        #    return None, None, None
```
jlaura: I think it is safe now to remove this deep try/except since subpixel_phase will return cleanly. I didn't fully pull because I want to test this on a few thousand points.

jlaura:
@Kelvinrr I pulled the asserts completely.

autocnet/matcher/subpixel.py
ladoramkershner: This change is already in dev

PR #437, opened 2020-03-20
Fixed a bug where planetographic lat/lon values were being passed to campt

jessemapel:
This was causing measures to be incorrectly reprojected from ground points.

### DIFF
--- a/autocnet/spatial/overlap.py
+++ b/autocnet/spatial/overlap.py
@@ -173,8 +173,6 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
         lon = v[0]
         lat = v[1]
 
-        print(f'Placing point at lat: {lat}, lon: {lon}')
-
         # Calculate the height, the distance (in meters) above or
         # below the aeroid (meters above or below the BCBF spheroid).
         if dem is None:
@@ -198,7 +196,6 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
             gnd = csmapi.EcefCoord(x, y, z)
             image_coord = node.camera.groundToImage(gnd)
             sample, line = image_coord.samp, image_coord.line
-        print(f'Reference image coordinate is line: {line}, sample: {sample}')
 
         # Extract ORB features in a sub-image around the desired point
         image_roi = roi.Roi(node.geodata, sample, line, size_x=size, size_y=size)
@@ -215,7 +212,6 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
         left_x, _, top_y, _ = image_roi.image_extent
         newsample = left_x + interesting.x
         newline = top_y + interesting.y
-        print(f'Interesting image coordinate is line: {newline}, sample: {newsample}')
 
         # Get the updated lat/lon from the feature in the node
         if cam_type == "isis":
@@ -252,7 +248,6 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
         # original point and hope the matcher can handle it when sub-pixel registering
         updated_lon, updated_lat, updated_height = reproject([x, y, z], semi_major, semi_minor,
                                                              'geocent', 'latlon')
-        print(f'Updated point at lat: {updated_lat}, lon: {updated_lon}')
         if not geom.contains(shapely.geometry.Point(updated_lon, updated_lat)):
             x, y, z = reproject([lon, lat, height],
                                 semi_major, semi_major, 'latlon', 'geocent')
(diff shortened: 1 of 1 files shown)


autocnet/spatial/overlap.py
```diff
@@ -239,29 +242,34 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
...
         # If the updated point is outside of the overlap, then revert back to the
         # original point and hope the matcher can handle it when sub-pixel registering
         updated_lon, updated_lat, updated_height = reproject([x, y, z], semi_major, semi_minor,
                                                              'geocent', 'latlon')
         if not geom.contains(shapely.geometry.Point(updated_lon, updated_lat)):
             x, y, z = reproject([lon, lat, height],
-                                semi_major, semi_minor, 'latlon', 'geocent')
+                                semi_major, semi_major, 'latlon', 'geocent')
+            updated_lon, updated_lat, updated_height = reproject([x, y, z], semi_major, semi_minor,
+                                                                 'geocent', 'latlon')
```
jessemapel: This was another bug I found while working on this. If it falls back to the original point, then it didn't re-update the lat/lon that gets used for the point.

autocnet/spatial/overlap.py
```diff
@@ -183,13 +183,16 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
 
         # Need to get the first node and then convert from lat/lon to image space
         node = nodes[0]
+        x, y, z = reproject([lon, lat, height],
+                             semi_major, semi_minor,
+                             'latlon', 'geocent')
         if cam_type == "isis":
-            line, sample = isis.ground_to_image(node["image_path"], lon ,lat)
+            # Convert to geocentric lon, lat
+            geocent_lon, geocent_lat, _ = reproject([x, y, z],
```
ladoramkershner: I am a confused because the footprints on our images (and therefore our geoms) were generated with ISIS, weren't they? Then distribute_points_in_geom is just grabbing the extents and placing points using a linspace call. So the lat type of 'lat' and 'lon' variables were driven by fooprintinit. So shouldn't it already be in the form campt wants?

jessemapel: I don't think campt and footprintinit use the same system :(

autocnet/spatial/overlap.py
```diff
@@ -239,29 +242,34 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
...
-                                semi_major, semi_minor, 'latlon', 'geocent')
+                                semi_major, semi_major, 'latlon', 'geocent')
+            updated_lon, updated_lat, updated_height = reproject([x, y, z], semi_major, semi_minor,
+                                                                 'geocent', 'latlon')
 
         point_geom = shapely.geometry.Point(x, y, z)
         point = Points(apriori=point_geom,
                        adjusted=point_geom,
                        pointtype=2, # Would be 3 or 4 for ground
                        cam_type=cam_type)
```
ladoramkershner: What latitude type is up in the database now? 

If we had to convert the geom into planetocentric earlier, shouldn't we convert back to planetographic when creating the point?

jessemapel: The geometry passed to the database is actually the X, Y, Z rectangular geometry and the model automatically converts it to the geodetic lat, lon geometry.

autocnet/spatial/overlap.py
ladoramkershner: .........................................

autocnet/spatial/overlap.py
thareUSGS: OC/OG rears it ugly head. For initial testing, perhaps set Mars as a sphere (then they are the same). Then change back to ellipse to finalize oc/og testing.

autocnet/spatial/overlap.py
jessemapel: campt was originally just for sensor model testing and doesn't exactly work the same as the other applications. I'll dig into the code a little and verify that they use different systems.

autocnet/spatial/overlap.py
ladoramkershner: Riiiighhhttt. 🤦‍♀

autocnet/spatial/overlap.py
ladoramkershner: I am nervous, because I don't know what lat type this is coming in as..
We have two places in this function were we could be messing this up; here and the reproject call where we go from x, y, z to lat lon. Who is to say we are messing up both? What if it was just the second projection to lat lon?

jessemapel: You can check if things are being converted correctly by looking at where the points get laid out in the overlaps. I also did a little testing by calling campt on the edges of the image and comparing it to the values in the footprint polygons

autocnet/spatial/overlap.py
jessemapel: @thareUSGS I don't know if we'll have time to do that as it would require generating our entire mars ctx database x.x

autocnet/spatial/overlap.py
ladoramkershner: By doing that, I see that the geometry coming in is already geocentric, but it doesn't hurt to run it through that conversion just in case. Wow I took a long walk on that one, thanks for the ride along... 🤦‍♀

ladoramkershner:
I think this looks good!

PR #438, opened 2020-04-02
Initial work to config after instantiation

jlaura:
Work in progress right now. 

~~Currently at the first step of being able to instantiate the NCG and then configure it. Not that any of the stuff calling works, but the object can be instantiated.~~

Refactor to have the config managed in the NCG is completed. This has been tested up to placing points in overlaps. Tomorrow I'll test the rest of the pipeline. Going to put comments inline to help with reviews.


autocnet/io/db/connection.py
```diff
@@ -34,6 +35,23 @@ def new_connection(dbconfig):
                                                   dbconfig['pgbouncer_port'],
                                                   dbconfig['name'])
     engine = sqlalchemy.create_engine(db_uri,
-                                      poolclass=sqlalchemy.pool.NullPool)
-    Session = orm.sessionmaker(bind=engine, autocommit=True)
+                poolclass=sqlalchemy.pool.NullPool,
+                connect_args={f"application_name":"AutoCNet_{socket.gethostname()}"},
+                isolation_level="AUTOCOMMIT")
+    Session = orm.sessionmaker(bind=engine, autocommit=False)
```
jlaura, line 41: autocommit to False here. I think we were getting lucky with it being True and were then keeping orphaned sessions around. When set to True now, I get (as expected) Transaction errors out of sqlalchemy.

autocnet/io/db/model.py
```diff
@@ -271,6 +269,11 @@ class PointType(enum.IntEnum):
 
 class Points(BaseMixin, Base):
     __tablename__ = 'points'
+    latitudinal_srid = -1
```
jlaura, line 272: Comment here applies to all of the -1 defaults above. Each model can need to have default values for the srids and radii. Previously these were loaded from the config file, but the config is no longer loaded on module import. Therefore, we need some defaults. Below, these defaults will then be set by the values in the config. Pros: This allows for a decoupling of the config and the models. Con: This means I could modify the srid/radii in the config and this will not be reflected in the DB.

autocnet/io/db/model.py
```diff
@@ -391,12 +397,25 @@ def measuretype(self, v):
         event.listen(Images.__table__, 'after_create', valid_geom_trigger)
         event.listen(Base.metadata, 'before_create', ignore_image_function)
         event.listen(Images.__table__, 'after_create', ignore_image_trigger)
-
+    
     Base.metadata.bind = engine
+    
+    # Set the class attributes for the SRIDs
+    spatial = config['spatial']
```
jlaura, line 404: This is where the default srid/radii are being properly set by the config. When the NCG is instantiated, it calls the `try_create_database` method. Even if the tables exist, the class attributes are set for the module (and any importing modules).

The con here is that you can't mix and match srids and radii between NCGs running on the same kernel. I would argue this is a seatbelt pro in disguise because you shouldn't be mixing those when we don't transform between them.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -144,10 +142,20 @@ def generate_ground_points(ground_mosaic, nspts_func=lambda x: int(round(x,1)*1)
...
+                    lat, 
+                    pointid, 
+                    paths, 
+                    lines, 
+                    samples, 
+                    verbose=False):
     """
 
     """
+    engine = Session.get_bind()
```
jlaura, line 158: the session carries the engine, so stop passing the session. Learned this today.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -201,11 +209,8 @@ def propagate_point(lon, lat, pointid, paths, lines, samples, verbose=False):
         if best_results[3] == None or best_results[3] < 0.7:
             continue
 
-        if dem is None:
-            height = 0
-        else:
-            px, py = dem.latlon_to_pixel(lat, lon)
-            height = dem.read_array(1, [px, py, 1, 1])[0][0]
+        px, py = dem.latlon_to_pixel(lat, lon)
```
jlaura, line 212: The dem has to exist, so don't pretend that it doesn't.

autocnet/spatial/overlap.py
```diff
@@ -55,81 +60,33 @@ def place_points_in_overlaps(nodes, size_threshold=0.0007,
...
-
-def place_points_in_overlap(nodes, geom, cam_type="csm",
+    
+    for overlap in Overlay.overlapping_larger_than(size_threshold, Session):
+        if overlap.intersections == None:
+            continue
+        place_points_in_overlap(overlap, 
+                                cam_type=cam_type,
+                                distribute_points_kwargs=distribute_points_kwargs,
+                                Session=Session,
```
jlaura, line 77: I need to refactor this further - if the ncg is being passed, not reason to also pass the Session, config, and dem....

autocnet/spatial/overlap.py
```diff
@@ -55,81 +60,33 @@ def place_points_in_overlaps(nodes, size_threshold=0.0007,
...
+                                ncg=ncg)
+    
+def place_points_in_overlap(overlap,  
+                            cam_type="csm",
                             size=71,
-                            distribute_points_kwargs={}):
+                            distribute_points_kwargs={},
+                            Session=None,
+                            config=None,
+                            dem=None,
```
jlaura, line 88: As above.

autocnet/utils/utils.py
```diff
@@ -563,4 +564,31 @@ def bytescale(data, cmin=None, cmax=None, high=255, low=0):
 
     scale = float(high - low) / cscale
     bytedata = (data - cmin) * scale + low
-    return (bytedata.clip(low, high) + 0.5).astype(np.uint8)
\ No newline at end of file
+    return (bytedata.clip(low, high) + 0.5).astype(np.uint8)
+
+def import_func(func):
```
jlaura, line 569: This func is pretty self explanatory I think? Supports acn_submit being generic.

bin/acn_submit
```diff
@@ -1,73 +1,106 @@
...
     #
     # All args/kwargs are passed through the RedisQueue, and then right on to the func.
-    func = getattr(obj, msg['func'])
+    if hasattr(obj, msg['func']): 
+        func = getattr(obj, msg['func'])
+    else:
+        func = import_func(msg['func'])
+        # For now, pass all the potential config items through
+        # most funcs will simply discard the unnecessary ones.
+        msg['kwargs']['Session'] = ncg.Session
```
jlaura, line 75: Refactor so that the msg doesn't need the session, dem, or config. Just the ncg.

autocnet/graph/network.py
```diff
@@ -1310,30 +1316,96 @@ class NetworkCandidateGraph(CandidateGraph):
...
             d.parent = self
+            d.popluate_db()
         for s, d, e in self.edges(data='data'):
             e.parent = self
 
-        # Setup the redis queues
-        redis = config.get('redis')
-        if redis:
-            self.processing_queue = redis['processing_queue']
+    def config_from_file(self, filepath):
```
jlaura, line 1330: These two methods configure the NCG from a file/dict. This is the new way to define the project configuration (no more env variables!).

autocnet/graph/network.py
```diff
@@ -1310,30 +1316,96 @@ class NetworkCandidateGraph(CandidateGraph):
...
+
+        # Setup the DEM
+        # I dislike having the DEM on the NCG, but in the short term it
+        # is the best solution I think. I don't want to pass the DEM around
+        # for the sensor calls.
+        self._setup_dem()
+
+    
+
+    def _setup_dem(self):
```
jlaura, line 1373: Two helpers for configuration.

autocnet/graph/network.py
```diff
@@ -1356,11 +1428,68 @@ def _execute_sql(self, sql):
         sql : str
               The SQL string to be passed to the DB engine and executed.
         """
-        conn = engine.connect()
+        conn = self.engine.connect()
         conn.execute(sql)
         conn.close()
+    
+    def _push_obj_message(self, onobj, function, walltime, args, kwargs):
```
jlaura, line 1435: I couldn't come up with a clean way to not differentiate between a node/edge and a DB utterable. These two could have the same name using single_dispatch, but that feels like icing on a suboptimal solution. Definitely open to thoughts on these two message pushing helpers.

autocnet/graph/network.py
```diff
@@ -1356,11 +1428,68 @@ def _execute_sql(self, sql):
...
+                    'args':args, 
+                    'kwargs':kwargs,
+                    'walltime':walltime}
+            msg['config'] = self.config  # Hacky for now, just passs the whole config dict
+            self.redis_queue.rpush(self.processing_queue,
+                                json.dumps(msg, cls=JsonEncoder))
+        session.close()
+        return len(res)
+
+    def apply(self, function, on='edge', args=(), walltime='01:00:00', filters={}, **kwargs):
```
jlaura, line 1492: This is an extension to the current apply that now also supports Overlay, Point, and Measure rows.

autocnet/graph/network.py
```diff
@@ -1394,50 +1525,54 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', **kwargs):
...
+            'm' : Measures,
+            2 : Measures,
+            'points' : Points,
+            'point' : Points,
+            'p' : Points,
+            3 : Points,
+            'overlaps': Overlay,
+            'overlap' : Overlay,
+            'o' :Overlay,
+            4: Overlay
```
jlaura, line 1540: This should be a constant up top I think, so that acn_submit can import it and use the same mapping. It is being double maintained right now.

autocnet/graph/network.py
```diff
@@ -1394,50 +1525,54 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', **kwargs):
...
-                    'param_step':1}
-
-            self.redis_queue.rpush(self.processing_queue, json.dumps(msg, cls=JsonEncoder))
-
-        # SLURM is 1 based, while enumerate is 0 based
-        job_counter += 1
+        if isinstance(onobj, (Node, Edge, NetworkNode, NetworkEdge)):
+            job_counter = self._push_obj_messages(onobj, function, walltime, args, kwargs)
+        elif isinstance(onobj, DeclarativeMeta):
+            job_counter = self._push_row_messages(onobj, on, function, walltime, filters, args, kwargs)
```
jlaura, line 1552: Depending upon what we are iterating over, generate the proper messages (pushed above onto the queue).

autocnet/graph/network.py
```diff
@@ -1394,50 +1525,54 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', **kwargs):
...
+        
+        env = self.config['env']
+        condaenv = env['conda']
+        isisroot = env['ISISROOT']
+        isisdata = env['ISISDATA']
+        
+        isissetup = f'export ISISROOT={isisroot} && export ISIS3DATA={isisdata}'
+        condasetup = f'conda activate {condaenv}'
+        job = f'acn_submit -r={rhost} -p={rport} {processing_queue}'
+        command = f'{condasetup} && {isissetup} && {job}'
```
jlaura, line 1568: So I don't love this. Since we no longer load the config via an env variable, the cluster submitted jobs need some way to know what env to use and potentially what the ISIS setup is. The job uses `#!/bin/bash -l` to carry the login shell, but that is likely not the env that a user wants to use for the project. Therefore, I added an env section to the config file and that gets used to setup the cluster job command. Open to ideas here!

autocnet/graph/network.py
```diff
@@ -1892,19 +2019,237 @@ def queue_length(self):
...
         Clear the processing queue of any left over jobs from a previous cluster
         job cancellation or hanging jobs.
         """
-        conf = config['redis']
-        queue = StrictRedis(host=conf['host'],
-                            port=conf['port'])
-        queue.flushdb()
+        self.redis_queue.flushdb()
+
+    def cluster_subpixel_register_measures(self, iterative_phase_kwargs={'size': 251},
```
jlaura, line 2032: Once tested, I am going to remove all these cluster_* methods because one can just use apply.

autocnet/graph/network.py
```diff
@@ -1892,19 +2019,237 @@ def queue_length(self):
...
+        return job_counter
+
+    def cluster_place_points_in_overlaps(self, size_threshold=0.0007,
+                                     distribute_points_kwargs={},
+                                     walltime='00:10:00',
+                                     chunksize=1000,
+                                     exclude=None,
+                                     cam_type="csm",
+                                     query_string='SELECT overlay.id FROM overlay LEFT JOIN points ON ST_INTERSECTS(overlay.geom, points.geom) WHERE points.id IS NULL AND ST_AREA(overlay.geom) >= {};'):
+        """
```
jlaura, line 2143: I need to maintain this query_string somewhere because it is so convenient to only place points into overlaps that don't already have points. Perhaps this will be the default in overlay.py?

autocnet/graph/network.py
```diff
@@ -1892,19 +2019,237 @@ def queue_length(self):
...
+                     job_name='cross_instrument_matcher',
+                     mem_per_cpu=config['cluster']['processing_memory'],
+                     time=walltime,
+                     partition=config['cluster']['queue'],
+                     output=config['cluster']['cluster_log_dir']+'/autocnet.cim-%j')
+        job_counter = len(groups.items())
+        submitter.submit(array='1-{}'.format(job_counter))
+        return job_counter
+    
+    def subpixel_register_points(self, **kwargs):
```
jlaura, line 2232: This is how I chose to solve the whole not on the NCG, but need a Session or config issue. Thoughts?

autocnet/cg/cg.py
```diff
@@ -290,15 +292,6 @@ def find_side(side):
...
     order = order[side]
-    query = f"""
-    select ST_AsText(geom) from images
-    order by {func}(geom) {order}
-    limit 1 """
-
-    session = Session()
-    geom = session.execute(query).first()
-    geom = wkt.loads(geom[0])
-    session.close()
```
ladoramkershner, line 301: This is intended to be used on the most eastern or western cube in the database. I am not sure how to do that outside of the database.  Finding the most eastern or western cube could be abstracted and placed into the ncg class, but idk if it has any other uses outside of this function, so I am not sure if it is worth it.

autocnet/graph/network.py
ladoramkershner: 👍

autocnet/io/db/connection.py
```diff
@@ -33,7 +34,25 @@ def new_connection(dbconfig):
...
+     """
+     session = Session()
+     try:
+         yield session
+         session.commit()
+     except:
+         session.rollback()
+         raise
+     finally:
+         session.close()
```
ladoramkershner, line 58: What is this doing? It looks like a test of some sort, how would it get used in the code?

autocnet/graph/network.py
```diff
@@ -1310,30 +1316,97 @@ class NetworkCandidateGraph(CandidateGraph):
...
+     """
+     session = self.Session()
+     try:
+         yield session
+         session.commit()
+     except:
+         session.rollback()
+         raise
+     finally:
+         session.close()
```
ladoramkershner, line 1409: Also, exists in io/db/connection.py

autocnet/graph/network.py
```diff
@@ -1356,11 +1429,69 @@ def _execute_sql(self, sql):
...
+                    'kwargs':kwargs,
+                    'walltime':walltime,
+                    'image_path':image_path,
+                    'param_step':1, 
+                    'config':self.config}
+
+            self.redis_queue.rpush(self.processing_queue, json.dumps(msg, cls=JsonEncoder))
+        return job_counter + 1
+    
+    def _push_row_messages(self, query_obj, on, function, walltime, filters, args, kwargs):
```
ladoramkershner, line 1468: Doc String:
"""
Push messages to the redis queue for database objee.g., Measures, Points, Images, etc.
"""

autocnet/graph/network.py
```diff
@@ -1356,11 +1429,69 @@ def _execute_sql(self, sql):
...
+        session = self.Session()
```
ladoramkershner, line 1469: Why not use context manager here?

autocnet/graph/network.py
ladoramkershner: Where are you thinking about putting it and how do you plan on passing it here and to acn_submit()?

ladoramkershner: Also, are the numerical options needed? It seems just as easy to use 'n', 'e', 'm', 'p' as it is to use 0, 1, 2, 3

autocnet/graph/network.py
ladoramkershner: 👍

autocnet/graph/network.py
ladoramkershner: Isn't the ISISROOT set in the notebook before importing anything? And then the env used is the activate environment (for local launch of jupyter) or the active kernel (on a remote launch of jupyter)? Are you trying to get away from this?

autocnet/graph/network.py
```diff
@@ -1699,24 +1826,26 @@ def from_remote_database(cls, source_db_config, path,  query_string='SELECT * FR
...
+        >>> ncg.add_from_remote_database(source_db_config, outpath, query_string=query)
         """
 
         sourceSession, _ = new_connection(source_db_config)
         sourcesession = sourceSession()
 
         sourceimages = sourcesession.execute(query_string).fetchall()
 
-        destinationsession = Session()
+        destinationsession = self.Session()
```
ladoramkershner, line 1848: Why not use context manager?

autocnet/graph/network.py
ladoramkershner: I worry adding a new function to this list will be forgotten. Is there a way to do a thin wrapper around functions that require a Session import in the relevant sub modules?

Also, is there no circular dependencies with this method?

autocnet/graph/network.py
jlaura: :+1: I will remove it from there!

autocnet/graph/network.py
jlaura: :+1:

autocnet/graph/network.py
jlaura: :+1:

autocnet/graph/network.py
jlaura: Not sure where it should go. Maybe just at the head of this file. Then acn_submit can import it. @Kelvinrr What was the thinking for the numeric options along side the textual options?

autocnet/graph/network.py
jlaura: ISISROOT is being set in the notebook, but that is a hack to allow one to run pysis inside the notebook (so inside the active kernel).

The env used by the cluster job is a standard login shell (!#/bin/bash -l) so if you don't have ISISROOT/ISIS3DATA set in your login shell it will not be in your env on the cluster.

So trying to get away from the hacky including of ISIS in the notebook and to get the cluster jobs to use ISIS when it isn't in your login shell.

autocnet/graph/network.py
jlaura: :+1:

autocnet/graph/network.py
jlaura: I'm not sure I follow what you mean. Can you give an example? Doesn't need to be code.

autocnet/io/db/connection.py
jlaura: Good catch - I removed it.

jlaura:
This should be updated for all comments now. I got the find_side stuff pushed back in.

autocnet/graph/network.py
ladoramkershner: Well all of these calls are pretty formulaic, so is there a way to find all the functions that require a Session input in the signature and attach it to the ncg class?

autocnet/config_parser.py
```diff
@@ -1,13 +1,9 @@
 import os
 import yaml
 
-def parse_config(name='autocnet_config'):
-    if name not in os.environ.keys():
-        return {}
-    
-    filepath = os.environ[name]
+def parse_config(filepath):
```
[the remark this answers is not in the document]
ladoramkershner: 👍

autocnet/graph/node.py
```diff
@@ -589,14 +589,11 @@ def keypoints(self):
...
+        
 
-        if res is None:
-            _ = self.keypoint_file
-            res = self._from_db(Keypoints)
-        res.nkeypoints = len(kps)
-        session.commit()
-        session.close()
+        res = self._from_db(Keypoints)
+        with self.parent.session_scope() as session:
```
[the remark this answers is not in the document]
ladoramkershner: I think it is a safe bet. Are there any downsides to this method?

autocnet/graph/node.py
```diff
@@ -618,14 +615,23 @@ def nkeypoints(self):
...
+        """
+        Creates a CSM sensor for the node and serializes the state
+        to the DB,
+
+        Parameters
+        ----------
+        url : str
+              The URI to a service that can create an ISD to instantiate
+              a sensor.
+        """
```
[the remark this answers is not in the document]
ladoramkershner: 👍

autocnet/io/db/connection.py
ladoramkershner: From what I am reading, this should pair nicely with the session scopes that you implemented.

autocnet/io/db/connection.py
ladoramkershner: I still see it here and I see that you have pushed some commits to address my comments

autocnet/utils/utils.py
ladoramkershner: 👍

bin/acn_submit
ladoramkershner: :+1:

autocnet/spatial/overlap.py
```diff
@@ -55,81 +60,33 @@ def place_points_in_overlaps(nodes, size_threshold=0.0007,
...
+            continue
+        place_points_in_overlap(overlap, 
+                                cam_type=cam_type,
+                                distribute_points_kwargs=distribute_points_kwargs,
+                                Session=Session,
+                                config=config,
+                                dem=dem, 
+                                ncg=ncg)
+    
+def place_points_in_overlap(overlap,
```
ladoramkershner, line 76: Should check to see is Session is not None (similar to matcher/subpixel functions)

autocnet/spatial/overlap.py
```diff
@@ -55,81 +60,33 @@ def place_points_in_overlaps(nodes, size_threshold=0.0007,
...
-                 mem_per_cpu=config['cluster']['processing_memory'],
-                 time=walltime,
-                 partition=config['cluster']['queue'],
-                 output=config['cluster']['cluster_log_dir']+'/autocnet.place_points-%j')
-    job_counter = i+1
-    submitter.submit(array='1-{}%24'.format(job_counter), chunksize=chunksize, exclude=exclude)
-    return job_counter
-
-def place_points_in_overlap(nodes, geom, cam_type="csm",
+
```
ladoramkershner, line 67: Should check to see is Session is not None (similar to matcher/subpixel functions)

autocnet/cg/cg.py
```diff
@@ -267,13 +266,16 @@ def nearest(pt, search):
     """
     return np.argmin(np.sum((search - pt)**2, axis=1))
 
-def find_side(side):
+def find_side(geom, side):
```
ladoramkershner: It seems like one of your commits (commit hash f2ff6c75e6dcf5b2fa02dd27dd588e710407f873) got backed out. In it you had this function taking Session as an input. Additionally, you had **kwargs getting passed into all the other functions. Which seems consistent with what you are doing in acn_submit. Something seems a little fishy here.

autocnet/cg/cg.py
ladoramkershner: If I am correct about your previous commit getting backed out, this did not get moved, but the signature changed. Could you confirm?

autocnet/graph/network.py
ladoramkershner: 👍

autocnet/graph/network.py
ladoramkershner: If it is moved to the top of the file, it would need a more descriptive name, but I think that would work. Is it poor coding etiquette?

autocnet/graph/network.py
ladoramkershner: I think this is a decent solution. The graph objects and the database rows are different enough that I think it is good to handle them seperately

autocnet/graph/network.py
ladoramkershner: Still waiting on removal

autocnet/graph/network.py
ladoramkershner: 👍

autocnet/graph/network.py
ladoramkershner: What happens when CSM is up and running, do we just have these options sitting in the config, requiring people to still have ISIS data areas, even if they are using CSM?

autocnet/graph/node.py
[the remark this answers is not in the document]
jlaura: Not that I am seeing. If anything, this is making it so that we have less dangling sessions hanging around.

autocnet/io/db/connection.py
jlaura: :+1:

autocnet/graph/network.py
jlaura: This will need to be refactored to be an on/off switch once CSM is in.

autocnet/graph/network.py
ladoramkershner: Where in overlap.py? place_points_in_overlap currently has no knowledge of overlaps, it is handed a geom and an id, it doesn't even deal with an Overlay object.

autocnet/graph/network.py
jlaura: Yes. I would argue though that it would not be very clean because we have no way to enforce that the func with a session is actually something that the NCG can work with/on. 

I intentionally didn't want to go this route in order to be more explicit with forcing the developer to enumerate them in here. So the opposite direction of being more restrictive.

autocnet/graph/network.py
jlaura: It should be out of there now.

autocnet/spatial/overlap.py
ladoramkershner: This refactor could tie in with your desire to preserve the sql query string that only places points in overlaps that do not already contain points. Although, I don't see a way of making it work within this function without remaking the same query every time we hit this function.

autocnet/spatial/overlap.py
jlaura: Just pushed an update for this.

autocnet/graph/network.py
jlaura: Moved in to be a class attribute. This should be a clean way for acn_submit to access it now and we only need it stored in one place.

autocnet/graph/node.py
```diff
@@ -566,10 +559,17 @@ def _from_db(self, table_obj, key='image_id'):
         """
         if 'node_id' not in self.keys():
             return
-        session = Session()
-        res = session.query(table_obj).filter(getattr(table_obj,key) == self['node_id']).first()
-        session.close()
+        with self.parent.session_scope() as session:
+            res = session.query(table_obj).filter(getattr(table_obj,key) == self['node_id']).first()
         return res
+
```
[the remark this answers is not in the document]
ladoramkershner: 👍

autocnet/graph/network.py
ladoramkershner: Okay, I think this is good for now!

autocnet/graph/network.py
ladoramkershner: I am not sure what would get put into matcher or spatial.overlap that would require information from the ncg (because it needs the session) and wouldn't operate on the graph. Not that I don't believe you but could you give an example?

autocnet/matcher/subpixel.py
```diff
@@ -548,81 +544,102 @@ def geom_match(base_cube, input_cube, bcenter_x, bcenter_y, size_x=60, size_y=60
...
 
     """
     Given some point, subpixel register all of the measures in the point to the
     first measure.
 
     Parameters
     ----------
-    pointid : int
-              The identifier of the point in the DB
+    Session : obj
```
ladoramkershner, line 638: Session isn't passed in, ncg is

autocnet/matcher/subpixel.py
```diff
@@ -640,20 +657,34 @@ def subpixel_register_point(pointid, iterative_phase_kwargs={}, subpixel_templat
...
                 the database.
     """
+    Session = ncg.Session
+    if not Session:
+        raise BrokenPipeError('This func requires a database session from a NetworkCandidateGraph.')
+    
+    if isinstance(pointid, Points):
+        pointid = pointid.id
+    
     session = Session()
```
ladoramkershner, line 667: Scope the session as you did in previous function (subpixel_register_measure)

autocnet/matcher/subpixel.py
```diff
@@ -548,81 +544,102 @@ def geom_match(base_cube, input_cube, bcenter_x, bcenter_y, size_x=60, size_y=60
...
-    source.ignore = False
+    return result
 
-    session.commit()
-    session.close()
 
-
-def subpixel_register_point(pointid, iterative_phase_kwargs={}, subpixel_template_kwargs={},
-                            cost_func=lambda x,y: 1/x**2 * y, threshold=0.005):
+def subpixel_register_point(pointid, iterative_phase_kwargs={},
```
ladoramkershner, line 626: I think this function is going to have a merge conflict with my PR that just went in. This function uses geom_match (which itself contains the subpixel_template and iterative_phase)

autocnet/matcher/subpixel.py
jlaura: Fixed.

autocnet/matcher/subpixel.py
jlaura: Fixed.

autocnet/matcher/subpixel.py
jlaura: I think that this should all be up to date and merged. Is this looking like I reverted when I merged?

autocnet/graph/network.py
jlaura: We could write a func in overlap that takes adjacent slivers and merges those using some algorithm. We would not want/need to associate that with a cluster spawning job.

I also am not comfortable doing this in an automated way - the code to do this is going to be gnarly and I would rather not put that kind of introspection into the library.

Having said all of that, I am also not happy with passing the NCG around. It is not a clean way to solve the problem. I'm just not 100% sure what a better solution is at this time.

autocnet/matcher/subpixel.py
ladoramkershner: It looks good now!

autocnet/matcher/subpixel.py
ladoramkershner: 👍

autocnet/matcher/subpixel.py
ladoramkershner: 👍

autocnet/graph/network.py
jessemapel: This approach adds some extra code, but I think it's simplicity a big benefit. By keeping the function separate, it's easier to pass in a mock session instead of having to mock a NCG for testing. Also, introspection stuff is nice, but it can result in unexpected behavior if devs aren't aware of it. This way, everything is explicit.

autocnet/graph/network.py
Kelvinrr: @jlaura No idea why there were number originally, probably a good reason at the time, or probably my just being dumb when I first wrote this (I think this was super early into working as a contractor?). In any case, class attr seems to make the most sense to me.

autocnet/graph/network.py
```diff
@@ -1308,32 +1314,122 @@ class NetworkCandidateGraph(CandidateGraph):
...
 
+    apply_iterable_options = {
+            'edge' : self.edges,
+            'edges' : self.edges,
+            'e' : self.edges,
+            0 : self.edges,
+            'node' : self.nodes,
+            'nodes' : self.nodes,
+            'n' : self.nodes,
+            1 : self.nodes,
```
ladoramkershner, line 1325: 'self' is not defined, causing tests to fail.

jlaura: Fixed. I moved this into the class because I think it belongs there.  The dict is mutable and a caller working with multiple NCG might be confused if they ever alter this dict on the fly in one instance to see if propagate to the other instance. So I think that this is cleaner.

ladoramkershner:
Alright, everything looks good to me!

PR #439, opened 2020-04-08
Explicit pointid, remove catch all try/excepts, adding isis point projection error handling

ladoramkershner:
Contains a lot of small alterations:
- specifying pointid and outputing to subpixel logs, for easier debugging
- lowering the default size of iterative phase
- removing the catch all try excepts in matcher/subpixel.py
- raising errors in spatial/isis.py when point does not properly project to/from ground (exposing error keyword output by isis campt)

### DIFF
--- a/autocnet/io/db/model.py
+++ b/autocnet/io/db/model.py
@@ -271,9 +271,10 @@ class PointType(enum.IntEnum):
 
 class Points(BaseMixin, Base):
     __tablename__ = 'points'
-    id = Column(Integer, primary_key=True, autoincrement=False)
+    id = Column(Integer, primary_key=True, autoincrement=True)
     _pointtype = Column("pointType", IntEnum(PointType), nullable=False)  # 2, 3, 4 - Could be an enum in the future, map str to int in a decorator
     identifier = Column(String, unique=True)
+    overlapid = Column(Integer, ForeignKey('overlay.id'), nullable=False)
     _geom = Column("geom", Geometry('POINT', srid=latitudinal_srid, dimension=2, spatial_index=True))
     cam_type = Column(String)
     ignore = Column("pointIgnore", Boolean, default=False)
--- a/autocnet/io/db/model.py
+++ b/autocnet/io/db/model.py
@@ -274,7 +274,7 @@ class Points(BaseMixin, Base):
     id = Column(Integer, primary_key=True, autoincrement=True)
     _pointtype = Column("pointType", IntEnum(PointType), nullable=False)  # 2, 3, 4 - Could be an enum in the future, map str to int in a decorator
     identifier = Column(String, unique=True)
-    overlapid = Column(Integer, ForeignKey('overlay.id'), nullable=False)
+    overlapid = Column(Integer, ForeignKey('overlay.id'))
     _geom = Column("geom", Geometry('POINT', srid=latitudinal_srid, dimension=2, spatial_index=True))
     cam_type = Column(String)
     ignore = Column("pointIgnore", Boolean, default=False)
--- a/autocnet/io/db/model.py
+++ b/autocnet/io/db/model.py
@@ -271,7 +271,7 @@ class PointType(enum.IntEnum):
 
 class Points(BaseMixin, Base):
     __tablename__ = 'points'
-    id = Column(Integer, primary_key=True, autoincrement=True)
+    id = Column(Integer, primary_key=True, autoincrement=False)
     _pointtype = Column("pointType", IntEnum(PointType), nullable=False)  # 2, 3, 4 - Could be an enum in the future, map str to int in a decorator
     identifier = Column(String, unique=True)
     _geom = Column("geom", Geometry('POINT', srid=latitudinal_srid, dimension=2, spatial_index=True))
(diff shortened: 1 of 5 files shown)


autocnet/io/db/model.py
```diff
@@ -271,7 +271,7 @@ class PointType(enum.IntEnum):
 
 class Points(BaseMixin, Base):
     __tablename__ = 'points'
-    id = Column(Integer, primary_key=True, autoincrement=True)
+    id = Column(Integer, primary_key=True, autoincrement=False)
```
jlaura, line 274: Why don't you want the id to auto increment?

autocnet/matcher/subpixel.py
```diff
@@ -541,7 +544,7 @@ def geom_match(base_cube, input_cube, bcenter_x, bcenter_y, size_x=60, size_y=60
       pcm = axs[2].imshow(corrmap**2, interpolation=None, cmap="coolwarm")
       plt.show()
 
-    dist = np.linalg.norm([center_x-x, center_y-y])
+    dist = np.linalg.norm([center_x-sample, center_y-line])
```
jlaura, line 547: :+1:

autocnet/spatial/overlap.py
```diff
@@ -127,7 +127,7 @@ def cluster_place_points_in_overlaps(size_threshold=0.0007,
     submitter.submit(array='1-{}%24'.format(job_counter), chunksize=chunksize, exclude=exclude)
     return job_counter
 
-def place_points_in_overlap(nodes, geom, cam_type="csm",
+def place_points_in_overlap(oid, nodes, geom, cam_type="csm",
```
jlaura, line 130: This is going to conflict with the API change I made to centralize the config, but not an issue for this PR to deal with.

autocnet/spatial/overlap.py
```diff
@@ -169,7 +169,10 @@ def place_points_in_overlap(nodes, geom, cam_type="csm",
         warnings.warn('Failed to distribute points in overlap')
         return []
 
-    for v in valid:
+    for ptid, v in enumerate(valid):
+        pointid = int('5'+oid+str(ptid).rjust(3,'0'))
+        print(f'pointid {pointid}')
```
jlaura, line 174: `int('5'+oid+str(ptid).rjust(3,'0'))` Why is the pointid being set this way?

bin/acn_overlaps
```diff
@@ -38,8 +38,11 @@ def main(msg, config):
         res = session.query(Images).filter(Images.id == id).one()
         nodes.append(NetworkNode(node_id=id, image_path=res.path))
     session.close()
-
-    points = place_points_in_overlap(nodes, geom, cam_type=msg["cam_type"],
+    
+    q = "SELECT MAX(id) FROM overlay"
+    oid_max = session.execute(q).first()[0]
+    oid = str(oid).rjust(len(str(oid_max)), '0')
```
jlaura, line 44: Same question as above - I just want to understand the logic behind the pointid auto increment to this form.

autocnet/io/db/model.py
jlaura: See the bigger question below really.

jlaura:
The CI failure is intermittent and I think cleaned up in the PR that I have in. It is a check to make sure we do not have an orphaned DB connection. As far as I am concerned, this PR is passing CI.

autocnet/io/db/model.py
ladoramkershner: If the point id autoincriments there is no way to access pointid until after it is added to the database. This makes it really hard to print out enough information to figure out which point failed during processing.

autocnet/spatial/overlap.py
ladoramkershner: 👍

autocnet/spatial/overlap.py
ladoramkershner: I wanted a quick way to link a point to the overlap it was in, instead of doing spatial queries in the database. In order to have unique values the pointids needed to be the same length with zero padding, otherwise, overlap 27 point 11 (2710) would come out the same as the overlap 270 point 1. The added five at the beginning of the id is so python doesn't throw away the preceding 0 in overlap.

jlaura: Makes sense. Can you add that as a comment for this line?

ladoramkershner: Additionally I could not do left alignments because then point 1 and point 10 (and all other integers and there ten multiple) would look the same. Then if I do overlap id left aligned and poitn id right aligned I would have all the zeros in the center, with an unkown amount belonging to overlap id and an unknown amount belonging to pointid. You could figure it out by looking up the max number of overlaps in the database, but I figured just ignoring the 5 out front is easier.

jlaura: Why not add a column to the Points table that has a foreign key to the overlay table? Then you don't have do to these gymnastics in code, can maintain the integer Primary Key (PK), and can do queries to find what you want to find?

jlaura:
Logic looks good. Tests need updating.

ladoramkershner:
The only tests failing at this point are the test_overlap_points.py tests, which are deleted in your PR. Should I fix them for sake of consistency and not banking on the next PR to fix it?

PR #442, opened 2020-04-13
Adds some examples using the refactored API

jlaura:
This beefs up the README ever so slightly with a micro-tutorial.


README.md
```diff
@@ -27,3 +27,163 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
...
+Line by line, this code first imports the network candidate graph, a collection
+of nodes and edges that represents a potential control network.
+`from autocnet.graph.network import NetworkCandidateGraph`.
+
+Next, a network candidate graph is instantiated. `ncg =
+NetworkCandidateGraph()`.
+
+The network candidate graph (or NCG) is assocaited with a collection of
+PostGreSQL database tables. We have to initiate the database connection via a
+configuration file. The configuration file. An example configuration file is
```
ladoramkershner, line 60: The name is Bond, James Bond ('The configuration file.' could probably be deleted)

README.md
```diff
@@ -27,3 +27,163 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
...
+provided in the config directory. `ncg.config_from_file('config/demo.yml')`
+
+Once configured, the images need to be loaded and the graph of potential
+overlapping images generated. We do this with `ncg.from_database()`.
```
ladoramkershner, line 64: This is assuming that they have a database populated with images, but you do not mention yet how to get images into the database. Maybe you should start with a section on that? It could include the from_filelist(), with the warning that it is slow, talk about the add_from_remote_database option, then go into this example.

README.md
```diff
@@ -27,3 +27,163 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
...
+'pgbouncer_port':5432, 'name':'someothertable'}.
+
+In this example, we want to use a spatial query to subset the data. We could
+also use an attribute query or some combination. The only restriction is that
+the quert string be valid SQL. `geom = 'LINESTRING(145 10, 145 10.25, 145.25
+10.25, 145.25 10, 145 10)'`
+
+The PostGIS query requires a valid SRID for the input geometry, so we
+explicitly define that here. This is the SRID that the footprints are being
+stored in inside of the data store. `srid = 949900`.
```
ladoramkershner, line 111: Can they make up any value that they want as long as it is carried through, can they use srid from the systems listed on that spatial reference site (I cannot find it, or I would post the link)

README.md
```diff
@@ -27,3 +27,163 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
...
+
+The query string is then constructed: `query = f"SELECT * FROM ctx WHERE
+ST_INTERSECTS(geom, ST_Polygon(ST_GeomFromText('{geom}'), {srid})) = TRUE"` 
+
+Finally, our database associated with the NCG is populated and the image data
+are copied. `ncg.add_from_remote_database(source_db_config, outpath,
+query_string=query)`
+
+### Operations on the NCG
+After we have an NCG, we want to perform operations on the graph or on database
```
ladoramkershner, line 125: Give examples of the purely graph objects
'perform operations on the graph (Node or Edges) or on...'

README.md
```diff
@@ -27,3 +27,163 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
...
+    from math import ceil
+    return ceil(round(x,1)*8)
+
+# Define a function to govern the distribution of points in the East/West direction
+def ew(x):
+    from math import ceil
+    return ceil(round(x,1)*1)
+
+# Pack a set of kwargs into a keyword that the called function is expecting
+distribute_points_kwargs = {'nspts_func':ns, 'ewpts_func':ew, 'method':'new'}
```
ladoramkershner, line 150: the method keyword is a little confusing because it actually gets used in a function of the function you pass it to (passing to place_points_in_overlap, but used in cg.distribute_points_in_geom). I am not sure if that should be explained here, or if this warrants a code change..

README.md
```diff
@@ -27,3 +27,163 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
...
+
+```
+njobs = ncg.apply('spatial.overlap.place_points_in_overlap',
+                  on='overlaps',
+                  cam_type='isis',
+                  distribute_points_kwargs=distribute_points_kwargs)
+```
+
+Here, we are applying the `spatial.overlap.place_points_in_overlap` function on
+an iterable (`overlaps`) with two keyword arguments (that the function is
```
ladoramkershner, line 170: You actually gave three keyword arguments.

README.md
jlaura: Good point. I punted by pointing to the example below. I think offering a from_filelist example would also be good.

README.md
jlaura: :+1:

README.md
jlaura: I added a section at the end. It is short, but only because the operation is simple.

README.md
jlaura: I removed it since it isn't pertinent to the example and does cause confusion. Changing the name would be totally cool here.

README.md
jlaura: Fixed.

README.md
jlaura: Haha. Fixed.

README.md
ladoramkershner: I am not sure we can assume all users will want to work with a database right away. I can say from experience, autocnet's full gamut is a lot all at once.

README.md
```diff
@@ -27,3 +27,187 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
...
+njobs = ncg.apply('network_to_matches', on='edges')
+```
+
+After the standard boilerplate, the `network_to_matches` function is applied to
+every edge in the graph. This function takes the points and measures from the
+database and expands them so that every edge now has the pairwise
+(measure-to-measure) information that is frequently quite useful when using
+computer vision techniques. Note that the function to be called is not longer
+being specificed with the import path (e.g.,
+spatial.overlap.place_points_in_overla-). Instead any method on the autocnet Edge or NetworkEdge objects can be called.
```
ladoramkershner, line 213: Can only Edge/NetworkEdge or Node/NetworkNode methods be called on these objects?

jlaura: That is correct.

README.md
jlaura: Added the line with a little text describing what's going on.

ladoramkershner: I saw that! I looks good, I was just reaffirming that I think a from_filelist example should be added so people don't have to have a pre-populated data store source to get images into a database.

README.md
ladoramkershner: So then maybe reword:

"
Note that the function to be called is not longer
being specificed with the import path (e.g.,
spatial.overlap.place_points_in_overlaps) because only Edge or NetworkEdge methods can be called on the autocnet Edge or NetworkEdge objects."

README.md
```diff
@@ -27,3 +27,204 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
 2. Get the Postgresql with Postgis container and run it `docker run --name testdb -e POSTGRES_PASSOWRD='NotTheDefault' -e POSTGRES_USER='postgres' -p 5432:5432 -d mdillon/postgis`
 3. create database template_postgis: `docker exec testdb psql -c 'create database template_postgis;' -U postgres`
 4. Run the test suite: `pytest autocnet`
+
+## Simple Network Examples:
+
+
+### Setup a project
+This first example imports the object that is used to orchestrate jobs, manage
```
ladoramkershner, line 35: This first example imports the NetworkCandidateGraph object, which is used to orchestrate jobs, manage...

README.md
```diff
@@ -27,3 +27,204 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
...
+list of ISIS cube files that have had footprints created (using *footprintinit*).
+
+```
+from autocnet.graph.network import NetworkCandidateGraph
+
+ncg = NetworkCandidateGraph.from_filelist(myimages.lis)
+```
+
+This method can take a bit of time to run if the filelist is large as the data
+are loaded into the database and then a spatial overlay oepration is performed
```
ladoramkershner, line 140: are loaded into the database _sequentially_ and then

jlaura:
Once the CI passes this should be good for a re-review and/or merge. Thanks for the feedback!

PR #445, opened 2020-04-14
Adds support for apply with query and overlay points

jlaura:
I was running into an issue where points were not being placed in overlaps. This was because I was using the 'new' method on CTX image overlaps with a large N/S extent and almost no E/W extent. 

I wanted to be able to place points into overlaps that did not already have points. When the config as refactored, calls to place_points_in_overlaps were replaced with an apply of place_points_in_overlap. The apply func on the network can take a filter kwarg. I started looking at how to pass the filter in for things like doing an inner join or applying a spatial function. Long story short, this would have resulted in a pretty crazy API where filters would need to define operators, external functions, functions on objects, etc. Not something to do.

So, this PR adds a query_string argument to the apply method. In the simple case of attribute equivalency it is probably easier to use the filter and the simple booleans it supports. In the more complex case (such as placing points with a spatial intersection) the method now supports straight SQL.

Piggybacking on this, a nice to have is a relationship between points and overlay. This has been added to the model so that one can do overlay.points and get back a list of the points that intersect a given overlap.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1558,6 +1558,40 @@ class NetworkCandidateGraph(CandidateGraph):
 
         walltime : str
                    in the format Hour:Minute:Second, 00:00:00
+
+        chunksize : int
+                    The maximum number of jobs to submit per job array. Defaults to 1000.
+                    This number may be have an actualy higher or lower limited based on 
+                    how the cluster has been configured.
+        
+        filters : dict
+                  Of simple filters to apply on database rows where the key is the attribute and
+                  the value is a boolean. This is usable only when applying to measures, points, or overlays.
+
+        query_string : str
+                       A SQL query to be applied to the iterable. This is usable only when applying to measures, points, or overlays.
+
+        kwargs : dict
+                 Of keyword arguments passed to the function being applied
+
+        Examples
+        --------
+        Apply a function to the overlay table omitting those overlay rows that already have 
+        points within them and have an area less than a given threshold.
+
+        >>> query_string = 'SELECT overlay.id FROM overlay LEFT JOIN points ON ST_INTERSECTS(overlay.geom, points.geom) WHERE points.id IS NULL AND ST_AREA(overlay.geom) >= 0.0001;'
+        >>> njobs = ncg.apply('spatial.overlap.place_points_in_overlap', on='overlaps', query_string=query_string)
+
+        Apply a function to the overlay table and pass keyword arguments (kwargs) to the function.
+
+        >>> def ns(x):
+                from math import ceil
+                return ceil(round(x,1)*8)
+        >>> def ew(x):
+                from math import ceil
+                return ceil(round(x,1)*2)
+        >>> distribute_points_kwargs = {'nspts_func':ns, 'ewpts_func':ew, 'method':'classic'}
+        >>> njobs = ncg.apply('spatial.overlap.place_points_in_overlap', on='overlaps', distribute_points_kwargs=distribute_points_kwargs)
         """
 
         # Determine which obj will be called
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1488,20 +1488,28 @@ class NetworkCandidateGraph(CandidateGraph):
             self.redis_queue.rpush(self.processing_queue, json.dumps(msg, cls=JsonEncoder))
         return job_counter + 1
     
-    def _push_row_messages(self, query_obj, on, function, walltime, filters, args, kwargs):
+    def _push_row_messages(self, query_obj, on, function, walltime, filters, query_string, args, kwargs):
         """
         Push messages to the redis queue for DB objects e.g., Points, Measures 
         """
-        with self.session_scope() as session:
-            query = session.query(query_obj)
-        
-            # Now apply any filters that might be passed in.
-            for attr, value in filters.items():
-                query = query.filter(getattr(query_obj, attr)==value)
-            
-            # Execute the query to get the rows to be processed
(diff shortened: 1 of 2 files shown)


autocnet/io/db/model.py
```diff
@@ -229,6 +229,12 @@ class Overlay(BaseMixin, Base):
     intersections = Column(ARRAY(Integer))
     #geom = Column(Geometry(geometry_type='POLYGON', management=True))  # sqlite
     _geom = Column("geom", Geometry('POLYGON', srid=latitudinal_srid, dimension=2, spatial_index=True))  # postgresql
+    points = relationship('Points',
+                          primaryjoin='func.ST_Contains(foreign(Overlay.geom), Points.geom).as_comparison(1,2)',
+                          backref=backref('overlay', uselist=False),
+                          viewonly=True,
+                          uselist=True)
+
```
ladoramkershner, line 237: The points table already has a overlapid column that contains the overlapid in which place_points_in_overlap created that point. This column does present different information, because there may be more than one overlap a point intersects with even if it is not the overlap of origin. I am just curious if we need both columns or if the addition of this column negates the need for the points.overlapid column (or vice versa).

ladoramkershner:
This functionality used to be taken care of in the cluster_place_points_in_overlaps as the default query string (although it could be overwritten). I am worried this is a little convoluted and puts a lot on the user to understand how kwargs, args, and such will be passed between functions and to which functions they will end up being applied. Should we add an option to place_points_in_overlaps specifically ('all' or 'empty' flags) and then stick with filters for points? 

Or can you see this functionality specifically coming in handy somewhere else in the process?

autocnet/io/db/model.py
jlaura: I think having the id is nice because you can go some really simple queries in Postgres if you know the ID. The column I added is for programmatic access via python. The only reason I added it too was on my journey trying to figure out how to get custom SQL queries passed in to then cg for place_points_in_overlaps in a nice way. This code did not result in anything nice there, but it looked useful in the future going from overlay to points through the overlay object. So I think keeping both makes sense, but I'm totally up for removing this as well.

jlaura:
I am also concerned about this. We could make the default be this like it was and then allow the more advanced user to push in a custom query? My only concern there is that we are hard coding the area threshold, but good documentation should alleviate that.

autocnet/io/db/model.py
ladoramkershner: I say more information is better, but I do worry about over saturating the database as well. This information can be found on the fly, however the overlap of origin cannot. On the OTHER hand I am not sure how handy the overlap of origin is outside of the trouble shooting spatial.overlap.place_point_in_overlap().

I don't mean to lean on you to make a decision, however I do not have the sense of what may come back to haunt us more:
a) Overpopulating the database, then having unnecessary column queries tied up in the code making it difficult to streamline the database if it becomes over saturated.
b) Not readily having access to the information we need, causing delays in processing every time.

ladoramkershner:
I think that is a good idea, but the where should this documentation and default settings live? 

If we do it in place_points_in_overlaps we have to take the time to populate every job and hit the database in every cluster job to do a check. However, it sounds pretty janky to specify it in ncg.apply(), since it is such an abstracted function.

autocnet/io/db/model.py
jlaura: Checkout the database named `adjacent` on smalls. This does not actually add anything to the DB. When the attribute is called on the overlay the query is run to get the results. So this is a 'freebie' from the perspective of storage and a cost from a computational perspective.

ladoramkershner: 👍

jlaura:
Yeah....I need to think about where that might live, because I'm not sure where to put the default.

jlaura:
@ladoramkershner I just pushed an update to the docs. I think docs are the way to go because passing the query string is useful for a lot of application, for example, applying subpixel to a subset of points where the subset is identified by an SQL query. Let me know what you think about the updated docs/examples on apply.

autocnet/graph/network.py
```diff
@@ -1542,6 +1558,40 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', chunksize=100
...
+
+        kwargs : dict
+                 Of keyword arguments passed to the function being applied
+
+        Examples
+        --------
+        Apply a function to the overlay table omitting those overlay rows that already have 
+        points within them and have an area less than a given threshold.
+
+        >>> query_string = 'SELECT overlay.id FROM overlay LEFT JOIN points ON ST_INTERSECTS(overlay.geom, points.geom) WHERE points.id IS NULL AND ST_AREA(overlay.geom) >= 0.0001;'
```
ladoramkershner, line 1582: Break up this query between lines for _aesthetics_

autocnet/graph/network.py
```diff
@@ -1542,6 +1558,40 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', chunksize=100
...
+                    The maximum number of jobs to submit per job array. Defaults to 1000.
+                    This number may be have an actualy higher or lower limited based on 
+                    how the cluster has been configured.
+        
+        filters : dict
+                  Of simple filters to apply on database rows where the key is the attribute and
+                  the value is a boolean. This is usable only when applying to measures, points, or overlays.
+
+        query_string : str
+                       A SQL query to be applied to the iterable. This is usable only when applying to measures, points, or overlays.
```
ladoramkershner, line 1572: Mention here how they are mutually exclusive and highlight instances where filters or queries are better (in their respective sections)

PR #446, opened 2020-04-16
Making to_isis write out sql query ordered like subpixel register

ladoramkershner:
I was not seeing the same reference measures in my written out control network as I was seeing in the database. This was because the to_isis default query did not order the database the same way is done when subpixel registering points. So when plio is handed the unordered data frame, it just picks the first measure it gets per point for reference.

I updated the query to order by pointid and then measure id. My concern with this code update is now it assumes that you will hand it a query containing fields 'id' and 'pointid'. I feel like is necessary for the proper write out, but does limit what a user can do with the sql query line.

### DIFF
--- a/autocnet/cg/cg.py
+++ b/autocnet/cg/cg.py
@@ -433,11 +433,15 @@ def distribute_points_new(geom, nspts, ewpts, Session):
     very simplistic approximation of the shape
     of the geometry and then place some number of
     north/south and east/west points into the geometry.
+    This function works best on bulky geometries, such as
+    a combination of all network image footprints.
 
     Parameters
     ----------
     geom : shapely.geom
-           A shapely geometry object
+           A shapely geometry object which is a union of
+           all of the image footprints in the network
+           that is being grounded.
 
     nspts : int
             The number of points to attempt to place
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1352,7 +1352,7 @@ class NetworkCandidateGraph(CandidateGraph):
     def config_from_file(self, filepath):
         """
         A NetworkCandidateGraph uses a database. This method parses a config
-        file to set up the connection. Additionally, this loads planetary 
+        file to set up the connection. Additionally, this loads planetary
         information and settings for other operations the candidate graph
         can perform.
 
@@ -1367,7 +1367,7 @@ class NetworkCandidateGraph(CandidateGraph):
     def config_from_dict(self, config_dict):
         """
         A NetworkCandidateGraph uses a database. This method loads a config
-        dict to set up the connection. Additionally, this loads planetary 
+        dict to set up the connection. Additionally, this loads planetary
         information and settings for other operations the candidate graph
         can perform.
 
@@ -1380,7 +1380,7 @@ class NetworkCandidateGraph(CandidateGraph):
 
         # Setup REDIS
         self._setup_queues()
-       
+
         # Setup the database
         self._setup_database()
 
@@ -1455,7 +1455,7 @@ class NetworkCandidateGraph(CandidateGraph):
         conn = self.engine.connect()
         conn.execute(sql)
         conn.close()
-    
+
     def _push_obj_messages(self, onobj, function, walltime, args, kwargs):
         """
         Push messages to the redis queue for objects e.g., Nodes and Edges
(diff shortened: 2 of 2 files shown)


autocnet/cg/cg.py
```diff
@@ -433,11 +433,15 @@ def distribute_points_new(geom, nspts, ewpts, Session):
...
+    This function works best on bulky geometries, such as
+    a combination of all network image footprints.
 
     Parameters
     ----------
     geom : shapely.geom
-           A shapely geometry object
+           A shapely geometry object which is a union of
+           all of the image footprints in the network
+           that is being grounded.
```
ladoramkershner, line 444: Hopefully clearing up the use of this function (I am thinking of fully renaming in a separate PR)

ladoramkershner:
I also have a couple ryders on this PR that I will comment on in the file area

autocnet/graph/network.py
```diff
@@ -1779,17 +1786,23 @@ def copy_images(self, newdir):
...
                 filename = os.path.basename(oldpath)
                 obj.path = os.path.join(newdir, filename)
-                oldnew.append((oldpath, obj.path))
+                if oldpath != obj.path:
+                    # this method is slower but allows you to run again if connection failed partway through
+                    copyfile(oldpath, obj.path)
+                    session.commit()
+                else:
+                    continue
+                # oldnew.append((oldpath, obj.path))
```
ladoramkershner, line 1800: Reworked this function, so that if a ncg.add_from_remote_database() is interrupted (like it was for me...TWICE), you can specifically run this file to finish copying over the files. It is slower, but maybe more effective for the reality of loading large areas.

autocnet/graph/network.py
```diff
@@ -1547,38 +1547,38 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', chunksize=100
...
-        
+
         submitter = Slurm(command,
                      job_name='AutoCNet',
                      mem_per_cpu=self.config['cluster']['processing_memory'],
                      time=walltime,
                      partition=self.config['cluster']['queue'],
                      output=self.config['cluster']['cluster_log_dir']+f'/autocnet.{function}-%j')
-        submitter.submit(array='1-{}'.format(job_counter), chunksize=chunksize)
+        submitter.submit(array='1-{}%24'.format(job_counter), chunksize=chunksize)
```
ladoramkershner, line 1581: Limiting array job submissions, so we don't look like total jerks. You can get more than 24 jobs by manipulating chunksize to split the jobs youare about to submit in n array jobs.

autocnet/graph/network.py
```diff
@@ -1779,17 +1786,18 @@ def copy_images(self, newdir):
...
                 oldpath = obj.path
                 filename = os.path.basename(oldpath)
                 obj.path = os.path.join(newdir, filename)
-                oldnew.append((oldpath, obj.path))
+                if oldpath != obj.path:
+                    # Copy the files
+                    copyfile(oldpath, obj.path)
+                    session.commit()
+                else:
+                    continue
```
ladoramkershner, line 1798: Reworked this function, so that if a ncg.add_from_remote_database() is interrupted (like it was for me...TWICE), you can specifically run this file to finish copying over the files. It is slower, but maybe more effective for the reality of loading large areas.

autocnet/cg/cg.py
jlaura: :+1:

autocnet/graph/network.py
jlaura: :+1:

autocnet/graph/network.py
jlaura: Does this need the commit in there? This is inside of the scope_session so commit is called when the context (the indented block of code that starts with 'with') is exited. I suspect the commit is so that the data aren't lost if something drops?

autocnet/graph/network.py
```diff
@@ -1847,8 +1855,7 @@ def add_from_remote_database(self, source_db_config, path,  query_string='SELECT
 
         sourceimages = sourcesession.execute(query_string).fetchall()
 
-        with session_scope() as destinationsession:
-            destinationsession = self.Session()
+        with self.session_scope() as destinationsession:
```
jlaura, line 1858: :+1: Good catch.

jlaura: This is where the merge conflict is. I think if you rebase off dev your local copy will show it too. It's a really easy merge after that.

PR #447, opened 2020-04-16
Adding template and phase match metrics and shifts to DB

ladoramkershner:
This is meant to be used for creating effective measure elimination metric based on its associated matching information. DO NOT MERGE

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1578,7 +1578,7 @@ class NetworkCandidateGraph(CandidateGraph):
                      time=walltime,
                      partition=self.config['cluster']['queue'],
                      output=self.config['cluster']['cluster_log_dir']+f'/autocnet.{function}-%j')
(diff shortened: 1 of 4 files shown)


autocnet/graph/network.py
```diff
@@ -1578,7 +1578,7 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', chunksize=100
                      time=walltime,
                      partition=self.config['cluster']['queue'],
                      output=self.config['cluster']['cluster_log_dir']+f'/autocnet.{function}-%j')
-        submitter.submit(array='1-{}%24'.format(job_counter), chunksize=chunksize)
+        submitter.submit(array='1-{}%25'.format(job_counter), chunksize=chunksize)
```
ladoramkershner, line 1581: Really getting greedy with nebula's resources with this change.

PR #452, opened 2020-05-05
Necessary fixes discovered during grounded product creation

ladoramkershner:
Most changes are small bug fixes, with the only new code implementation being a new ground point grid creation. 

Previously I tried to evaluate the most east/west cubes in an area to help create a slanted grid along the travel direction of the images. However, this most east/west cube evaluation was too vague and ended up being incorrect for the area I was grounding.

I replaced this code with a more robust method that calculates an affine transformation between two sets of points (envelope and minimum_rotated_rectangle) that are existing properties on the footprint object. It then lays down a normal grid and applies the affine transformation to get a grid with a track direction slant.

### DIFF
--- a/autocnet/cg/cg.py
+++ b/autocnet/cg/cg.py
@@ -7,6 +7,7 @@ import networkx as nx
 import geopandas as gpd
 import ogr
 
+from skimage import transform as tf
 from scipy.spatial import ConvexHull
 from scipy.spatial import Voronoi
 import shapely.geometry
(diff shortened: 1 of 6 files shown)


autocnet/graph/network.py
```diff
@@ -1784,8 +1784,9 @@ def update_from_jigsaw(self, session, path):
         DROP TABLE temp_measures;
         """
 
-        session.execute(sql)
-        session.commit()
+        with self.session_scope() as session:
```
ladoramkershner, line 1787: Just getting this up to date with Jay's session refactor (PR #438 )

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -123,40 +127,56 @@ def generate_ground_points(Session, ground_mosaic, nspts_func=lambda x: int(roun
...
     ground_cnet["path"] = [ground_mosaic.file_name]*len(coord_list)
     ground_cnet["pointid"] = list(range(len(coord_list)))
+    ground_cnet["original point"] = old_coord_list
     ground_cnet["point"] = coord_list
-    ground_cnet['line'] = lines
-    ground_cnet['sample'] = samples
+    ground_cnet['original line'] = lines
+    ground_cnet['original sample'] = samples
+    ground_cnet['line'] = newlines
+    ground_cnet['sample'] = newsamples
```
ladoramkershner, line 154: Just adding more information. I found these values really helpful when troubleshooting

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -123,40 +127,56 @@ def generate_ground_points(Session, ground_mosaic, nspts_func=lambda x: int(roun
...
                     verbose=False):
     """
 
     """
-    engine = Session.get_bind()
-    images = gpd.GeoDataFrame.from_postgis(f"select * from images where ST_Intersects(geom, ST_SetSRID(ST_Point({lon}, {lat}), {config['spatial']['latitudinal_srid']}))", engine, geom_col="geom")
+    session = Session()
+    engine = session.get_bind()
+    string = f"select * from images where ST_Intersects(geom, ST_SetSRID(ST_Point({lon}, {lat}), {config['spatial']['latitudinal_srid']}))"
+    images = pd.read_sql(string, engine)
```
ladoramkershner, line 178: from_postgis did not seem to like working with our SRID, read_sql did not throw a fit

autocnet/cg/cg.py
```diff
@@ -456,43 +457,36 @@ def distribute_points_new(geom, nspts, ewpts, Session):
...
-    ur_actual = np.array(two_point_extrapolate(ur[1], elat, elon))[::-1]
-    lr_actual = np.array(two_point_extrapolate(lr[1],elat, elon))[::-1]
-
-    wlon, wlat = find_side('west', Session)
-    ul_actual = np.array(two_point_extrapolate(ul[1], wlat, wlon))[::-1]
-    ll_actual = np.array(two_point_extrapolate(ll[1], wlat, wlon))[::-1]
+    rr_coords = np.array(list(zip(*geom.minimum_rotated_rectangle.exterior.xy))[:-1])
+    w = min([i[0] for i in rr_coords])
+    s = min([i[1] for i in rr_coords])
+    SWid = nearest([w, s], rr_coords)
```
jlaura, line 469: pedantic - can we make this lowercase. I was like, 'what class is this?'.

autocnet/cg/cg.py
```diff
@@ -456,43 +457,36 @@ def distribute_points_new(geom, nspts, ewpts, Session):
...
-    newtop = create_points_along_line(ul_actual+dt, ur_actual-dt, ewpts)
-    newbot = create_points_along_line(ll_actual+db, lr_actual-db, ewpts)
+    top = create_points_along_line(ul, ur, ewpts)
+    bot = create_points_along_line(ll, lr, ewpts)
 
-    points = []
-    for i in range(len(newtop)):
-        top = newtop[i]
-        bot = newbot[i]
+    grid = create_points_along_line(top[0], bot[0], nspts)
```
jlaura, line 475: I think this is totally fine here. Another way to do this would be to use numpy:

```python
xstep = 5
xmin = 0
xmax = 10
x = np.linspace(xmin, xmax, xstep)
ystep = 3
ymin = 5
ymax = 15
y = np.linspace(ymin, ymax, ystep)

coords = np.transpose([np.tile(x, len(y)), np.repeat(y, len(x))])
```

This would get rid of a bunch of the logic to draw points along these lines. I think this should work because you need a regular grid to start and then you check intersection and rotate it

autocnet/graph/network.py
jlaura: I think that the test needs to be updated?

autocnet/matcher/cross_instrument_matcher.py
jlaura: underscores in the names instead of spaces?

autocnet/matcher/cross_instrument_matcher.py
jlaura: :+1:

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -231,8 +255,37 @@ def propagate_point(Session,
 
     return new_measures
 
-def propagate_control_network(Session, config, dem, base_cnet, verbose=False):
+def propagate_control_network(Session, config, dem, base_cnet, \
+        size_x=40, size_y=40, \
```
jlaura, line 259: `\` aren't needed inside the '()' for a func signature.

autocnet/cg/cg.py
ladoramkershner: 👍

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: :+1:

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: :+1:

autocnet/graph/network.py
ladoramkershner: :+1:

ladoramkershner: Will resolve when tests are passing

PR #454, opened 2020-05-06
Adds in automatic covariance computation on writes

jlaura:
This PR adds in covariance matrix computation for constrained and fixed points. This happens all at once on write, so a user can't specify per point sigmas yet, but that could be refactored in in the future should it be valuable.


autocnet/graph/network.py
```diff
@@ -1728,6 +1735,21 @@ def to_isis(self, path, flistpath=None,sql = """
...
+            """
+            geom = shapely.wkt.loads(row.geom)
+            covar = covariance.compute_covariance(geom.y, 
+                                                  geom.x, 
+                                                  radius, 
+                                                  latsigma=latsigma, 
+                                                  lonsigma=lonsigma, 
+                                                  radsigma=radsigma, 
+                                                  semimajor_axis=None)
+            return covar
```
ladoramkershner, line 1750: I am not sure how I feel about this function sitting inside the ncg.to_isis method. Is this a normal convention? Is there value in have this be a method on the ncg?

ladoramkershner:
Should we add tests for this?

autocnet/graph/network.py
jlaura: This is normal in that the function is only ever called on each row in the data frame that is created inside this method. In other words the inner method is tightly coupled to the data structure that the outer method creates. I can only think of maybe 2 cases where I would use this approach. All have to do with how the variables are being scoped.

ladoramkershner: I was more saying, does calculating have value outside of being written out to a ISIS control network. Should it be this coupled or should we have it more general and directly tied to the ncg?

Additionally, I pulled this locally and you call row.geom, which is not a field in the DB/SQL query that generates the DF the row is being taken from. Also shapely.wkt is imported in this module as 'swkt', but if you look at how to_isis is prepping the geoms, it used swkb.loas(hex=True)

ladoramkershner: So essential, replace line 1742 with 
```
geom = swkb.loads(row['adjusted'], hex=True)
```

jlaura: Fixed the geom issue in the latest push

jlaura:
Hmm - scope of changes looks too large (for example the session scope stuff) but I do not see the commit where that is merged in....any ideas?

autocnet/graph/network.py
```diff
@@ -1753,6 +1686,11 @@ def to_isis(self, path, flistpath=None,sql = """
             if f not in fpaths:
                 warnings.warn(f'{f} in candidate graph but not in output network.')
 
+        # Remap the df columns back to ISIS
+        df.rename(columns={'pointtype':'pointType',
```
jlaura, line 1690: I did this because the logic for pulling the df is now removed from here, so I am trying to normalize the column names to the python standard (snake case at 'worst') and letting the function that hits the DB worry about camelCase renaming.

docs/library/io/db/controlnetwork.rst
```diff
@@ -0,0 +1,10 @@
+:mod:`io.db.controlnetwork` --- Database to Control Networks
+============================================================
+
+The :mod:`io.db.controlnetwork` module contains logic for reading a database stored control network into a pandas dataframe.
+
+.. versionadded:: 0.1.0
+
+.. automodule:: autocnet.io.db.controlnetwork
+   :synopsis: Database to control network logic
+   :members:
```
ladoramkershner, line 10: Should this be filled out?

autocnet/graph/network.py
ladoramkershner: 👍

ladoramkershner:
I tested the actual covariance logic on a local pull of the PR, the rest of this just looks like breaking up the logic!

docs/library/io/db/controlnetwork.rst
jlaura: `:members:` auto populates with all of the classes/functions in the file. Does that answer your question?

autocnet/io/db/controlnetwork.py
```diff
@@ -0,0 +1,87 @@
...
+    measures."measureJigsawRejected" = FALSE AND
+    measures."imageid" NOT IN
+        (SELECT measures."imageid"
+        FROM measures
+        INNER JOIN points ON measures."pointid" = points."id"
+        WHERE measures."measureIgnore" = False and measures."measureJigsawRejected" = False AND points."pointIgnore" = False
+        GROUP BY measures."imageid"
+        HAVING COUNT(DISTINCT measures."pointid")  < 3)
+ORDER BY measures."pointid", measures."id";
+"""):
```
Kelvinrr, line 35: Minor detail, but would it be better to maybe save this as a separate string in the module to make this slightly more readable?

jlaura: I can absolutely do that. Comfortable with that going into io.db.sql.py and then importing that where it is needed? I would probably put it into a py file so that it can be assigned a variable and so that every SQL statement doesn't require an individual file.

docs/library/io/db/controlnetwork.rst
ladoramkershner: Sure does!

autocnet/io/db/controlnetwork.py
jlaura: https://stackoverflow.com/questions/30218963/storing-sql-statements-in-a-properties-file-to-be-used-by-python-scripts for some thoughts on different approaches. The YAML approach looks nice because it is possible to have both a doc string and the SQL. Soliciting thoughts?

ladoramkershner: I think that will only touch spatial/overlap.py and network/graph.py, so should that be added to this PR?

jlaura: I opened #455 as an issue to consolidate the SQL. @ladoramkershner I like the idea of keeping that a separate PR if only because it hopefully also means the PR has some additional tests. I am good to merge this if others agree.

ladoramkershner: I am good to merge! 

I also like the idea of keeping the two changes separate. Otherwise, that sql consolidation is a rider on this PR and would be hard to find again if needed.

PR #456, opened 2020-06-30
Added modified change detection algorithm

Kelvinrr:
Still not sure if Autocnet is the best place for this. On one hand it fits autocnet's functionality very well, but on the other hand, it has nothing to do with generating cnets. I am hesitant to start a new repo/project. I am open to suggestions if people feel this should be moved elsewhere. 

@jlaura thoughts?

### DIFF
--- a/autocnet/cg/change_detection.py
+++ b/autocnet/cg/change_detection.py
@@ -18,7 +18,7 @@ def image_diff(arr1, arr2):
      diff = arr1-arr2
      diff[np.isnan(diff)] = 0
 
-     return bytescale(diff)
+     return diff
 
 
 def okubogar_detector(image1, image2, nbins=50, extractor_method="orb", extractor_kwargs={"nfeatures": 2000, "scaleFactor": 1.1, "nlevels": 1}, cluster_params={"min_samples": 10, "max_eps": 10, "eps": .5, "xi":.5},  image_func=image_diff):
--- a/autocnet/cg/change_detection.py
+++ b/autocnet/cg/change_detection.py
@@ -44,10 +44,10 @@ def okubogar_detector(image1, image2, nbins=50, extractor_method="orb", image_fu
      ----------
 
      image1 : np.array, plio.GeoDataset
-             Image representing the "before" state of the ROI
+             Image representing the "before" state of the ROI, can be a 2D numpy array or plio GeoDataset
 
      image2 : np.array, plio.GeoDataset
-             Image representing the "after" state of the ROI
+             Image representing the "after" state of the ROI, can be a 2D numpy array or plio GeoDataset
 
      image_func : callable
                   Function used to create a derived image from image1 and image2, which in turn is
@@ -122,10 +122,10 @@ def okbm_detector(image1, image2, nbins=50, extractor_method="orb",  image_func=
      ----------
 
      image1 : GeoDataset
-             Image representing the "before" state of the ROI
+             Image representing the "before" state of the ROI, can be a 2D numpy array or plio GeoDataset
 
      image2 : GeoDataset
-             Image representing the "after" state of the ROI
+             Image representing the "after" state of the ROI, can be a 2D numpy array or plio GeoDataset
 
      image_func : callable
                   Function used to create a derived image from image1 and image2, which in turn is
--- a/autocnet/cg/change_detection.py
+++ b/autocnet/cg/change_detection.py
@@ -93,7 +93,7 @@ def okubogar_detector(image1, image2, nbins=50, extractor_method="orb", image_fu
      return points, heatmap, bdiff
 
 
-def obm_detector(image1, image2, nbins=50, extractor_method="orb",  image_func=image_diff,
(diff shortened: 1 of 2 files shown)


autocnet/graph/edge.py
```diff
@@ -343,6 +343,7 @@ def compute_fundamental_matrix(self, clean_keys=[], maskname='fundamental', **kw
         _, mask = self.clean(clean_keys)
         s_keypoints, d_keypoints = self.get_match_coordinates(clean_keys=clean_keys)
         self.fundamental_matrix, fmask = fm.compute_fundamental_matrix(s_keypoints, d_keypoints, **kwargs)
+        fmask = fmask.flatten()
```
Kelvinrr, line 346: fmask was coming back as a 2D shape? Only in the offline version though,

jlaura: I bet it was coming back as a (n,1) array and we needed it to be an (n) vector?  That's just a guess though.

jlaura:
I am torn. We have quite a bit of CV logic in the library. We also have photogrammetric logic and outlier detection stuff, etc. I can see an argument for pushing this logic in (and then maybe having to rename the library at some point in the future.) I can also see an argument for keeping this logic out in a distinct CV library and then refactoring out additional CV stuff.

Right now, I would probably go the former and use `autocnet` like a meta library that has modules for doing different things. Our focus has been really hard in the graph (network) section of the library, but perhaps this is an opportunity to expand the logic for change detection? This also reduces fragmentation in our larger collection of software.

I am comfortable with either route.

autocnet/cg/change_detection.py
```diff
@@ -18,10 +19,114 @@ def image_diff(arr1, arr2):
...
+     TODO: Paper/abstract might exist, cite
+
+     Parameters
+     ----------
+
+     image1 : GeoDataset
+             Image representing the "before" state of the ROI
+
+     image2 : GeoDataset
+             Image representing the "after" state of the ROI
```
jlaura, line 51: Does it makes sense to identify these as plio image? Just in case an external caller wants to make use of the code.

autocnet/cg/change_detection.py
```diff
@@ -18,10 +19,114 @@ def image_diff(arr1, arr2):
...
+
+     image_func : callable
+              Function use to create a derived image from image1 and image2, which in turn is
+              the input for the feature extractor. Default function create a difference image.
```
jlaura, line 55: I like this - it continues the API in a really functional way. Have you considered an example in the docs string showing some other callable being passed in? Or perhaps beefing up this docstring to show what the output of the callable needs to be?

autocnet/cg/change_detection.py
```diff
@@ -18,10 +19,114 @@ def image_diff(arr1, arr2):
...
+
+     See Also
+     --------
+
+     feature extractor: autocnet.matcher.cpu_extractor.extract_features
+
+     """
+
+     arr1 = image1.read_array()
+     arr2 = image2.read_array()
```
jlaura, line 75: Another option here is to make this method take the arrays themselves? If this func using the GeoDataset object for anything other than the read?

autocnet/cg/change_detection.py
```diff
@@ -18,10 +19,114 @@ def image_diff(arr1, arr2):
...
+
+     keys, descriptors = extract_features(bdiff, extractor_method, extractor_parameters=extractor_kwargs)
+     x,y = keys["x"], keys["y"]
+
+     points = [Point(xval, yval) for xval,yval in zip(x,y)]
+
+     heatmap, xedges, yedges = np.histogram2d(y, x, bins=nbins, range=[[0, bdiff.shape[0]], [0, bdiff.shape[1]]])
+     heatmap = cv2.resize(heatmap, dsize=(bdiff.shape[1], bdiff.shape[0]), interpolation=cv2.INTER_NEAREST)
+
+     return points, heatmap, bdiff
```
jlaura, line 91: It is really nice to see just how much of the lib is being used to make this func. :+1:

autocnet/cg/change_detection.py
```diff
@@ -18,10 +19,114 @@ def image_diff(arr1, arr2):
...
+
+
+     Parameters
+     ----------
+
+     image1 : GeoDataset
+             Image representing the "before" state of the ROI
+
+     image2 : GeoDataset
+             Image representing the "after" state of the ROI
```
jlaura, line 109: Same as above re: GeoDataset from plio and/or switching to take arrays (or taking either).

autocnet/graph/edge.py
Kelvinrr: yeah, basically a one column array.

autocnet/cg/change_detection.py
Kelvinrr: It was originally written against with opening Cubes as GeoDatasets and it wasn't really changed, I'll just switch it to use either.

Kelvinrr:
@jlaura +1 on not making our portfolio more fragmented if we don't have to.

PR #457, opened 2020-07-02
Database/Project Merging

jlaura:
This PR executes a few refactors in order to support merging autocnet projects/databases. The most significant is that the underlying database is now using schemas. From my reading, we should be using schemas anyway because the schemas are designed to support a higher number of instances, within a single database. Apparently postgres can be unhappy when the instance has a high number of discrete databases.

As it stands now, the PR merges all the easy stuff: images, overlaps, non-overlapping points. Working on the overlapping areas where points have to be merged still.

### DIFF
--- a/autocnet/io/db/connection.py
+++ b/autocnet/io/db/connection.py
@@ -4,6 +4,9 @@ import sqlalchemy
 from sqlalchemy import create_engine, pool, orm
 from sqlalchemy.orm import create_session, scoped_session, sessionmaker
 
+from sqlalchemy.ext.compiler import compiles
+from sqlalchemy.sql.expression import Insert
+
 import os
 import socket
 import warnings
@@ -41,3 +44,7 @@ def new_connection(dbconfig):
                 isolation_level="AUTOCOMMIT")
     Session = orm.sessionmaker(bind=engine, autocommit=False)
     return Session, engine
+
+@compiles(Insert)
+def prefix_inserts(insert, compiler, **kw):
+    return compiler.visit_insert(insert, **kw) + " ON CONFLICT DO NOTHING"
--- a/autocnet/io/db/connection.py
+++ b/autocnet/io/db/connection.py
@@ -37,6 +37,7 @@ def new_connection(dbconfig):
     engine = sqlalchemy.create_engine(db_uri,
                 poolclass=sqlalchemy.pool.NullPool,
                 connect_args={"application_name":f"AutoCNet_{hostname}"},
+                execution_options={'schema_translate_map':{'dynamic' : dbconfig['schema']}},
                 isolation_level="AUTOCOMMIT")
     Session = orm.sessionmaker(bind=engine, autocommit=False)
     return Session, engine
--- a/autocnet/io/db/connection.py
+++ b/autocnet/io/db/connection.py
@@ -47,4 +47,4 @@ def new_connection(dbconfig):
 
 @compiles(Insert)
 def prefix_inserts(insert, compiler, **kw):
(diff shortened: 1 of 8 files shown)


autocnet/io/db/connection.py
```diff
@@ -37,6 +40,11 @@ def new_connection(dbconfig):
...
                 poolclass=sqlalchemy.pool.NullPool,
                 connect_args={"application_name":f"AutoCNet_{hostname}"},
+                execution_options={'schema_translate_map':{'dynamic' : dbconfig['schema']}},
                 isolation_level="AUTOCOMMIT")
     Session = orm.sessionmaker(bind=engine, autocommit=False)
     return Session, engine
+
+@compiles(Insert)
+def prefix_inserts(insert, compiler, **kw):
+    return compiler.visit_insert(insert, **kw) + " ON CONFLICT DO NOTHING"
```
jlaura, line 50: This makes it so that inserts when the unique constraint would fail do not throw an error. Instead the INSERT silently never occurs.

autocnet/io/db/model.py
```diff
@@ -101,7 +126,7 @@ class Keypoints(BaseMixin, Base):
     __tablename__ = 'keypoints'
     latitudinal_srid = -1
     id = Column(Integer, primary_key=True, autoincrement=True)
-    image_id = Column(Integer, ForeignKey("images.id", ondelete="CASCADE"))
+    image_id = Column(Integer, ForeignKey("dynamic.images.id", ondelete="CASCADE"))
```
jlaura, line 129: dynamic here is a key that is dynamically replaced for each NetworkCandidateGraph. So the model is now more like a meta-model describing the tables across schemas. The PK/FK relationships are maintained within a schema because the ORM dynamically replaces 'dynamic' with the NCG instance schema name.

autocnet/io/db/model.py
```diff
@@ -401,6 +429,18 @@ def try_db_creation(engine, config):
...
+    if not engine.dialect.has_schema(engine, schema_name):
+        engine.execute(sqlalchemy.schema.CreateSchema(schema_name))
+    
+    #meta = sqlalchemy.MetaData(schema=schema_name)
+    
+    """# Set the schema for the table objs
+    for cls in [Measures, Overlay, Edges, Costs, Matches, Cameras, Points, 
+                Images, Keypoints]:
+        setattr(getattr(cls, '__table__'), 'schema', schema_name)"""
+
```
jlaura, line 443: I should remove this block.

autocnet/spatial/overlap.py
```diff
@@ -27,16 +27,16 @@
...
-    WHERE images.geom is NOT NULL AND
-    ST_INTERSECTS(intersectiongeom.geom, images.geom) AND
-    ST_AREA(ST_INTERSECTION(intersectiongeom.geom, images.geom)) > 0.000001
+    FROM {0}images, intersectiongeom
+    WHERE {0}images.geom is NOT NULL AND
+    ST_INTERSECTS(intersectiongeom.geom, {0}images.geom) AND
+    ST_AREA(ST_INTERSECTION(intersectiongeom.geom, {0}images.geom)) > 0.000001
 )
-INSERT INTO overlay(intersections, geom) SELECT row.intersections, row.geom FROM
+INSERT INTO {0}overlay(intersections, geom) SELECT row.intersections, row.geom FROM
```
jlaura, line 39: Using schemas, our SQL calls then require a schema definition that we can dynamically populate.

jlaura:
I am going to close. Using schemas is not super well supported in SQLAlchemy, unfortunately.

PR #458, opened 2020-07-08
Initial blob_detector function

AustinSanders:
Prototype blob_detector functionality.

Currently returns 2 arrays -- one for "type 1 changes" and one for "type 2 changes."  One is "there used to be a feature in this location, now there is not," and the other is "there didn't used to be a feature here, but now there is."  This still needs to be better documented in the code, but I'm having trouble identifying which is which, particularly due to the fact that it is dependent on the order of images.   Documentation updates will be included in later commits.

Another important note is that the logic only allows for sub solar azimuth for one image (or a scalar to be used for the entire image), but the azimuth values are almost certainly different between images.  I'm not sure how to account for this.

There's one "readable" list comprehension and one ugly list comprehension.  I'd prefer to make it more readable, but not at the expense of efficiency.  Any hints would be appreciated.

### DIFF
--- a/autocnet/cg/change_detection.py
+++ b/autocnet/cg/change_detection.py
@@ -219,8 +219,9 @@ def okbm_detector(image1, image2, nbins=50, extractor_method="orb",  image_func=
      return polys, weights, bdiff
 
 
-def blob_detector(image1, image2, sub_solar_azimuth, image_func=image_diff, max_sigma=30, num_sigma=10,
-        threshold=.075, n_neighbors=3, dist_upper_bound=5, angle_tolerance=10):
+def blob_detector(image1, image2, sub_solar_azimuth, image_func=image_diff,
+                  subtractive=False, max_sigma=30, num_sigma=10, threshold=.075,
+                  n_neighbors=3, dist_upper_bound=5, angle_tolerance=10):
      """
 
      Parameters
@@ -258,6 +259,11 @@ def blob_detector(image1, image2, sub_solar_azimuth, image_func=image_diff, max_
                   >>> image1, image2 = random.random((50,50)), random.random((50,50))
                   >>> results = okbm_detector(image1, image2, image_func=lambda im1, im2: im1/im2)
 
+     subtractive : Boolean
+                   Find subtractive features instead of additive features.  In other
+                   words, find locations in which a feature "used to be present"
+                   but has since moved.
+
      max_sigma : scalar or sequence of scalars
                  The maximum standard deviation for Gaussian kernel. Keep this
                  high to detect larger blobs. The standard deviations of the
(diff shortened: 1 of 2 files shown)


autocnet/cg/change_detection.py
```diff
@@ -216,3 +219,135 @@ def okbm_detector(image1, image2, nbins=50, extractor_method="orb",  image_func=
...
+     neighbors = [blobs_log[j] for j in [i[i!=len(blobs_log)]for i in idx_log] if j.size > 0]
+
+     def is_azimuth_colinear(pt1, pt2, subsolar_azimuth, tolerance):
+         """ Returns true if pt1, pt2, and subsolar azimuth are colinear within
+             some tolerance.
+         """
+         x, y = (pt2[1]-pt1[1], pt2[0]-pt1[0])
+         # Find angle of vector w.r.t. x axis
+         angle = (atan2(y, x) * 180 / pi)%360
+         return -tolerance <= subsolar_azimuth - angle <= tolerance
```
Kelvinrr, line 336: I would put this at the top of the function

autocnet/cg/change_detection.py
```diff
@@ -216,3 +219,135 @@ def okbm_detector(image1, image2, nbins=50, extractor_method="orb",  image_func=
...
+     for idx, pt1 in enumerate(close_points):
+         for pt2 in neighbors[idx]:
+             try:
+                 azimuth = sub_solar_azimuth[int(pt1[0]), int(pt1[1])]
+             except IndexError as e:
+                 azimuth = sub_solar_azimuth
+             if is_azimuth_colinear(pt1, pt2, azimuth, angle_tolerance):
+                 type1_change.append([pt1,pt2])
+             elif is_azimuth_colinear(pt2, pt1, azimuth, angle_tolerance):
+                 type2_change.append([pt1,pt2])
```
Kelvinrr, line 349: same with this, I believe this should just be configured to do one or the other.

autocnet/cg/change_detection.py
```diff
@@ -216,3 +219,135 @@ def okbm_detector(image1, image2, nbins=50, extractor_method="orb",  image_func=
...
+     image1[image1 == image1.min()] = 0
+     image2[image2 == image2.min()] = 0
+     arr1 = bytescale(image1)
+     arr2 = bytescale(image2)
+
+     bdiff = image_func(arr1, arr2)
+
+     # Laplacian of Gaussian only finds light blobs on a dark image.  In order to
+     #  find dark blobs on a light image, we invert.
+     inv = 255-bdiff
```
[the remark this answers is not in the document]
AustinSanders: I think I misunderstood what you were talking about before, or I'd have explained better.

We're required to invert the image or we won't be able to find shadows.  The Laplacian of Gaussian algorithm isn't capable of finding dark blobs (shadows), so if we don't invert the image, we can't find shadows, and the the whole idea of filtering based on rock + shadow in line with azimuth goes away.

Or maybe I'm still misunderstanding -- are you talking about making the 255 configurable?

autocnet/cg/change_detection.py
```diff
@@ -216,3 +219,135 @@ def okbm_detector(image1, image2, nbins=50, extractor_method="orb",  image_func=
      return polys, weights, bdiff
 
 
+def blob_detector(image1, image2, sub_solar_azimuth, image_func=image_diff, max_sigma=30, num_sigma=10,
+        threshold=.075, n_neighbors=3, dist_upper_bound=5, angle_tolerance=10):
+     """
+
+     Parameters
+     ----------
```
Kelvinrr, line 228: Dont' forget a description of the algorithm and cite the paper we used, even if we didn't use most of it I think we should credit it.

autocnet/cg/change_detection.py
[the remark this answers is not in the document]
Kelvinrr: Making the 255 configurable doesn't make sense since I'm sure the bytescale call should make the diff image have a range of [-255, 255]. I was thinking more about the difference between looking for additive vs subtractive change as depending what kind of change you are trying to detect you may or may not want to inverse (i.e. if looking for a new shadow, the shadow would be a high value in the difference image, the opposite if you are looking for a shadow that disappeared). Assuming I am not wrong in thinking that, whether or not to inverse might be a relatively clean way to toggle what kind of change you are looking for.

Kelvinrr:
once checks finish I think this is good to merge.

PR #459, opened 2020-07-09
Added change detection bin script 

Kelvinrr:


bin/acn_cd.py
```diff
@@ -0,0 +1,110 @@
...
+        exit()
+
+    before_proj_geo = GeoDataset(before_proj)
+    after_proj_geo = GeoDataset(after_proj)
+    ret = _cd_functions_[args.method](before_proj_geo, after_proj_geo, **config[args.method])
+
+    # for now, write out raster files assuming okb
+    # make it match one of the projected images
+    match_srs = before_proj_geo.dataset.GetProjection()
+    match_gt = before_proj_geo.geotransform
```
Kelvinrr, line 104: gonna update this after homogenizing the returns. Something to talk about tomorrow AM.

autocnet/cg/change_detection.py
```diff
@@ -92,6 +101,12 @@ def okubogar_detector(image1, image2, nbins=50, extractor_method="orb", image_fu
      if isinstance(image2, GeoDataset):
          image2 = image2.read_array()
 
+     if isinstance(image_func, str):
+         try:
+             image_func = func_map[image_func]
+         except Exception as e:
```
AustinSanders, line 107: Are we expecting anything other than KeyError here?

bin/acn_cd.py
```diff
@@ -0,0 +1,110 @@
...
+from autocnet.utils.utils import bytescale
+
+import numpy as np
+
+import pysis
+from pysis.exceptions import ProcessError
+from pysis import isis
+
+import warnings
+warnings.simplefilter("ignore")
```
AustinSanders, line 28: Which warnings are we squelching?

Kelvinrr: Pandas warnings about df copies when Autocnet is generating matches. They get obnoxious after a while and they are not describing real problems.

autocnet/cg/change_detection.py
Kelvinrr: no, I can make it more specific, but I don't think it matters in this case.

PR #460, opened 2020-07-09
Adding cost kwargs to propagate_control_network plus bug fix

ladoramkershner:
This allows us to select who many/which points to propagate from one network to another. This is particularly helpful when grounding a network. The initial propagate_control_network call can propagate a single point corresponding to the best match (as defined by geom_match output template metric) between a control network existing in the grounding data to the control data. Then the second propagate_control_network call can propagate that best point (now in the control data image space) to the rest of the overlapping control data images.

During this code change I also added doc strings to the propagation functions and reconciled the use of ographic/ocentric latitudes throughout the code. Autocnet now expects ocentric latitude inputs from the DB geometry columns.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -50,7 +50,7 @@ from autocnet.vis.graph_view import plot_graph, cluster_plot
 from autocnet.control import control
 from autocnet.spatial.overlap import compute_overlaps_sql
 from autocnet.spatial.isis import point_info
-from autocnet.transformation.spatial import reproject
+from autocnet.transformation.spatial import reproject, og2oc
 
 #np.warnings.filterwarnings('ignore')
 
@@ -1392,7 +1392,7 @@ class NetworkCandidateGraph(CandidateGraph):
         # is the best solution I think. I don't want to pass the DEM around
         # for the sensor calls.
         self._setup_dem()
-    
+
     @contextmanager
     def session_scope(self):
      """
@@ -1664,9 +1664,9 @@ class NetworkCandidateGraph(CandidateGraph):
         if msg['success'] == True:
             return
 
-    def to_isis(self, 
-                path, 
-                flistpath=None, 
+    def to_isis(self,
+                path,
+                flistpath=None,
                 latsigma=10,
                 lonsigma=10,
                 radsigma=15,
@@ -1680,9 +1680,9 @@ class NetworkCandidateGraph(CandidateGraph):
                Outpath to write the control network
 
         flishpath : str
-                    Outpath to write the associated file list. If None (default), 
+                    Outpath to write the associated file list. If None (default),
                     the file list is written alongside the control network
-        
+
         latsigma : int/float
                The estimated sigma (error) in the latitude direction
 
@@ -1701,16 +1701,16 @@ class NetworkCandidateGraph(CandidateGraph):
         Returns
         -------
         None
-        
-        """        
+
+        """
         # Read the cnet from the db
         df = io_controlnetwork.db_to_df(self.engine, **db_kwargs)
-        
+
         # Add the covariance matrices to ground measures
-        df = control.compute_covariance(df, 
-                                        latsigma, 
-                                        lonsigma, 
-                                        radsigma, 
+        df = control.compute_covariance(df,
+                                        latsigma,
+                                        lonsigma,
+                                        radsigma,
                                         self.config['spatial']['semimajor_rad'])
 
         if flistpath is None:
@@ -1801,7 +1801,7 @@ class NetworkCandidateGraph(CandidateGraph):
 
         obj = cls.from_database()
         # Execute the computation to compute overlapping geometries
-        obj._execute_sql(compute_overlaps_sql) 
+        obj._execute_sql(compute_overlaps_sql)
         return obj
 
     def copy_images(self, newdir):
@@ -2020,7 +2020,8 @@ class NetworkCandidateGraph(CandidateGraph):
 
             row = cnetpoint.iloc[0]
             x,y,z= row.adjustedX, row.adjustedY, row.adjustedZ
-            lon, lat, alt = reproject([x, y, z], semi_major, semi_minor, 'geocent', 'latlon')
+            lon_og, lat_og, alt = reproject([x, y, z], semi_major, semi_minor, 'geocent', 'latlon')
+            lon, lat = og2oc(lon_og, lat_og, semi_major, semi_minor)
 
 
             point = Points(identifier=id,
@@ -2043,7 +2044,7 @@ class NetworkCandidateGraph(CandidateGraph):
         networkobj = cls.from_filelist(filelist)
         networkobj.place_points_from_cnet(cnet)
         return networkobj
-    
+
     @property
     def measures(self):
         df = pd.read_sql_table('measures', con=self.engine)
--- a/autocnet/io/db/controlnetwork.py
+++ b/autocnet/io/db/controlnetwork.py
@@ -82,6 +82,6 @@ ORDER BY measures."pointid", measures."id";
                     row['adjustedX'] = adjusted_geom.x
                     row['adjustedY'] = adjusted_geom.y
                     row['adjustedZ'] = adjusted_geom.z
-            df[i] = row
-            
+                df.iloc[i] = row
+
         return df
--- a/autocnet/matcher/cross_instrument_matcher.py
+++ b/autocnet/matcher/cross_instrument_matcher.py
@@ -194,7 +194,7 @@ def propagate_point(Session,
                 longitude of point you want to project
 
     lat       : np.float
-                latitude of point you want to project
+                planetocentric latitude of point you want to project
 
     pointid   : int
                 clerical input used to trace point from generate_ground_points output
@@ -251,26 +251,16 @@ def propagate_point(Session,
     p = Point(lon, lat)
     new_measures = []
 
-    print("point id: ", pointid)
-    print("number of CTX images: ", len(images))
-    print("images: ", images)
-    print("lines: ", lines)
-    print("samples: ", samples)
-    print("paths: ", paths)
-    print()
-
     # list of matching results in the format:
     # [measure_index, x_offset, y_offset, offset_magnitude]
     match_results = []
     # lazily iterate for now
     for k,m in image_measures.iterrows():
-        print('base source: ', m['path'])
         base_image = GeoDataset(m["path"])
 
         sx, sy = m["sample"], m["line"]
 
         for i,image in images.iterrows():
-            print('destination source: ', image['path'])
             dest_image = GeoDataset(image["path"])
 
             if os.path.basename(m['path']) == os.path.basename(image['path']):
@@ -282,19 +272,14 @@ def propagate_point(Session,
                         template_kwargs=template_kwargs, \
                         verbose=verbose)
             except Exception as e:
-                print('RAISING EXCEPTION')
                 raise Exception(e)
                 match_results.append(e)
                 continue
 
             if len(images) > 1:
-                print('append to match_results')
-                print('shit to append: ', [k, x, y, metrics, dist, m['path'], image['path']])
             match_results.append([k, x, y,
                                      metrics, dist, corrmap, m["path"], image["path"],
                                      image['id'], image['serial']])
-            if len(images) > 1:
-                print('match_results length: ', len(match_results))
 
     # get best offsets
     match_results = np.asarray([res for res in match_results if isinstance(res, list) and all(r is not None for r in res)])
@@ -309,8 +294,6 @@ def propagate_point(Session,
         # no matches satisfying cost
         return new_measures
 
-    print('best_results: ', best_results)
-    print('\n')
     if verbose:
         print("Full results: ", best_results)
         print("Winning CORRs: ", best_results[:,3], "Themis Pixel shifts: ", best_results[:,4])
@@ -455,8 +438,6 @@ def propagate_control_network(Session,
         lon = point.point.x
         lat = point.point.y
         res = session.execute(f"SELECT * FROM points WHERE ST_Intersects(geom, ST_Buffer(ST_SetSRID(ST_Point({lon}, {lat}), {config['spatial']['latitudinal_srid']}), 1e-10))").fetchall()
-        print(f'lon: {lon}, lat: {lat}')
-        print('length of result: ', len(res))
 
         if len(res) > 1:
             warnings.warn(f"There is more than one point at lon: {lon}, lat: {lat}")
--- a/autocnet/matcher/cross_instrument_matcher.py
+++ b/autocnet/matcher/cross_instrument_matcher.py
@@ -172,7 +172,7 @@ def propagate_point(Session,
                     cost=lambda x, y: y == np.max(x)):
     """
     Conditionally propagate a point into a stack of images. The point and all corresponding measures
-    are matched against database netwrok (to which you are propagating), best result(s) is(are) kept.
+    are matched against database network (to which you are propagating), best result(s) is(are) kept.
 
     Parameters
     ----------
@@ -209,17 +209,19 @@ def propagate_point(Session,
                 apriori sample(s) corresponding to point projected in 'paths' image(s)
 
     size_x    : int
-                size of GeoDataset used in geom_match, must be larger than template_kwargs 'image_size'
+                half width of GeoDataset that is cut from full image and affinely transfromed in geom_match;
+                must be larger than 1/2 template_kwargs 'image_size'
 
     size_y    : int
-                size of GeoDataset used in geom_match, must be larger than template_kwargs 'image_size'
+                half height of GeoDataset that is cut from full image and affinely transfromed in geom_match;
+                must be larger than 1/2 template_kwargs 'image_size'
 
     template_kwargs : dict
                       kwargs passed through to control matcher.subpixel_template()
 
     verbose   : boolean
                 If True, this will print out the results of each propagation, including prints of the
-                matcher areas and their correaltion map.
+                matcher areas and their correlation map.
 
     cost      : anonymous function
                 determines to which image(s) the point should be propagated. x corresponds to a list
@@ -228,12 +230,12 @@ def propagate_point(Session,
                 Example:
                 cost = lambda x,y: y == np.max(x) will get you one result corresponding to the image that
                 has the maximum correlation with the source image
-                cost = lambda x,y: y > 0.6 will propegate the point to all images whose correlation
+                cost = lambda x,y: y > 0.6 will propagate the point to all images whose correlation
                 result is greater than 0.6
 
 
-    Output
-    ----------
+    Returns
+    -------
     new_measures : pd.DataFrame
                    Dataframe containing pointid, imageid, image serial number, line, sample, and ground location (both latlon
                    and cartesian) of successfully propagated points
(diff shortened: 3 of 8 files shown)


autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -206,62 +276,81 @@ def propagate_point(Session,
...
-                'point_ground' : Point(x*1000, y*1000, z*1000)
-        })
+            'pointid' : pointid,
+            'imageid' : row[8],
+            'serial' : row[9],
+            'path': row[7],
+            'line' : line,
+            'sample' : sample,
+            'point' : p,
+            'point_ecef' : Point(x, y, z)
```
ladoramkershner, line 332: The multiplication by 1000 was causing the latitude to shift slightly when it was converted back to lon/lat in model.py

autocnet/graph/network.py
```diff
@@ -2020,7 +2020,8 @@ def get_measures(row):
 
             row = cnetpoint.iloc[0]
             x,y,z= row.adjustedX, row.adjustedY, row.adjustedZ
-            lon, lat, alt = reproject([x, y, z], semi_major, semi_minor, 'geocent', 'latlon')
+            lon_og, lat_og, alt = reproject([x, y, z], semi_major, semi_minor, 'geocent', 'latlon')
+            lon, lat = og2oc(lon_og, lat_og, semi_major, semi_minor)
```
jlaura, line 2024: Because ISIS is expecting OC? Makes total sense, I just want to confirm.

autocnet/io/db/controlnetwork.py
```diff
@@ -82,6 +82,6 @@ def db_to_df(engine, sql = """
                     row['adjustedX'] = adjusted_geom.x
                     row['adjustedY'] = adjusted_geom.y
                     row['adjustedZ'] = adjusted_geom.z
-            df[i] = row
-            
-        return df
\ No newline at end of file
+                df.iloc[i] = row
```
jlaura, line 85: :+1:

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -168,8 +168,75 @@ def propagate_point(Session,
                     size_x=40,
                     size_y=40,
                     template_kwargs={'image_size': (39, 39), 'template_size': (21, 21)},
-                    verbose=False):
+                    verbose=False,
+                    cost=lambda x, y: y == np.max(x)):
     """
+    Conditionally propagate a point into a stack of images. The point and all corresponding measures
+    are matched against database netwrok (to which you are propagating), best result(s) is(are) kept.
```
jlaura, line 175: /s/netwrok/network

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -168,8 +168,75 @@ def propagate_point(Session,
...
+    config    : dict
+                configuation file associated with database you want to propagate to
+                In the form: {'username':'somename',
+                              'password':'somepassword',
+                              'host':'somehost',
+                              'pgbouncer_port':6543,
+                              'name':'somename'}
+
+    dem       : plio.io.io_gdal.GeoDataset
+                digital elevation model of target body
```
jlaura, line 191: not now but something to think about for the future, these first three args are on the ncg - does the ncg just get passed?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -168,8 +168,75 @@ def propagate_point(Session,
...
+                apriori line(s) corresponding to point projected in 'paths' image(s)
+
+    samples   : list of np.float
+                apriori sample(s) corresponding to point projected in 'paths' image(s)
+
+    size_x    : int
+                size of GeoDataset used in geom_match, must be larger than template_kwargs 'image_size'
+
+    size_y    : int
+                size of GeoDataset used in geom_match, must be larger than template_kwargs 'image_size'
```
jlaura, line 215: Does this also need to be odd?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -168,8 +168,75 @@ def propagate_point(Session,
...
+
+    template_kwargs : dict
+                      kwargs passed through to control matcher.subpixel_template()
+
+    verbose   : boolean
+                If True, this will print out the results of each propagation, including prints of the
+                matcher areas and their correaltion map.
```
jlaura, line 222: s/correaltion/correlation

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -168,8 +168,75 @@ def propagate_point(Session,
...
+
+    cost      : anonymous function
+                determines to which image(s) the point should be propagated. x corresponds to a list
+                of all match correlation metrics, while y corresponds to each indiviudal element
+                of the x array.
+                Example:
```
jlaura, line 228: I don't know if the docs will build with the indentation/example like this

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -168,8 +168,75 @@ def propagate_point(Session,
...
+                cost = lambda x,y: y == np.max(x) will get you one result corresponding to the image that
+                has the maximum correlation with the source image
+                cost = lambda x,y: y > 0.6 will propegate the point to all images whose correlation
+                result is greater than 0.6
+
+
+    Output
```
jlaura, line 235: /s/Output/Returns

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -207,61 +277,79 @@ def propagate_point(Session,
...
-        semi_major = config['spatial']['semimajor_rad']
-        semi_minor = config['spatial']['semiminor_rad']
-        # The CSM conversion makes the LLA/ECEF conversion explicit
-        x, y, z = reproject([lon, lat, height],
+                                 metrics, dist, corrmap, m["path"], image["path"],
+                                 image['id'], image['serial']])
+
+    # get best offsets
+    match_results = np.asarray([res for res in match_results if isinstance(res, list) and all(r is not None for r in res)])
+    print("match_results final length: ", len(match_results))
```
jlaura, line 285: @Kelvinrr Checkout this list comprehension! :wink:

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -207,61 +277,79 @@ def propagate_point(Session,
...
+    if match_results.shape[0] == 0:
+        # no matches
+        return new_measures
+
+    best_results = np.asarray([match for match in match_results if cost(match_results[:,3], match[3])])
```
jlaura, line 290: Do you want to document the [:,3]. I know it is metrics and is right above, but a 

`# column index 3 is the metric returned by the geom matcher`

comment might be good. Something to help unravel this list comprehension since it is also indexing into a complex data structure.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -207,61 +277,79 @@ def propagate_point(Session,
...
+    print("best_results length: ", len(best_results))
```
jlaura, line 291: This should be in the verbose block?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -207,61 +277,79 @@ def propagate_point(Session,
...
+
+    if verbose:
+        print("Full results: ", best_results)
+        print("Winning CORRs: ", best_results[:,3], "Themis Pixel shifts: ", best_results[:,4])
+        print("Themis Images: ", best_results[:,6], "CTX images:", best_results[:,7])
+        print("Themis Sample: ", sx, "CTX Samples: ", best_results[:,1])
+        print("Themis Line: ", sy, "CTX Lines: ", best_results[:,2])
+        print('\n')
+
+    if len(best_results[:,3])==1 and best_results[:,3][0] is None:
```
jlaura, line 304: Add a comment here (?) because I have to go back up, unravel the list comprehension and then try to figure out what these indices are.

autocnet/matcher/cross_instrument_matcher.py
jlaura: Are we maintaining the correct units still?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -207,61 +277,79 @@ def propagate_point(Session,
...
+        config,
+        dem,
+        base_cnet,
+        size_x=40,
+        size_y=40,
+        template_kwargs={'image_size': (39,39), 'template_size': (21,21)},
+        verbose=False,
+        cost=lambda x,y: y == np.max(x)):
     """
+    Propagate a contorl network made of line, sample, image path to a network contained within
```
jlaura, line 346: /s/contorl/control

jlaura: Also, the function is called propagate and the doc string says it propagates. But what does that mean? What does the function do?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -276,16 +364,28 @@ def propagate_control_network(Session, config, dem, base_cnet,
...
+                the measure and the 'path' to the corresponding image
 
     verbose   : boolean
                 Increase the level of print outs/plots recieved during propagation
 
+    cost      : anonymous function
+                determines to which image(s) the point should be propagated. x corresponds to a list
+                of all match correlation metrics, while y corresponds to each indiviudal element
+                of the x array.
+                Example:
```
jlaura, line 377: Same example question as above about formatting

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -317,37 +417,67 @@ def propagate_control_network(Session, config, dem, base_cnet,
                                       size_x,
                                       size_y,
                                       template_kwargs,
-                                      verbose=verbose)
+                                      verbose=verbose,
+                                      cost=cost)
+        if len(gp_measures) == 0:
+            continue
```
jlaura, line 423: Why have a check that then does nothing?

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -317,37 +417,67 @@ def propagate_control_network(Session, config, dem, base_cnet,
...
 
-    ground = gpd.GeoDataFrame.from_dict(constrained_net).set_geometry('point_latlon')
+    ground = gpd.GeoDataFrame.from_dict(constrained_net).set_geometry('point')
     groundpoints = ground.groupby('pointid').groups
 
     points = []
 
-    # upload new points
+    # conditionally upload a new point to DB or updated existing point with new measures
+    session = Session()
```
jlaura, line 433: Something to think about - most (all?) of the rest of the codebase uses the session context manager on an ncg. That might be cleaner.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -317,37 +417,67 @@ def propagate_control_network(Session, config, dem, base_cnet,
...
+
+        if len(res) > 1:
+            warnings.warn(f"There is more than one point at lon: {lon}, lat: {lat}")
+
+        elif len(res) == 1:
+            for i in indices:
+                row = ground.loc[i]
+                pid = res[0][0]
+                meas = session.execute(f"SELECT serialnumber FROM measures WHERE pointid={pid}").fetchall()
+                serialnumbers = [m[0] for m in meas]
```
jlaura, line 449: What about: `serialnumbers = session.query(Measures.serialnumber).filter(Measures.pointid == pid).all()`

This uses the ORM so that we don't have raw SQL in the codebase and so that we can possibly make changes to the database without impacting this code.

autocnet/transformation/tests/test_spatial.py
```diff
@@ -3,6 +3,19 @@
...
+    lon = 0
+    lat = 20
+    lon_og, lat_og = spatial.oc2og(lon, lat, 3396190, 3376200)
+    assert lat_og == 20.218400434636393
+
+def test_og2oc():
+    lon = 0
+    lat = 20
+    lon_oc, lat_oc = spatial.oc2og(lon, lat, 3396190, 3376200)
+    assert lat_oc == 19.78356596059272
```
jlaura, line 16: Does it make sense to test 0 or 90 as well so that we are very comfortable that we have parametrized the proj pipeline properly.

jlaura:
I think that the test failures for the wkb geom tests are because of the change you made in model.py.

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: @Kelvinrr wrote most of it, I just added another condition to the if (and the addition was actually Kelvin's suggestion!). He is still king of the (comprehension) rock.

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: I believe so, because the semi_major and semi_minor radius are the same order of magnitude as the x, y, z directly out of reproject (line316)

autocnet/graph/network.py
ladoramkershner: I am not sure what EVERY function of ISIS expects, but I know that campt/mappt expects ocentric latitudes

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: That does seem like a really good idea!
I may have time to do that while mosaics are cooking

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: I don't think so, 'image_size' and 'template_size' arguments on subpixel_template need to be odd because of the assumption subpixel_template (and then subsequent roi.Roi.clip) that those values are the FULL width of the chips. However, these keywords get passed to geom_match which uses it as the half width of the area it cuts from the full image, preforms an affine transform on and then passes to subpixel_template. It is never divide in half and then used for pixel distance, so I think we are okay!

I am glad you brought this up, this distinction is confusing and I was lazy and did not document it well here. Additionally, there is no doc string on geom_match, so I had to go digging for this information (hence why I lazily did not do this digging/documenting before).

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: The docs built fine when I do
```
from autocnet.matcher.cross_instrument_matcher import propagate_control_network
propagate_control_network?
```
Is this what you were referencing?

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: 👍

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: 👍

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: 👍

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: 👍

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: 👍

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: 👍

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: Fair point

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: This check keeps us from appending empty results to constrained_net. If propagate_point either cannot find a match, or none of the matchs fit the 'best' criteria, it returns an empty list

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: I think this would go hand in hand with passing in the ncg to the function instead of Session, config, and dem. These function were written before the big refactor you did and as such does not currently have access to the ncg session context manager

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: Can do! I should also probably figure something out for that gnarly raw SQL about 10 lines up

autocnet/matcher/cross_instrument_matcher.py
jlaura: with a continue or it needs a return?

autocnet/matcher/cross_instrument_matcher.py
jlaura: :+1:

autocnet/matcher/cross_instrument_matcher.py
jlaura: That one is going to be a little harder which is why I only commented here. :)

autocnet/transformation/tests/test_spatial.py
ladoramkershner: I am not sure what you mean. 90 degrees lat is the same in both systems.

ladoramkershner: I do agree that these tests are not sufficient, I am also not even sure if these tests are good because I should really be hand calculating the conversion and then making sure the function output and the 'hand calcs' are the same.

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: continue should work because it is in a for loop and there is one more line in the for loop, which is the one we want to skip

autocnet/matcher/cross_instrument_matcher.py
ladoramkershner: I think I got it!

autocnet/matcher/cross_instrument_matcher.py
jlaura: Nope, I meant with the `Example:` block. But if they are building that is pretty awesome to know!

autocnet/matcher/cross_instrument_matcher.py
jlaura: haha, nope - all good. I missed the 'not changed' code expansion not once, but twice.

autocnet/matcher/cross_instrument_matcher.py
jlaura: :+1:

autocnet/transformation/tests/test_spatial.py
jlaura: I would use some of the code from Trent to do the math 'by hand'. Specifically:

```
import math

def oc2og(dlat, dMajorRadius = 3396190.0, dMinorRadius = 3376200.0):
    try:
        dlat = math.radians(dlat)
        dlat = math.atan(((dMajorRadius / dMinorRadius)**2) * (math.tan(dlat)))
        dlat = math.degrees(dlat)
    except:
        print ("Error in oc2og conversion")
    return dlat

def og2oc(dlat, dMajorRadius = 3396190.0, dMinorRadius = 3376200.0):
    try:
        dlat = math.radians(dlat)
        dlat = math.atan((math.tan(dlat) / ((dMajorRadius / dMinorRadius)**2)))
        dlat = math.degrees(dlat)
    except:
        print ("Error in oc2og conversion")
    return dlat
```

The math is so simple that I think we can trust this and use it to test our implementation. Thoughts?

ladoramkershner: I think that is perfect! I just wanted something more than what I had which was passing because it was essentially self confirming. I will add that in.

thareUSGS: I need to check the licensing on that 15 year-old code... :-)

PR #462, opened 2020-07-10
Adjusted parameters and added image_diff_squared 

AustinSanders:
Adjusted parameters based on results from parameterization comparison notebook (which is too large to create a gist) and added an additional image func.

### DIFF
--- a/autocnet/cg/change_detection.py
+++ b/autocnet/cg/change_detection.py
@@ -24,6 +24,10 @@ def image_diff(arr1, arr2):
      return diff
 
 
+def image_diff_sq(arr1, arr2):
+     return image_diff(arr1, arr2)**2
+
+
 def okubogar_detector(image1, image2, nbins=50, extractor_method="orb", image_func=image_diff,
                       extractor_kwargs={"nfeatures": 2000, "scaleFactor": 1.1, "nlevels": 1}):
      """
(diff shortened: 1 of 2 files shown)


autocnet/cg/change_detection.py
```diff
@@ -219,9 +223,9 @@ def okbm_detector(image1, image2, nbins=50, extractor_method="orb",  image_func=
      return polys, weights, bdiff
 
 
-def blob_detector(image1, image2, sub_solar_azimuth, image_func=image_diff,
-                  subtractive=False, max_sigma=30, num_sigma=10, threshold=.075,
-                  n_neighbors=3, dist_upper_bound=5, angle_tolerance=10):
+def blob_detector(image1, image2, sub_solar_azimuth, image_func=image_diff_sq,
+                  subtractive=False,  min_sigma=.45, max_sigma=30,  threshold=.25,
+                  n_neighbors=3, dist_upper_bound=5, angle_tolerance=3):
```
Kelvinrr, line 228: When creating the defaults for the config I am going to base it off these defaults, seem reasonable?

autocnet/cg/change_detection.py
```diff
@@ -272,16 +276,19 @@ def blob_detector(image1, image2, sub_solar_azimuth, image_func=image_diff,
                    words, find locations in which a feature "used to be present"
                    but has since moved.
 
+     min_sigma : scalar or sequence of scalars
+                 The minimum standard deviation for Gaussian kernel. Keep this
+                 low to detect smaller blobs. The standard deviations of the
+                 Gaussian filter are given for each axis as a sequence, or as a
+                 single number, in which case it is equal for all axes.
+
+
```
Kelvinrr, line 285: Might be worth exposing more blob args through additional kwargs, easy enough to do and makes the function a bit more powerful

autocnet/cg/change_detection.py
AustinSanders: I think these are all reasonable values, but I only tested them on the 2 example images

autocnet/cg/change_detection.py
AustinSanders: I considered that too... I'm just worried that it's already a very bloated signature.  Should we just go for it?  I can default them to whatever the skimage defaults are.

Kelvinrr: I'll approve it for now since this function is probably gonna change and the more verbose params are a good thing in this case, but typically you just use kwargs and pass them through to the function, see the other change detection algorithms.

PR #464, opened 2020-07-20
typo, 'conda condig' should be 'conda config'

mol3earth:
bigger issue is that
`> conda install -c jlaura autocnet-dev`
responds
`PackagesNotFoundError:`


jessemapel:
@jlaura Is there a reason we're not pushing this up to conda anywhere?

jlaura:
@jessemapel It should be deploying (.travis.yml calls the deploy.sh script). So this is a bug that should be fixed.

PR #466, opened 2020-07-22
typo in conda install command

mol3earth:
Regardless of the typo fix, these instructions do not seem to work.

upon following steps 2, 3, and 4 (`conda install -c usgs-astrogeology autocnet`)
I am unable to succesfully use autocnet. 

`> from autocnet.graph.network import NetworkCandidateGraph`  
returns this error  
```
~/anaconda3/envs/autocnet/lib/python3.7/site-packages/setuptools/distutils_patch.py:26: UserWarning: Distutils was imported before Setuptools. This usage is discouraged and may exhibit undesirable behaviors or errors. Please use Setuptools' objects directly or at least import Setuptools first.
  "Distutils was imported before Setuptools. This usage is discouraged "
~/anaconda3/envs/autocnet/lib/python3.7/site-packages/pysis/env.py:31: RuntimeWarning: Could not find isis. Is `ISISROOT` set?
  warnings.warn('Could not find isis. Is `ISISROOT` set?', RuntimeWarning)
```

What is ISISROOT? Is [the isis package](https://github.com/USGS-Astrogeology/ISIS3#installation) a dependency that I need to download and configure? I can't find the referenced script to configure ISISROOT.

thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ We suggest using Anaconda Python to install Autocnet within a virtual environmen
 1. [Download](https://www.continuum.io/downloads) and install the Python 3.x Miniconda installer.  Respond ``Yes`` when prompted to add conda to your BASH profile.  
 2. Install the autocnet environment using the supplied environment.yml file: `conda env create -n autocnet -f environment.yml` 
 3. Activate your environment: `conda activate autocnet`
-4. If you are doing to develop autocnet or would like to use the bleeding edge version: `python setup.py develop`. Otherwise, `conda install -c usgs-astrogeology` autocnet.
+4. If you are doing to develop autocnet or would like to use the bleeding edge version: `python setup.py develop`. Otherwise, `conda install -c usgs-astrogeology autocnet`
 
 ## How to run the test suite locally
 


Kelvinrr:
ISIS is an optional dependency, we should probably silence that warning unless you explicitly call the functions it's needed for. That first message is a warning that has popped up recently, not sure what is causing it, but it seems like an upstream change we need to handle. In either case, these are warnings not errors. That import seems to work fine assuming those warnings were not accompanied by a stack-trace. 

The ISIS Warning is alluding to the fact that errors will happen if you try to use the functions in [isis.py](https://github.com/USGS-Astrogeology/autocnet/blob/dev/autocnet/spatial/isis.py) down the line.

mol3earth:
thanks for the reply. You're correct, it does seem that autocnet works despite the warnings.
Acutally, I am trying to install autocnet in order to use [your branch of bayleef](https://github.com/Kelvinrr/bayleef) which does appear to require isis as I am getting some import errors referencing it. but this is probably not the correct place to discuss that repo. ive tried installing isis a few times and continually get conflict errors.

jessemapel:
@mol3earth ISIS has some very restrictive environment requirements that make it not play nicely with many packages. Since, this only uses it through pysis, you can actually install ISIS into a different environment and then just set your ISISROOT to that environment and pysis will be able to use it.

Kelvinrr:
> thanks for the reply. You're correct, it does seem that autocnet works despite the warnings.
> Acutally, I am trying to install autocnet in order to use [your branch of bayleef](https://github.com/Kelvinrr/bayleef) which does appear to require isis as I am getting some import errors referencing it. but this is probably not the correct place to discuss that repo. ive tried installing isis a few times and continually get conflict errors.

Yeah, I am currently working in getting an up-to-date deployment of Autocnet to make that process easier, as you currently need to do a manual dev build of Autocnet in order to get things working (as you can see, it can be a pain sometimes). This repo also sees high velocity changes which make stability an issue so we need to tag and deploy releases. 

For your case, you will need an ISIS install. As Jesse suggested,  installing ISIS in a different environment and then setting ISISROOT to that install will be easier as ISIS has some very specific dependencies that currently make it difficult to co-install with other big libraries (e.g GDAL). Also see https://github.com/USGS-Astrogeology/ISIS3#installation on how to get ISISDATA set up.

Kelvinrr:
Im going to merge this, as it only about the typo. If you run into other problems feel free to open an issue here or on Bayleef, whatever is more appropriate.

mol3earth:
thanks for the info. I'll head over to the bayleef repo and start an issue for install instructions as they don't seem all that relevant here.

PR #468, opened 2020-07-28
Updating travis to build on MacOS

Kelvinrr:
Might get loud

### DIFF
--- a/autocnet/io/db/tests/test_model.py
+++ b/autocnet/io/db/tests/test_model.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 import json
+import sys
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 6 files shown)


autocnet/io/db/tests/test_model.py
```diff
@@ -12,6 +13,9 @@
 
 from shapely.geometry import MultiPolygon, Polygon, Point
 
+if sys.platform.startswith("darwin"):
+    pytest.skip("skipping DB tests for MacOS", allow_module_level=True)
+
```
Kelvinrr, line 18: Basically ignore the DB related tests for now on MacOS. Originally attempted a approach of using our docker-compose file to stand up postgres since services are not supported on MacOS, but that turned into a hot mess really fast. 

An alternative would be to only run Linux tests like before but then use [conda build convert](https://docs.conda.io/projects/conda-build/en/latest/resources/commands/conda-convert.html) in the end.

PR #469, opened 2020-08-03
New visualizations for geom_match

ladoramkershner:
Not sure if this should get pulled into the code base, just posting it for the new visualization code

### DIFF
--- a/autocnet/matcher/subpixel.py
+++ b/autocnet/matcher/subpixel.py
@@ -520,8 +520,16 @@ def geom_match(base_cube,
         raise Exception(f"Window: {base_starty} < 0, center: {bcenter_x},{bcenter_y}")
 
     # specifically not putting this in a try/except, this should never fail
-    mlat, mlon = spatial.isis.image_to_ground(base_cube.file_name, bcenter_x, bcenter_y)
-    center_x, center_y = spatial.isis.ground_to_image(input_cube.file_name, mlon, mlat)[::-1]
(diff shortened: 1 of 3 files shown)


autocnet/matcher/subpixel.py
```diff
@@ -520,8 +520,16 @@ def geom_match(base_cube,
...
+    # 07/28 - putting it in a try/except because of how we ground points
+    try:
+        mlat, mlon = spatial.isis.image_to_ground(base_cube.file_name, bcenter_x, bcenter_y)
+        center_x, center_y = spatial.isis.ground_to_image(input_cube.file_name, mlon, mlat)
+    except ProcessError as e:
+            if 'Requested position does not project in camera model' in e.stderr:
+                print(f'Skip geom_match; Region of interest center located at ({mlon}, {mlat}) does not project to image {input_cube.base_name}')
+                print('This should only appear when propagating ground points')
+                return None, None, None, None, None
+
```
ladoramkershner, line 532: When we ground a point we select a lon, lat and feature match the THEMIS feature at that lon, lat up to CTX. We hold onto the THEMIS lon, lat but (because of the feature matching) we massage the sample/line in CTX from its original projected value. 

That means this block of code can fail during the second propagate_control_network call (when we propagate the best THEMIS to CTX match to all other overlapping CTX images) because a:
a.) overlapping images are found through a spatial query using the lon, lat
b.) a points lon, lat correspond to an adjusted sample/line

PR #473, opened 2020-08-13
Session cleanup

jlaura:
These changes ensure that the sessions are being closed properly. I tested this via a notebook. I need to think more about if/how this could be CI tested. Not sure it can be without some pretty significant mocking.

Here is the test I ran:

1. Create an NCG with 2 images
2. Get the current count of number of connections to the db using: 

```
def check_connection_count(ncg):
    with ncg.session_scope() as session:
        num_con = session.execute('SELECT sum(numbackends) FROM pg_stat_database;').scalar()
        #num_con = session.execute('select count(*) from pg_stat_activity;').scalar()

    return num_con
```

3. Confirm that the count is either 1 or 1+any pgadmin connections. To do this cleanly, I stood up a test_db container on smalls so that nothing was colliding with other people working.
4. Print the number of connections, run place points in overlap, and then run the following while place points is happening on the cluster:

```
for i in range(120):
    nconn = check_connection_count(ncgA)
    print(nconn)
    sleep(0.5)
```

results were as expected. One job was running so we had one additional connection.

5. This resulted in 22 points. Then I sub pixel registered these and saw the following:

```
3
3
3
3
3
6
25
25
25
25
25
25
25
25
25
25
25
17
16
3
3
3
3
3
3
```

Since I had pgadmin with one query and the initial NCG sessions live I was expecting to see 25 total connections at the peak. So it looks like the sessions are properly expunging. The repeated 25s (at 0.5 second) intervals look to me like we do not have any additional spurious sessions being created inside of the function call or dangling from acn_submit.

### DIFF
--- a/bin/acn_submit
+++ b/bin/acn_submit
@@ -43,17 +43,21 @@ def _instantiate_row(msg, ncg):
     # Get the dict mapping iterable keyword types to the objects
     objdict = ncg.apply_iterable_options
     rowid = msg['id']
-
-    session = ncg.Session()
     obj = objdict[msg['along']]
-    res = session.query(obj).filter(getattr(obj, 'id')==msg['id']).one()
-    session.close()
+    with ncg.session_scope() as session:
+        res = session.query(obj).filter(getattr(obj, 'id')==msg['id']).one()
+        print(session.execute('SELECT sum(numbackends) FROM pg_stat_database;').scalar())
+        print('Done instantiating the obj')
+        session.expunge(res) # Disconnect the object from the session
     return res
 
 def main(msg):
 
     ncg = NetworkCandidateGraph()
     ncg.config_from_dict(msg['config'])
+    with ncg.session_scope() as session:
+        print(session.execute('SELECT sum(numbackends) FROM pg_stat_database;').scalar())
+    print('Done with backened sum')
 
     if msg['along'] in ['node', 'edge']:
         obj = _instantiate_obj(msg, ncg)
@@ -67,6 +71,8 @@ def main(msg):
     if hasattr(obj, msg['func']): 
         func = getattr(obj, msg['func'])
     else:
+        # In this case, obj is just the integer ID. This is only called for database
+        # related objects.
         func = import_func(msg['func'])
         # Get the object begin processed prepended into the args.
         msg['args'] = (obj, *msg['args'])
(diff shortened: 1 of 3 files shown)


bin/acn_submit
```diff
@@ -43,17 +43,21 @@ def _instantiate_row(msg, ncg):
...
     rowid = msg['id']
-
-    session = ncg.Session()
     obj = objdict[msg['along']]
-    res = session.query(obj).filter(getattr(obj, 'id')==msg['id']).one()
-    session.close()
+    with ncg.session_scope() as session:
+        res = session.query(obj).filter(getattr(obj, 'id')==msg['id']).one()
+        print(session.execute('SELECT sum(numbackends) FROM pg_stat_database;').scalar())
+        print('Done instantiating the obj')
```
jlaura: I need to remove these...

bin/acn_submit
```diff
@@ -43,17 +43,21 @@ def _instantiate_row(msg, ncg):
...
+        session.expunge(res) # Disconnect the object from the session
     return res
 
 def main(msg):
 
     ncg = NetworkCandidateGraph()
     ncg.config_from_dict(msg['config'])
+    with ncg.session_scope() as session:
+        print(session.execute('SELECT sum(numbackends) FROM pg_stat_database;').scalar())
```
jlaura: and these. whoops.

ladoramkershner:
I am going to play devil's advocate. Why shouldn't we create an example local database that lives in the autocnet repo? The whole point of autocnet it to work on database's, so we have one and alter it after read in in the tests if we want to fabricate a failure, instead of mocking and remocking and remocking. I get the idea of not wanting to bog down the database with memory intensive individual images, but one database should be fine?

That doesn't solve the issue of needing to mock a redis queue for this purpose, but it does prevent us from having to mock session connects (and hoping our mocking properly represents sqlaclemy connections)

jlaura:
The reason not to include a DB here is because we are pretty tightly coupled with PostGreSQL. We have had some small efforts to also support SQLite (which we could push into the repo), but those are 1.5+ years old at this point. So having a DB in the repo currently is not too feasible.

ladoramkershner:
We could make it a future goal to include support for SQLite if that is the most robust solution we can think of and then put off testing this for now...but that 
a.) runs a high risk of disappearing into the ether and then nothing happening. 
b.) risks confounding failures with interfacing with SQLite with failures in the autocnet code base
c.) still does not solve the problem of mocking redis

Issue #478, opened 2020-08-25
Parallelize update_from_jigsaw?


ladoramkershner:
update_from_jigsaw is currently a single iterating query that takes a long time on larger networks. I am not sure the best way to speed it up, but the process is definitely a hindrance at the moment.

I think you could read in the network, split the resulting DF into a user defined number of networks, and then use queries to update portions of the database with the split DFs. Does anyone have objection or alternative suggestions?

PR #479, opened 2020-09-01
Adding reapply() and update_measure_from_jigsaw() function

ladoramkershner:
The reapply function is being added in case an error occurs on the slurm side and all the jobs in the queue do not fire off properly. 

update_measure_from_jigsaw is being added so the cluster can be used to update the database from jigsaw. The current function is serial and takes forever for a full quad. I am not sure if I love the name update_measure_from_jigsaw, but points do not hold any jigsaw information the measures do, but making it plural (measure -> measures) doesn't imply what I would like either. Open to ideas!

Finally, broke up the serial update from jigsaw, currently if the function fails part way through (because of VPN connection timing out after 8 hours) then the whole query is lost. The new version of the function breaks up the jigsaw network and updates in chunks, not the perfect solution, but maybe better?


autocnet/graph/network.py
```diff
@@ -1639,6 +1639,56 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', chunksize=100
         submitter.submit(array='1-{}%{}'.format(job_counter,arraychunk), chunksize=chunksize)
         return job_counter
 
+    def reapply(self, function, job_counter, chunksize=1000, arraychunk=None, walltime="01:00:00"):
```
ladoramkershner, line 1642: I was thinking about how to write a test for this, could we populate a local redis queue and then mock Slurm.submit to be a redis pop, or is that way too over simplified?

Also, I was looking in autocnet/graph/tests/test_network_graph.py to find the ncg.apply() tests for a basis.. there are not any.

autocnet/graph/network.py
```diff
@@ -1639,6 +1639,56 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', chunksize=100
...
+        walltime:    string
+                     In the format "Hour:Minute:Second" (e.g.: "00:30:00" is 30 minutes of walltime)
+        """
+
+        # Submit the left over jobs
+        rconf = self.config['redis']
+        rhost = rconf['host']
+        rport = rconf['port']
+        processing_queue = rconf['processing_queue']
+
```
jlaura, line 1672: self.processing_queue is the attribute on the ncg that is the queue.

autocnet/graph/network.py
```diff
@@ -1639,6 +1639,56 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', chunksize=100
...
+        command = f'{condasetup} && {isissetup} && {job}'
+
+        submitter = Slurm(command,
+                          job_name='AutoCNet',
+                          mem_per_cpu=self.config['cluster']['processing_memory'],
+                          time=walltime,
+                          partition=self.config['cluster']['queue'],
+                          output=self.config['cluster']['cluster_log_dir']+f'/autocnet.{function}-%j')
+        submitter.submit(array='1-{}%{}'.format(job_counter, arraychunk), chunksize=chunksize)
+        return '1-{}%{}'.format(job_counter, arraychunk)
```
jlaura, line 1690: This looks identical to the apply func except the pushing of messages is removed. Is that correct? If so, instead of duplicating this code, why not just offer a 'reapply' or something flag to apply and then wrapping the row push stuff into an if block? By default push the messages, but if some kwarg is True skip that step?

ladoramkershner:
Made to address issue #478

autocnet/graph/network.py
ladoramkershner: If the look at rconf in line 1668, processing_queue is coming from self.config['redis']['processing_queue'], this is also the method that apply uses. I am going to follow your other suggestion and combine apply and reapply with a flag, but is there a more correct way to get processing_queue?

jlaura (approved):

autocnet/graph/network.py
jlaura: 1668? Checkout: https://github.com/USGS-Astrogeology/autocnet/blob/3336915aac4c4c4435a84ec866089b94040aae1b/autocnet/graph/network.py#L1435 This is where the queue gets set as a member variable.

jlaura:
I think this is good to merge whenever you are happy with it.

PR #484, opened 2020-09-16
affine centered on the ROI

jlaura:
This PR alters the subpixel registration to use an affine transform that is centered on the ROI (instead of the upper left corner). I have a notebook that demos how this is working that I would be happy to show via a private gist link. 

I will comment inline if that helps though I tried to comment in the code directly so that the why is longer lived.

### DIFF
--- a/autocnet/matcher/cross_instrument_matcher.py
+++ b/autocnet/matcher/cross_instrument_matcher.py
@@ -267,7 +267,6 @@ def propagate_point(Session,
         sx, sy = m["sample"], m["line"]
 
         for i,image in images.iterrows():
-            print('PATH', image['path'])
             # When grounding to THEMIS the df has a PATH to the QUAD
             dest_image = GeoDataset(image["path"])
 
--- a/autocnet/matcher/cross_instrument_matcher.py
+++ b/autocnet/matcher/cross_instrument_matcher.py
@@ -105,8 +105,12 @@ def generate_ground_points(Session, ground_mosaic, nspts_func=lambda x: int(roun
         # res = ground_session.execute(formated_sql)
         p = Point(*coord)
         print(f'point {i}'),
-
+ 
+        
         linessamples = isis.point_info(ground_mosaic.file_name, p.x, p.y, 'ground')
+        if linessamples is None:
+            print('unable to find point in ground image')
+            continue
         line = linessamples[0].get('Line')
         sample = linessamples[0].get('Sample')
 
@@ -259,10 +263,12 @@ def propagate_point(Session,
     # lazily iterate for now
     for k,m in image_measures.iterrows():
         base_image = GeoDataset(m["path"])
-
+        
         sx, sy = m["sample"], m["line"]
 
         for i,image in images.iterrows():
+            print('PATH', image['path'])
+            # When grounding to THEMIS the df has a PATH to the QUAD
             dest_image = GeoDataset(image["path"])
 
             if os.path.basename(m['path']) == os.path.basename(image['path']):
@@ -273,7 +279,6 @@ def propagate_point(Session,
                 print(f'prop point: dest_image: {dest_image}')
                 print(f'prop point: (sx, sy): ({sx}, {sy})')
                 x,y, dist, metrics, corrmap = geom_match(base_image, dest_image, sx, sy, \
-                        size_x=size_x, size_y=size_y, \
                         template_kwargs=template_kwargs, \
                         verbose=verbose)
             except Exception as e:
@@ -431,7 +436,7 @@ def propagate_control_network(Session,
         if len(gp_measures) == 0:
             continue
         constrained_net.extend(gp_measures)
-
+        
     ground = gpd.GeoDataFrame.from_dict(constrained_net).set_geometry('point')
     groundpoints = ground.groupby('pointid').groups
 
--- a/autocnet/matcher/subpixel.py
+++ b/autocnet/matcher/subpixel.py
@@ -325,6 +325,14 @@ def subpixel_transformed_template(sx, sy, dx, dy,
     s_image = bytescale(s_roi.clip(dtype=s_image_dtype))
     d_template = bytescale(d_roi.clip(dtype=d_template_dtype))
 
+    if verbose:
+        fig, axs = plt.subplots(1, 5, figsize=(20,10))
+        # Plot of the original image and template
+        axs[0].imshow(s_image, cmap='Greys')
+        axs[0].set_title('Destination')
+        axs[1].imshow(d_template, cmap='Greys')
+        axs[1].set_title('Original Source')
+
     # Build the transformation chance
     shift_x, shift_y = d_roi.center
     
@@ -341,18 +349,25 @@ def subpixel_transformed_template(sx, sy, dx, dy,
                                 itrans,
                                 order=3)
 
-    # Scale the CTX arr to the proper size
+    # Scale the source arr to the destination array size
     scale_y, scale_x = transform.scale
     template_shape_y, template_shape_x = d_template.shape
-
     scaled_roi = tf.resize(transformed_roi, (int(template_shape_x/scale_x), int(template_shape_y/scale_x)))
-    d_template = bytescale(scaled_roi)
     
     # Clip the transformed template to avoid no data around around the edges
-    buffered_template = d_template[template_buffer:-template_buffer,template_buffer:-template_buffer]
+    buffered_template = scaled_roi[template_buffer:-template_buffer,template_buffer:-template_buffer]
 
     # Apply the matcher on the transformed array
-    shift_x, shift_y, metrics, corrmap = func(buffered_template, s_image, **kwargs)
+    shift_x, shift_y, metrics, corrmap = func(bytescale(buffered_template), s_image, **kwargs)
+
+    if verbose:
+        axs[2].imshow(transformed_roi, cmap='Greys')
+        axs[2].set_title('Affine Transformed Source')
+        axs[3].imshow(buffered_template, cmap='Greys')
+        axs[3].set_title('Scaled and Buffered Source')
+        axs[4].imshow(corrmap)
+        axs[4].set_title('Correlation')
+        plt.show()
 
     # Project the center into the affine space
     projected_center = itrans(d_roi.center)[0]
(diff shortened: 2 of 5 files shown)


jlaura:
Some debugging print statements remain. I will work to pull those. Wanted to get this up in case anyone wanted to test locally.

autocnet/matcher/subpixel.py
```diff
@@ -600,57 +689,65 @@ def geom_match(base_cube,
         dist = (dist_temp, dist_phase)
         metric = (maxcorr, perror, pdiff)
     else:
-        x,y,maxcorr,temp_corrmap = restemplate
+        x,y,metric,temp_corrmap = restemplate
```
Kelvinrr, line 692: I feel like we might as well remove the phase_kwargs options, still seems half implemented here (e.g. no error checking to make sure subpixel template returns even enable a good phase match).

autocnet/matcher/subpixel.py
```diff
@@ -600,57 +689,65 @@ def geom_match(base_cube,
...
             axs[2].scatter(x, y, s=10, color='red')
             axs[2].plot([0, size_x*2], [y, y], linewidth=0.75, color='red')
             axs[2].plot([x, x], [0, size_y*2], linewidth=0.75, color='red')
-            axs[2].set_xlim([0, dst_arr[2:].shape[0]])
-            axs[2].set_ylim([0, dst_arr[2:].shape[1]])
+            #axs[2].set_xlim([0, dst_arr[2:].shape[0]])
+            #axs[2].set_ylim([0, dst_arr[2:].shape[1]])
 
-            plt.show()
+            plt.show()"""
```
Kelvinrr, line 746: I assume this is being moved to subpixel template?

Kelvinrr (commented):
Can you message me the gist link? probably try it out sometime tomorrow to see how well it works. Is this a draft because you're not confident in the logic or is it simply clean-up still required?

autocnet/matcher/subpixel.py
jlaura: Exactly. I need to add a verbose option pass through into that function. I think that is probably a better place for that vis to happen as it is closer to the actual place where this stuff is used.

autocnet/matcher/subpixel.py
jlaura: I would be down with fully yanking phase. It hasn't been consistently effective.

jlaura:
And still a draft because I would appreciate more eyes on the logic and I want to clean up the prints/debugs/etc.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -105,8 +105,12 @@ def generate_ground_points(Session, ground_mosaic, nspts_func=lambda x: int(roun
         # res = ground_session.execute(formated_sql)
         p = Point(*coord)
         print(f'point {i}'),
-
+ 
+        
         linessamples = isis.point_info(ground_mosaic.file_name, p.x, p.y, 'ground')
+        if linessamples is None:
+            print('unable to find point in ground image')
+            continue
```
ladoramkershner, line 113: I think this should be wrapped in a try/except. point_info is set up to throw exceptions if there is an issue, I am not sure if it can return None. Did you run into None returns?

autocnet/matcher/subpixel.py
```diff
@@ -227,6 +232,7 @@ def subpixel_template(sx, sy, dx, dy,
                       image_size=(251, 251),
                       template_size=(51,51),
                       func=pattern_match,
+                      transform = None,
```
ladoramkershner, line 235: Add a description to function doc string

autocnet/matcher/cross_instrument_matcher.py
jlaura: Yup. It was throwing an error and returning None. So I went with the message.

autocnet/matcher/subpixel.py
ladoramkershner: I will third that motion

jlaura:
I just updated:

1. Removed the phase stuff
1. Cleaned up prints
1. Cleaned up the verbose stuff, located primarily in subpixel. Default is False so that we maintain current functionality.
1. Added doc string entries where they were missing.

If this looks good to people and they have tested to their satisfaction locally, I am comfortable converting the PR from draft status.

ladoramkershner:
I pulled it locally and it looks good, there is just a lot of crap ground getting through because we have no threshold on the original propagation up from THEMIS to CTX. Actually no threshold kwargs in propagate_cnet at all, I believe I was originally against adding that kwargs, but idk why. 

Also, I think we could benefit from having an error similar to ISIS where, if the max correlation value is on the edge of the fit chip, an 'Insufficient Data' error is thrown and that match is thrown out.

jlaura:
@ladoramkershner I can add both of those things.

ladoramkershner:
I think that capability is off topic for this PR, which already rather large. I am going to open a ticket and we can address it in a separate PR down the line.

jlaura:
@ladoramkershner Last CI test is running now and I added tests to increase total test coverage. This should be good for another review whenever.

autocnet/matcher/cross_instrument_matcher.py
```diff
@@ -259,10 +263,12 @@ def propagate_point(Session,
     # lazily iterate for now
     for k,m in image_measures.iterrows():
         base_image = GeoDataset(m["path"])
-
+        
         sx, sy = m["sample"], m["line"]
 
         for i,image in images.iterrows():
+            print('PATH', image['path'])
+            # When grounding to THEMIS the df has a PATH to the QUAD
```
ladoramkershner, line 271: possible print clean up? Not necessary if you really think it is valuable here.

autocnet/matcher/subpixel.py
```diff
@@ -271,26 +430,56 @@ def subpixel_template(sx, sy, dx, dy,
...
 
+    try:
+        s_image_dtype = isis2np_types[pvl.load(s_img.file_name)["IsisCube"]["Core"]["Pixels"]["Type"]]
+    except:
+        s_image_dtype = None
+    
+    try:
+        d_template_dtype = isis2np_types[pvl.load(d_img.file_name)["IsisCube"]["Core"]["Pixels"]["Type"]]
+    except:
+        d_template_dtype = None
```
ladoramkershner, line 458: What happens if this goes through the exception and we try and load s_image and d_template with None dtypes?

autocnet/matcher/subpixel.py
```diff
@@ -271,26 +430,56 @@ def subpixel_template(sx, sy, dx, dy,
...
+    if verbose:
+         fig, axs = plt.subplots(1, 5, figsize=(20,10))
+         axs[0].imshow(s_image, cmap='Greys')
+         axs[1].imshow(d_template, cmap='Greys')
+
+    transformed_roi = d_template
+
+    if verbose:
+        axs[2].imshow(transformed_roi, cmap='Greys')
+        axs[3].imshow(d_template[5:-5,5:-5], cmap='Greys')
```
ladoramkershner, line 472: These are both equal to d_template and that has already been plotted. Are these plots useful?

autocnet/matcher/subpixel.py
jlaura: The default is None once it winds all the way through to plio where it does the read. The reality is that this should end up in plio at some point and be removed from here.

autocnet/matcher/cross_instrument_matcher.py
jlaura: Removed.

jlaura:
@ladoramkershner Updated based on your comments.

autocnet/matcher/subpixel.py
```diff
@@ -546,129 +760,25 @@ def geom_match(base_cube,
...
-            axs[1].imshow(dst_arr_org, cmap="Greys_r")
-            axs[1].scatter(ssample, lline, s=10, color='blue')
-            axs[1].set_title("Relative\n Unprojected Image\nw/registered point")
-            plt.show()
-
-
-    return sample, line, dist, metric, temp_corrmap
+
+    # Estimate the transformation
+    affine = estimate_affine_transformation(base_corners, dst_corners)
```
ladoramkershner, line 765: The function call for this is: estimate_affine_transformation(destination_coordinates, source_coordinates)

That seems like the opposite of how you are calling it. I remember looking at you result and they looked good (especially since when an affine transformation is used in the non intended direction it is pretty obvious), but it terms of documentation and technical correctness, I think either this line (and the corresponding down stream code) should be swapped, or the estimate_affine() could be swapped?

autocnet/matcher/subpixel.py
ladoramkershner: I think these plots may be useful in subpixel_transformed_template, but replace the 5's for template_buffer kwarg

ladoramkershner:
I finished reviewing. I think my two comments are:
1.) Resolving the function signature on estimate_affine_transformation (or explaining to me why it is fine as is)
2.) Moving the plots that were in subpixel_template into subpixel_transformed_template and removing/adding the 'verbose' kwargs accordingly. I really liked those plots!

Always encourage push back if you don't agree, I just wanted to sum up my review!

autocnet/matcher/subpixel.py
jlaura: The naming here sucks. So I am going to address the really good comment by making the naming consistent with skimage.

jlaura:
@ladoramkershner This is updated based on the always appreciated comments. I updated the figures for the transformed work as well to include titles.

Screen Shot 2020-09-25 at 1 50 06 PM

jlaura:
I have an idea how to also avoid high correlations near the edge of the template. Once this PR looks good and is merged I will open a PR with that modification.

autocnet/matcher/subpixel.py
```diff
@@ -645,10 +655,10 @@ def geom_match(base_cube,
 
     Parameters
     ----------
-    base_cube:  plio.io.io_gdal.GeoDataset
+    destination_cube:  plio.io.io_gdal.GeoDataset
                 source image
 
-    input_cube: plio.io.io_gdal.GeoDataset
+    sourcecube: plio.io.io_gdal.GeoDataset
                 destination image; gets matched to the source image
```
ladoramkershner, line 662: These seem backwards..

autocnet/matcher/subpixel.py
```diff
@@ -703,76 +713,76 @@ def geom_match(base_cube,
...
-    if not isinstance(base_cube, GeoDataset):
-        raise Exception("match cube must be a geodataset obj")
+    if not isinstance(destination_cube, GeoDataset):
+        raise Exception("destination cube must be a geodataset obj")
 
     # Can we validate this inside the ROI object?
-    base_size_x = template_kwargs['image_size'][0]
-    base_size_y = template_kwargs['image_size'][1]
+    destination_size_x = template_kwargs['image_size'][0]
+    destination_size_y = template_kwargs['image_size'][1]
```
ladoramkershner, line 724: This also seems backwards. The image is the larger out of template and image chips. So you are trying to find out how template moves to line up with image. So (I am thinking in terms of ground matching) image -> Themis -> source

autocnet/matcher/subpixel.py
jlaura: Updated the doc string.

autocnet/matcher/subpixel.py
jlaura: This is correct. The image_size kwarg is not named the best, but needs to be consistent with sub pixel_template naming conventions. The destination is Themis. We are going to warp the source, CTX, into the destination.

ladoramkershner (approved):

PR #486, opened 2020-09-17
Adds arbitrary weights to the edges for use in then finding islands

jlaura:
This PR gets edges setup to be persistent into the DB. Once persistent in the DB, we can use the weights to do all kinds of different masking operations.

Examples:

## Applying a weight using a function
```
# Set the edge weights using some function:
def weight_func(edge):
    edge.weights = ('byid', edge.source['node_id'] % 2)
    return True

ncg.apply(weight_func, on='edge')
```

Here the arbitrary `weight_func` is being applied to every edge. Since we are applying to an edge, the first arg is the edge itself. Here we access the weights attribute (that is a dict under the hood) and use the setter to add or update the `byid` key with a value derived from the `node id` of the source node. Since the function in here runs on the edge, it can access anything that the edge can access and since the edge has a `parent` attribute that is the graph, that is basically everything.

So, we can write a function like so to apply a weight based on the number of active (not ignored) measures:
```
# Set the edge weights using some function:
def weight_func(edge):
    number_activemeasures = sum([True for i in edge.measures if i.ignore == False])
    edge.weights = ('number_activemeasure', number_activemeasures)
    return True

ncg.apply(weight_func, on='edge')
``` 

Once the edge has been updated, it is possible to filter edges based on another arbitrary function:
```
# Arbitrary function that masks based on a given threshold.
def func(edge, key, threshold=2):    
    return edge.weights[key] > threshold

subgraph = ncg.filter_edges(func, key='number_activemeasure', threshold=400)
```
Here, we return True on an edge if the weight for a given key is greater than a threshold. This example is very arbitrary in that it is requesting to retain all nodes where the edge weight is greater than 400. The arbitrary function needs to take in the object, any arguments, and then any kwargs. It needs to return a boolean.

Once we have the subgraph(s), it is possible to look at the number of independent connected components to find islands or completely disconnect subgraphs. For example: 

```
print(subgraph.connected_subgraphs())
```

This last capability is still a WIP in this PR.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -950,9 +950,11 @@ class CandidateGraph(nx.Graph):
         : Object
           A networkX graph object
         """
-        edges = [(u, v) for u, v, edge in self.edges.data(
-            'data') if func(edge, *args, **kwargs)]
-        return self.create_edge_subgraph(edges)
+        edges_to_remove = [(u, v) for u, v, edge in self.edges.data(
+                            'data') if func(edge, *args, **kwargs)]
+        subgraph = nx.create_empty_copy(self)
+        subgraph.add_edges_from(edges_to_remove)
+        return subgraph
 
     def compute_cliques(self, node_id=None):  # pragma: no cover
         """
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -750,15 +750,6 @@ class CandidateGraph(nx.Graph):
         """
         return nx.isolates(self)
 
-    def filter_by_weight(self, threshold):
-        def filter_edge(n1, n2):
-            if self[n1][n2]['weight'] > threshold:
-                return True
-            return False
-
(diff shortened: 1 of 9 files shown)


jlaura:
PR updated so that filter_edges now maintains nodes that have an incidence of zero (orphaned nodes). This should make filter edges operate more, as expected.

jlaura:
It looks like the error on the PR is due to the PVL module updates. I am going to wait for the plio work to be done to get migrated to PVL >= 1.0.0 and then rerun the CI over here.

autocnet/graph/network.py
```diff
@@ -1423,6 +1425,32 @@ def _setup_database(self):
         # Attempt to create the database (if it does not exist)
         try_db_creation(self.engine, self.config)
 
+    def _setup_edges(self):
```
ladoramkershner, line 1428: I am confused in what use case is this function useful?

ladoramkershner (approved):

PR #491, opened 2020-09-28
Redis

jlaura:
Updates redis config to not persist to disk. Addresses issue @ladoramkershner was seeing.

### DIFF
--- a/services/docker-compose.yml
+++ b/services/docker-compose.yml
@@ -8,7 +8,7 @@ services:
       POSTGRES_USER: jay
       POSTGRES_DB: kaguyasp
     volumes:
-      - "${AUTOCNET_SERVER_DB}:/var/lib/postgresql/data"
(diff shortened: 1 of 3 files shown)


services/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
       POSTGRES_USER: jay
       POSTGRES_DB: kaguyasp
     volumes:
-      - "${AUTOCNET_SERVER_DB}:/var/lib/postgresql/data"
+      - "/data/fast/ctx_db:/var/lib/postgresql/data"
```
ladoramkershner, line 11: Why change this to a hard code?

jlaura: Only because this is not an environment variable set on smalls. No idea why I (??) made this a variable in the first place.

ladoramkershner (approved):

PR #492, opened 2020-09-29
Adds arbitrary iterable submission

jlaura:
This code change lets you pass an arbitrary iterable to the submission script that then has access to the ncg. For example:

```
iterable = [list_of_coordinates]

def simple(value, ncg=None):
    print(ncg.config)
    print(value)
    
ncg.apply(simple, on=valid_iterable)
```

It is then possible to apply the func over the arbitrary iterable while having access to the ncg. This is nice if say, we have an iterable of ground control point coordinates that we want to, in parallel, add to the database.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1450,7 +1450,7 @@ class NetworkCandidateGraph(CandidateGraph):
                                         weights=json.dumps({})))
             session.add_all(to_add)
             session.commit()
-            print(len(to_add))
+
     def _setup_queues(self):
         """
         Setup a 2 queue redis connection for pushing and pulling work/results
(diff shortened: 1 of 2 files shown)


autocnet/graph/network.py
```diff
@@ -1450,7 +1450,7 @@ def _setup_edges(self):
                                         weights=json.dumps({})))
             session.add_all(to_add)
             session.commit()
-            print(len(to_add))
+
```
jlaura, line 1453: Cleanup.

autocnet/graph/network.py
```diff
@@ -1653,10 +1667,11 @@ def apply(self, function, on='edge', args=(), walltime='01:00:00', chunksize=100
             # Dispatch to either the database object message generator or the autocnet object message generator
             if isinstance(onobj, DeclarativeMeta):
                 job_counter = self._push_row_messages(onobj, on, function, walltime, filters, query_string, args, kwargs)
+            elif isinstance(onobj, list):
+                job_counter = self._push_iterable_message(onobj, function, walltime, args, kwargs)
             else:
                 job_counter = self._push_obj_messages(onobj, function, walltime, args, kwargs)
```
ladoramkershner, line 1673: Should we put this in another elif and have the else error?

jlaura:
@ladoramkershner Updated for the comment. This should be good to go. :) All the reviews the last few days are appreciated!

ladoramkershner (approved):

PR #497, opened 2020-10-26
CTX 10/23 Closeout

ladoramkershner:


autocnet/matcher/subpixel.py
```diff
@@ -399,6 +399,70 @@ def subpixel_transformed_template(sx, sy, dx, dy,
 
     return dx, dy, metrics, corrmap
 
+def subpixel_template_classic(sx, sy, dx, dy,
```
jessemapel, line 402: Did new subpixel_templace bomb in consumer testing?

ladoramkershner: They were QUITE insistent..

ladoramkershner: We just realized we took too many steps forward and need to do some more before/after comparison to make sure we are improving the code, not just changing it. Additionally, registering measures with the new method introduces an undefined  subpixel 'shiftiness' to measures on a point.

PR #498, opened 2020-10-27
Fixing NetworkCandidateGraph.from_cnet()

ladoramkershner:
'from_cnet' is a NCG class method which calls another class method 'from_filelist', however, 'from_filelist' then calls 'from_database' which is an instance method. This causes an error because it does not have access to 'self'. To fix this we created an intermediate function 'add_from_filelist' which holds the logic and 'from_filelist' just instantiates an NCG, attaches the config, and calls 'add_from_filelist'.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1844,7 +1844,7 @@ class NetworkCandidateGraph(CandidateGraph):
                 session.commit()
 
     @classmethod
-    def from_filelist(cls, filelist, clear_db=False):
+    def from_filelist(cls, filelist, config, clear_db=False):
         """
         Parse a filelist to add nodes to the database. Using the
         information in the database, then instantiate a complete,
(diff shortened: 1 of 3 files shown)


jessemapel:
@acpaquette @jlaura I am confused how the NetworkCandidateGraph is suppose to be setup for use. Should it be:

```
ncg = NetworkCandidateGraph.from_something(data, config)
```

or

```
ncg = NetworkCandidateGraph
ncg.config_from_file(config)
ncg.add_data(data)
```

jlaura:
It should be the later. This way the config can be dynamically updated and altered. 

@ladoramkershner This looks good. I realize this is sans tests and I think this is appropriate given the project you are on. Two questions: (1) is this working as expected locally? and (2) am I stepping on your toes if I write some tests after this is merged? 

Thanks, 
J

jlaura:
@ladoramkershner I think we have two options:

1. Pass a config into the from_* methods for the NCG.
2. Add a step where the user first runs config_from_* and then run from_*.

I am fine with either. The former is more convenient to the user as it is a single call. The later is really explicit (which I like). Like I said though, I think both are totally viable solutions. What do you think? And did I answer your question or miss the point you were making?

ladoramkershner:
I don't really have a preference. I think it is nice to have less requirements to instantiate an NCG, but for new users it may be confusing if you can create an NCG but can't access your database with no loud error. They would have to know to check config['database']. 

Maybe we go with option 2 and put checks for database config in the corresponding functions that access the database?

ladoramkershner:
The PR is working. However, the only ISIS generated network that I personally made/own is the kaguya TC network and it has over 300 images. So, I started working on getting the ncg.add_from_filelist() to use ncg.apply(). 

The issue I am having is that from_db() and _execute_sql() need a fully populated images table. If we stick ncg.apply(add_image, on=filelist) in the current add_from_filelist(), those functions would be run directly after all the individual slurm jobs uploading the images to the database were submitted . Additionally, having the ncg.apply(add_image, on=filelist) in the add_from_filelist() now requires the user to hand in apply kwargs (walltime, logdir, arraychunk, chunksize, etc) to add_from_filelist().

On the other hand, if we do this in a cell it is much less attractive, _execute_sql is "private", and execute_sql requires a raw sql string.

I didn't really present a solution here, so I apologize, but I think there are just some more decisions that need to be made on how we want to operate autocnet.

jessemapel (approved):
Approving even though CI is broken, but not by this PR :(

PR #502, opened 2020-11-03
Adds tests for 498

jlaura:
This PR adds:

- tests for 498
- an adapter for PosixPath types and SQLAlchemy. For testing using tmp_path (pytest) the paths are Python3 PosixPaths. This is great because of the added a capabilities, but causes issues with SQLAlchemy because SQLAlchemy does not natively work with that type. After some googling, it looks like the best place to centralize type conversions is at the psycopg2 level. The new adapters.py centralizes these and the modification to the `__init__` is needed to get the adapter registered.


autocnet/graph/node.py
```diff
@@ -514,20 +514,26 @@ def __init__(self, *args, **kwargs):
 
     def populate_db(self):
         with self.parent.session_scope() as session:
+            ip = self['image_path']
             res = session.query(Images).filter(Images.path == self['image_path']).first()
```
ladoramkershner, line 518: Why define ip in 517 if you use self['image_path] in 518?

jlaura: Good catch - a remnant from before I figured out how to get the adapter working properly. I will remove.

jlaura: @ladoramkershner this is removed. thx

ladoramkershner:
I think this looks good! There are no tests for from_cnet or place_points_from_cnet(), but those functions did not fail/were not heavily altered in PR 498..so do we want to try and incorporate tests for those in this PR?

jlaura:
@ladoramkershner Nope - I am done with tests on this for now. Someone else can pick up those tests or I will look to write some additional tests in the future.

ladoramkershner (approved):

PR #504, opened 2020-11-05
Adding functions and NB for overlap connectivity check 

ladoramkershner:
Not sure if we need a size threshold on the connected overlaps, but I added it just in case.


docs/users/installation.rst
```diff
@@ -28,9 +28,9 @@ Manual Development Environment
 ------------------------------
 To manually install AutoCnet (for example in a development environment) we must install the necessary dependencies.
 
-1. Create a virtual environment:  ``conda create -n autocnet python=3.5 && source activate autocnet``
```
ladoramkershner, line 31: One of the installation notes in the shared document stated "Do not need to specify python=3 on environment creation" I am not sure if it applied to this environment creation as well.

autocnet/graph/network.py
```diff
@@ -2221,6 +2222,87 @@ def queue_flushdb(self):
...
+        Find overlaps that do not contain any valid points. At minimum, valid points
+        include not ignored points, but additional point properties can be used to
+        further define a valid point. For example, to look at not ignored, free
+        (not ground) points; filters = {pointtype: 2}.
+
+        Parameters
+        ----------
+        filters: dict
+                 Points object properties to filter on, excluding "ignore" which is
+                 always included.
```
jessemapel, line 2236: I would instead default this argument to filters={'ignore':False} so that it's clear it's a default filter and that would allow users to not filter on it if they want to.

autocnet/graph/network.py
```diff
@@ -2221,6 +2222,87 @@ def queue_flushdb(self):
...
+
+        size_threshold: float
+                        Minimum area requirment for returned overlaps. Helps filter
+                        out overlaps that have unreasonably small areas for points.
```
jessemapel, line 2240: What units is this in?

docs/users/installation.rst
jessemapel: You are correct, we don't need it here either.

autocnet/graph/network.py
ladoramkershner: The documentation on ST_Area says that is dependent on the spatial reference frame that you use..

jessemapel: Okay, we should point to that then. Could use a see also.

jessemapel (approved):

jessemapel:
More env solve failures disregard travis.

PR #505, opened 2020-11-06
Added ISIS outlier detection notebook

jessemapel:
This notebook demonstrates how to use Autocnet's matching outlier detection to check an existing ISIS control network for bad measures.


autocnet/graph/edge.py
```diff
@@ -757,6 +757,67 @@ def _from_db(self, table_obj):
             session.expunge_all()
         return res
 
+    def compute_homography(self, method='ransac', maskname='homography', **kwargs):
```
ladoramkershner, line 760: Over arching questions. This function exists in the Edge class and the NetworkEdge inherits from the Edge class. Why not alter that function?

jessemapel: Basically the differences between how things are stored in the CandidateGraph and the NetworkCandidateGraph. The two store their matches in different ways and so you have to access them in different ways to get the inputs to the homography function.

autocnet/graph/edge.py
```diff
@@ -757,6 +757,67 @@ def _from_db(self, table_obj):
...
+        ----------
+        method : {'ransac', 'lmeds', 'normal', '8point'}
+                 The method that will be used when computing the homography.
+
+        maskname : str
+                   The column that the mask will be saved under in the masks dataframe.
+                   If the column already exists, then the mask in that column will be overwritten.
+
+        kwargs : dict
+                 Extra arguments that will be passed to the homography computation.
```
ladoramkershner, line 805: Extra arguments that will be passed to the **fundamental matrix** computation.

autocnet/graph/edge.py
ladoramkershner: 👍

PR #515, opened 2021-01-19
updated NBs plus code mods to get autocnet to work with ISIS4

ladoramkershner:


autocnet/matcher/subpixel.py
```diff
@@ -1474,7 +1474,8 @@ def subpixel_register_point(pointid,
 
     version = version.lower()
     geom_funcs = {"classic": geom_match_classic,
-                "new": geom_match
+                "new": geom_match,
+                "simple": geom_match_simple
```
jessemapel, line 1478: Is this going to impact any current workflows? I'm assuming this is just an extra option.

jessemapel (approved):

autocnet/matcher/subpixel.py
ladoramkershner: It shouldn't, but the 'new' and 'classic' methods actually do not work at the moment

ladoramkershner: Not relevant to this PR, but frustrating

ladoramkershner:
It should be good for a rereview! The notebook is formatted for the workshop as well, blanks where blanks are needed and the autocnet_workshop kernel

jessemapel (approved):

PR #516, opened 2021-01-20
populate template metric (not tested)

### LINKED ISSUE #510: Propagate Control Network doesn't seem to populate template_metric
After running `cross_instrument_matcher.propagate_control_network`, template metric is not populated on the database, which means you cannot refilter based on matching results after the points are updated. This is an issue because of how long `propagate_control_network` takes to adjust and upload points.

tgiroux:
Looks like it should be as simple as this to resolve #510, as far as I can tell the data is right there in propagate_point it just needed to get passed through. Haven't been able to test due to an environment problem I'm having with the pyproj library

### DIFF
--- a/autocnet/matcher/cross_instrument_matcher.py
+++ b/autocnet/matcher/cross_instrument_matcher.py
@@ -338,6 +338,8 @@ def propagate_point(Session,
             'path': row[7],
             'line' : line,
             'sample' : sample,
+            'template_metric' : row[3],
+            'template_shift' : row[4],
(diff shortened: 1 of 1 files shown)


tgiroux:
Ran a few test cases and it seems to work as intended. After running, measures in db reflect the same template metrics output by propagate_control_network.

jlaura (approved):

jlaura:
@Kelvinrr If these also look good to you I think we are good to merge.

PR #519, opened 2021-01-25
Parallel Ground 

Kelvinrr:
Except it doesn't work. 

Gist to get it running: 

https://gist.github.com/Kelvinrr/d81afea468fd025d2495fad6909dcc38


autocnet/matcher/subpixel.py
```diff
@@ -49,7 +50,7 @@ def check_geom_func(func):
         return func
 
     if func in geom_funcs.keys():
-        return match_funcs[func]
+        return geom_funcs[func]
```
Kelvinrr, line 53: @jlaura I think this is where you are getting your error?

jlaura: ohhh - this is a unmarked draft. I understand!

PR #521, opened 2021-02-05
added dem generation code

Kelvinrr:
Added initial DEM generation code and blob change generation from DEM depressions and planform curvature. 

Kinda dumped things into utils for now, open to requests to move them elsewhere. 

Something to consider: Colins hill shade works pretty well but I'd be interested if there is a reasonable way to use [Gdal's hill shade utils](https://gdal.org/programs/gdaldem.html), wasn't able to find an easy python interface to it, though. 

example of output: 

topos


autocnet/utils/utils.py
```diff
@@ -591,4 +597,289 @@ def import_func(func):
...
+        topo = topo + random.rand(int(nn), int(nn))*(200/(nn**alpha))
+    
+    topo = rd.rdarray(topo, no_data=0)
+    
+    curvatures = rd.TerrainAttribute(topo, attrib='slope_riserun')
+    curvatures = (curvatures - np.min(curvatures))/np.ptp(curvatures) * scale_factor
+    return topo - curvatures
+
+
+def hillshade(img, azi=255, min_slope=20, max_slope=100, min_bright=0, grayscale=False):
```
acpaquette, line 782: This is missing the alt argument

autocnet/utils/utils.py
```diff
@@ -591,4 +597,289 @@ def import_func(func):
...
+    hillshade a DEM, based on IDL code by Colin Dundas 
+    
+    Parameters
+    ----------
+    
+    img : np.array
+          DEM to hillshade
+    
+    azi : float 
+          Sun azimuth
```
acpaquette, line 793: Can we mention that this should be in degrees?

autocnet/utils/utils.py
```diff
@@ -591,4 +597,289 @@ def import_func(func):
...
     module = importlib.import_module(module, package='autocnet')
     func = getattr(module, func)
-    return func
\ No newline at end of file
+    return func
+
+
+def compute_depression(input_dem, scale_factor=1, curvature_percentile=75):
+    """
+    Compute depressions and return a new image with larges depressions filled in.
```
acpaquette, line 605: larges to large

autocnet/utils/utils.py
```diff
@@ -591,4 +597,289 @@ def import_func(func):
...
+
+    base_array = np.zeros(shape, dtype=dtype)  # Initialize your array of zeros
+
+    fill = np.ones(base_array.shape) * True  # Initialize boolean array defining shape fill
+
+    # Create check array for each edge segment, combine into fill array
+    for k in range(vertices.shape[0]):
+        fill = np.all([fill, check(vertices[k-1], vertices[k], base_array)], axis=0)
+    
+    print(fill.any())
```
acpaquette, line 734: Remove print

autocnet/utils/utils.py
```diff
@@ -591,4 +597,289 @@ def import_func(func):
...
+    for k in range(len(scales)):
+        nn = size/scales[k]
+        topo = scipy.misc.imresize(topo, (int(nn), int(nn)), "cubic", mode="F")
+        topo = topo + random.rand(int(nn), int(nn))*(200/(nn**alpha))
+    
+    topo = rd.rdarray(topo, no_data=0)
+    
+    curvatures = rd.TerrainAttribute(topo, attrib='slope_riserun')
+    curvatures = (curvatures - np.min(curvatures))/np.ptp(curvatures) * scale_factor
+    return topo - curvatures
```
acpaquette, line 779: What's the curvature for here?

autocnet/utils/utils.py
```diff
@@ -591,4 +597,289 @@ def import_func(func):
...
+    nx = (np.roll(dem, 1, axis = 1) - dem)
+    ny = (np.roll(dem, 1, axis = 0) - dem)
+    sz = np.shape(nx)
+    nz = np.ones(sz)
+    nl = np.sqrt(np.power(nx, 2.0) + np.power(ny, 2.0) + np.power(nz, 2.0))
+    nx = nx/nl
+    ny = ny/nl
+    nz = nz/nl
+
+    math.cos(math.radians(1))
```
acpaquette, line 845: This can be removed, left over from prototyping

acpaquette (commented):
Couple small things to change, and a small question regarding topology generation

autocnet/utils/utils.py
Kelvinrr: "Eroding" the cliff edges helps refine the topology a little creating something more organic, especially when using higher alphas.

PR #526, opened 2021-02-11
Affine and Grayscale update

acpaquette:
Adds the ability to generate an affine matrix and it's subsequent delta matrix. These can be applied to copies of a dem to moc a bad image registration.

This PR also updates the grayscale parameter to avoid checking the same variable more often than necessary.

### DIFF
--- a/autocnet/cg/change_detection.py
+++ b/autocnet/cg/change_detection.py
@@ -2,8 +2,6 @@ import numpy as np
 
 import matplotlib
 from matplotlib.path import Path
-from matplotlib import pyplot as plt
-
 import cv2
 
 from sklearn.cluster import  OPTICS
--- a/autocnet/cg/change_detection.py
+++ b/autocnet/cg/change_detection.py
@@ -909,3 +909,62 @@ def generate_boulder_field(dem, num_boulders, x_shift_min = 5, x_shift_max = 10,
             after_polys.append(after_geom)
 
     return before_dem, before_polys, after_dem, after_polys
+
+def create_affine(x_scale, y_scale, x_shift, y_shift, rotation,
+                  x_scale_delta, y_scale_delta, x_shift_delta, y_shift_delta, rotation_delta):
+
+    """
+    Generate an initial affine transformation matrix and generate a second
+    transformation matrix based on deltas applied to the initial affine matrix
+
+    x_scale : float
+              Value to scale the image by in the x direction. Where 1 is no scale,
+              > 1 is scaled up and < 1 is scaled down
+
+    y_scale : float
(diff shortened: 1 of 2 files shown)


autocnet/cg/change_detection.py
```diff
@@ -912,3 +902,62 @@ def generate_boulder_field(dem, num_boulders, x_shift_min = 5, x_shift_max = 10,
             after_polys.append(after_geom)
 
     return before_dem, before_polys, after_dem, after_polys
+
+def create_affine(x_scale, y_scale, x_shift, y_shift, rotation,
```
jlaura, line 906: Does it make sense to do this manually (it is great to know how to!) or is it better to use the SKImage transformation API: https://scikit-image.org/docs/dev/api/skimage.transform.html

We are using the transformation API in other places in the code base, but I am not sure if that API fits your needs here.

autocnet/cg/change_detection.py
```diff
@@ -2,8 +2,6 @@
 
 import matplotlib
```
jlaura, line 3: Is matplotlib in the environment.yml and the meta.yml?

environment.yml
```diff
@@ -30,6 +30,7 @@ dependencies:
   - pytest-cov
   - pysis
   - pvl < 1.0
+  - richdem
```
jlaura, line 33: What are the implications of the added dependency? Is this a stand alone thing? Also, has it been added to the meta.yml so that conda knows to include it in the package dependencies?

autocnet/cg/change_detection.py
```diff
@@ -753,21 +751,13 @@ def hillshade(img, azi=255, alt=60, min_slope=20, max_slope=100, min_bright=0, g
     bright = ((dprod - min_dprod) + min_bright)/((max_dprod - min_dprod) + min_bright)
 
     if grayscale:
-        qq=(255*bright)
+        arrforout = 255*bright
```
Kelvinrr, line 754: makes more sense, didn't really understand why grayscale before wasn't.

Kelvinrr (commented):
does it make sense to move to the scipy affines in this PR? 

Also, richdem might as well be added to meta.yml

environment.yml
Kelvinrr: It's pretty standalone package, from conda info: 

```
dependencies:
    libgcc-ng >=7.3.0
    libstdcxx-ng >=7.3.0
    numpy >=1.14.6,<2.0a0
    python >=3.6,<3.7.0a0
    setuptools
```

definitely needs to be added to meta.yml

Kelvinrr (approved):

PR #527, opened 2021-02-11
Multiple updates - separated by commit message

### LINKED ISSUE #542: Conditionally compute overlaps
add_from_filelist() skips already loaded images, but appends same overlaps into database. We should only calculate the overlaps on images that were newly added to the database.

jlaura:
These are the changes that have been made during CTX work.

- Place points in overlaps defaults to using the minimum rotated rectangle instead of drawing points between shortest edges. This fixes issues with points drawing East/West when they were specified N/S.
- Updates to proj4 have deprecated the use of radians=True. This fixes the og2oc and oc2og calls to use degrees.
- Fixes #542, find most insteresting to first attempt using orb, then vlfeat(sift), and finally orb with a lower edge threshold. If all three attempts fail, the a priori pointing is used.
- Changes from Kelvin's PR to fix minot bugs in signatures / returns have been fixed
- Fixes an issue where flushing the queue flushed all users (connected to db=0, the default in autocnet) queues (redis).
- Merges an upstream change for use with ISIS >= 4.1 (or whenever we transitioned to ISISDATA instead of ISIS3DATA)

If these look okay, I would suggest not squashing and merging.


autocnet/graph/network.py
```diff
@@ -1468,14 +1468,18 @@ def _setup_queues(self):
...
+    def clear_queues(self):
         """
         Delete all messages from the redis queue. This a convenience method.
         The `redis_queue` object is a redis-py StrictRedis object with API
         documented at: https://redis-py.readthedocs.io/en/latest/#redis.StrictRedis
         """
-        return self.redis_queue.flushall()
+        queues = [self.processing_queue, self.completed_queue, self.working_queue]
+        for q in queues:
+            self.redis_queue.delete(q)
```
AustinSanders, line 1482: The docs suggest that we're deleting messages from the queues, but .delete() removes the queue entirely.  Should we use [.empty()](https://github.com/rq/rq/blob/016da14723d21f6801b19cbff97f2fb9bee5c5db/rq/queue.py#L115) or can we clarify in the docs that the queues will be deleted?

autocnet/cg/cg.py
```diff
@@ -333,14 +332,21 @@ def distribute_points_classic(geom, nspts, ewpts, **kwargs):
     ewpts : int
             The number of points to attempt to place
             in the E/W (right/left) direction
+    
+    use_mrr : boolean
+              If True (default) compute the minimum rotated rectable bounding
```
AustinSanders, line 337: Should this be "minimum rotated rectangle"?

AustinSanders (changes requested):

autocnet/cg/cg.py
jlaura: yes

jlaura:
@AustinSanders Updated for all comments. Thanks for the review!

AustinSanders (approved):

Kelvinrr (approved):
seems fine to me

autocnet/matcher/subpixel.py
```diff
@@ -1431,19 +1448,19 @@ def subpixel_register_point(pointid,
...
-    if not ncg.Session:
-        raise BrokenPipeError('This func requires a database session from a NetworkCandidateGraph.')
-
+    
     match_func = check_match_func(match_func)
     geom_func = check_geom_func(geom_func)
 
-    if isinstance(pointid, Points):
-        pointid = pointid.id
-
```
ladoramkershner, line 1443: This either needs to be kept in the code, or subpixel_register_point needs to work with the object handed in. Right now subpixel_register_point just fails when trying to preform the query on line 1459.

I tried using the query_string argument to get around this (SELECT id FROM points) but the query return is still considered a point object when it gets down to line1459

ladoramkershner:
@jlaura @Kelvinrr 
This still does not have the fix in it I was asking for..should we revert or open a new PR?

PR #530, opened 2021-02-26
initial commit for variable pointid

ladoramkershner:
update the doc strings in spatial.overlap and added a kwarg for filling out the point's "identifier" field when laying down the points using either place_points_in_overlap or place_points_in_image

### DIFF
--- a/autocnet/io/db/model.py
+++ b/autocnet/io/db/model.py
@@ -225,7 +225,7 @@ class Images(BaseMixin, Base):
     @classmethod
     def union(cls, session):
         """
-        The boundary formed by unioning (or merging) all of the input footprints. The result 
+        The boundary formed by unioning (or merging) all of the input footprints. The result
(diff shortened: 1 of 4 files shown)


autocnet/io/db/model.py
```diff
@@ -296,7 +296,7 @@ class Points(BaseMixin, Base):
 
     id = Column(Integer, primary_key=True, autoincrement=True)
     _pointtype = Column("pointType", IntEnum(PointType), nullable=False)  # 2, 3, 4 - Could be an enum in the future, map str to int in a decorator
-    identifier = Column(String, unique=True)
+    identifier = Column(String)
```
ladoramkershner, line 299: I am removing this unique requirement because it seems like we are still using the auto-incrementing id to enforce uniqueness on write out, so it would be unnecessary here.

jlaura (approved):

PR #531, opened 2021-03-10
Parallel Ground

jlaura:
I rebased to get some PRs from last month properly merged locally. I do not believe I messed anything up.  As always, reviews appreciated.

This adds parallel ground to the codebase.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -21,15 +21,15 @@ jobs:
         shell: bash -l {0}
     steps:
       - uses: actions/checkout@v2
-        with:
-          submodules: true
       - uses: conda-incubator/setup-miniconda@v2
         with:
           miniconda-version: "latest"
+          mamba-verion: "*"
+          use-mamba: true
+          channels: conda-forge, defaults
+          channel-priority: true
           activate-environment: autocnet
           environment-file: environment.yml
-          auto-activate-base: false
-          auto-update-conda: true
           python-version: ${{ matrix.python-version }}
       - name: Check build environment
         run: |


autocnet/matcher/ground.py
```diff
@@ -0,0 +1,323 @@
...
+                    match_kwargs=match_kwargs, \
+                    verbose=verbose)
+        except Exception as e:
+            raise Exception(e)
+            match_results.append(e)
+            continue
+
+        # Element 0 of each list is an id. We don't need this id because of how this was refactored to work
+        # in parallel. If we set to none though, the check below fails because one of the elements is None...
+        match_results.append(['foo',
```
Kelvinrr, line 70: What is the foo? I think in the original version this was an ID?

autocnet/io/db/model.py
```diff
@@ -365,6 +366,37 @@ def pointtype(self, v):
     #def subpixel_register(self, Session, pointid, **kwargs):
     #    subpixel.subpixel_register_point(args=(Session, pointid), **kwargs)
 
+class GroundPoint(BaseMixin, Base):
+    __tablename__ = 'groundpoints'
+    latitudinal_srid = -1
+
+    id = Column(Integer,primary_key=True, autoincrement=True)
+    path = Column(String, nullable=False)
```
Kelvinrr, line 374: Path to what? The image it's attached to? If so, maybe there is a cleaner solution considering the image table?

autocnet/matcher/ground.py
```diff
@@ -0,0 +1,323 @@
...
+                imageid=node['node_id'],
+                serial=node.isis_serial,
+                measuretype=3,
+                choosername='add_measures_to_ground')
+
+    with ncg.session_scope() as session:
+        point = session.query(Points).filter(Points.id == point.id).one()
+
+        point.measures.append(m)
+        point.reference_index = len(point.measures) - 1  # The measure that was just appended is the new reference
```
Kelvinrr, line 321: If I am understanding this correctly, the reference measure is always the last in the measure set? If every time you make a point with the ORM and get point.measures, and it's always in the same order, and the last index is always reference, do we need the reference index member?

Kelvinrr (commented):
just some questions

autocnet/matcher/ground.py
jlaura: It was the ID, but ID is not longer needed and I did not want to refactor any more than necessary. It can be any value except None. I originally changed it to None, but then the line that checks for any None values tosses the whole measure. I'll be comfortable refactoring this further once we have some tests or at least some manual test cases in place.

autocnet/io/db/model.py
jlaura: This is the path to file used as the base, so in our case the THEMIS mosaic. I'll add a comment in here so that we know that is what this path is. 

I originally thought about adding the Themis base to the images table, but decided against it because I did not want to mix datasets with different purposes in there. In my mind, the images table is implicitly the 'input' data that is being controlled. Perhaps that is a poor assumption?

autocnet/matcher/ground.py
jlaura: That is correct. When I added the reference measure code, I did it because of ground. When I started working ground and trying to minimally change the current serial implementation I realized that we could get away with not having a reference measure.

Fortunately, having the reference solved a different problem. Sometimes during relative control, the reference measure is in a no data area on the image with the lowest image id. In that case, the code now falls over gracefully and tries the next image in the stack. So, we can have a reference index on a point that is non zero. Here I am implicitly suggesting that the above should always be zero, and I think the is should be (e.g., no reason to even do the len(point.measures)-1). The only reason that the reference_index would be non-zero is if someone ran this func twice. If that were the case, we would output an invalid ISIS control network because we would have a point with multiple measures in the same image.

jlaura:
Marked as a draft because I need to rebase to get this to be a clean set of changes. I'll do that Monday.

autocnet/cg/cg.py
```diff
@@ -516,7 +513,7 @@ def distribute_points_in_geom(geom, method="classic",
     # Decision Tree
     if ratio < 0.16 and geom.area < 0.01:
         # Class: Slivers - ignore.
-        return []
+        return np.array([])
     elif geom.area <= 0.004 and ratio >= 0.25:
         # Single point at the centroid
         valid = single_centroid(geom)
```
ladoramkershner, line 519: single_centroid() returns a list not an np.array, which is the new functin return type. This comment also applies to line 525 and 533

autocnet/graph/network.py
```diff
@@ -1723,7 +1727,7 @@ def apply(self,
                      time=walltime,
                      partition=queue,
                      output=log_dir+f'/autocnet.{function}-%j')
-        submitter.submit(array='1-{}%{}'.format(job_counter,arraychunk), chunksize=chunksize)
+        job_str = submitter.submit(array='1-{}%{}'.format(job_counter,arraychunk), chunksize=chunksize, exclude=exclude)
```
ladoramkershner, line 1730: It doesn't seem like anything is happening with this job_str, did you want to return it for logging? Or just assign in for FY22 logging efforts?

autocnet/graph/network.py
```diff
@@ -1813,6 +1818,11 @@ def to_isis(self,
         cnet.to_isis(df, path, targetname=target)
         cnet.write_filelist(fpaths, path=flistpath)
 
+
+        # Even though this method writes, having a non-None return 
+        # let's a user work with the data that is passed to plio
+        return df
```
ladoramkershner, line 1824: Loooveeee this

autocnet/graph/network.py
```diff
@@ -2451,7 +2464,71 @@ def distribute_ground(self, distribute_points_kwargs={}):
...
+        """
+        valid = []
+        with self.session_scope() as session:
+            res = session.query(Overlay).filter(func.array_length(Overlay.intersections, 1) >= threshold).all()
+            for r in res:
+                coords = cg.distribute_points_in_geom(r.geom, **distribute_points_kwargs)
+                if len(coords) > 0:
+                    valid.append(coords)
+        valid = np.vstack(valid)
+        return valid
```
ladoramkershner, line 2534: Wondering if the convention you set up with cg.cg.distribute_points_in_geom returning np.ndarrays should be extended to these ground distribution functions

autocnet/io/db/model.py
ladoramkershner: I am slightly confused why we are using a whole separate table for ground points. We have the point types to indicate which points are ground (or at least constrained). The ground images could go into the images table and we could add a 'control' or 'ground' boolean column to indicate which images to write out (aka control) or include in the overlap calculations. 

I am sorry to be a pain, I just think this is a lot of redundant information. I am also not opposed to keeping it like this if there is something I am missing.

autocnet/io/db/model.py
```diff
@@ -365,6 +366,37 @@ def pointtype(self, v):
...
+    choosername = Column("ChooserName", String)
+    apriorisample = Column(Float)
+    aprioriline = Column(Float)
+    sample = Column(Float, nullable=False)
+    line = Column(Float, nullable=False)
+    _geom = Column("geom", Geometry('POINT', srid=latitudinal_srid, dimension=2, spatial_index=True))
+    ignore = Column(Boolean, default=False)
```
ladoramkershner, line 381: If we keep this table, should we have reference_index, identifier (as it serves a different function than choosername), measures, and pointtype columns/attributes?

autocnet/matcher/subpixel.py
```diff
@@ -58,7 +59,8 @@ def check_match_func(func):
     match_funcs = {
         "classic": subpixel_template_classic,
         "phase": iterative_phase,
-        "template": subpixel_template
+        "template": subpixel_template,
+        "mutualinformation": mutual_information
```
ladoramkershner, line 63: This needs to be taken out for now (I know it is already done locally, just want to put a 'flag' for here)

autocnet/matcher/subpixel.py
```diff
@@ -839,12 +843,16 @@ def geom_match_simple(base_cube,
     autocnet.matcher.subpixel.subpixel_template: for list of kwargs that can be passed to the matcher
     autocnet.matcher.subpixel.subpixel_phase: for list of kwargs that can be passed to the matcher
     """
-
+    t1 = time.time()
```
ladoramkershner, line 846: Do you want to keep these timers in here? On one hand it could be good while optimizing our subpixel registration functions, but on the other hand it is messy and not functionally necessary.

autocnet/matcher/subpixel.py
```diff
@@ -908,19 +917,22 @@ def geom_match_simple(base_cube,
...
+    print(f'Affine warp took {t3-t2} seconds.')
     if verbose:
         fig, axs = plt.subplots(1, 2)
         axs[0].set_title("Base")
         axs[0].imshow(roi.Roi(bytescale(base_arr, cmin=0), bcenter_x, bcenter_y, 25, 25).clip(), cmap="Greys_r")
         axs[1].set_title("Projected Image")
         axs[1].imshow(roi.Roi(bytescale(dst_arr, cmin=0), bcenter_x, bcenter_y, 25, 25).clip(), cmap="Greys_r")
         plt.show()
-
+    print(base_arr.shape, dst_arr.shape)
```
ladoramkershner, line 930: remove this print?

autocnet/matcher/subpixel.py
```diff
@@ -1455,29 +1470,41 @@ def subpixel_register_point(pointid,
...
+    if isinstance(pointid, Points):
+        pointid = pointid.id
+    
+    t1 = time.time()
     with ncg.session_scope() as session:
-        measures = session.query(Measures).filter(Measures.pointid == pointid).order_by(Measures.id).all()
-        
+        point = session.query(Points).filter(Points.id == pointid).one()
+        measures = point.measures
+        #measures = session.query(Measures).filter(Measures.pointid == pointid).order_by(Measures.id).all()
```
ladoramkershner, line 1480: delete this line?

autocnet/matcher/subpixel.py
```diff
@@ -1593,3 +1623,125 @@ def subpixel_register_points(subpixel_template_kwargs={'image_size':(251,251)},
                                 subpixel_template_kwargs=subpixel_template_kwargs,
                                 **cost_kwargs)
 
+def register_to_base(pointid,
```
ladoramkershner, line 1626: How does this function differ from geom_match conceptually / why is it needed? Even in autocnet/matcher/ground.py propagate_ground_point() just uses geom_match_simple.

autocnet/spatial/overlap.py
```diff
@@ -518,3 +517,49 @@ def place_points_in_image(image,
...
+        reference_measure = point.measures[reference_index]
+        reference_image_id = reference_measure.imageid
+
+        images = session.query(Images).filter(Images.geom.ST_Intersects(point._geom)).all()
+        print(f'Placing measures into {len(images)-1} images.')
+        for image in images:
+            if image.id == reference_image_id:
+                continue  # This is the reference image, so pass on adding a new measure
+            
+            if cam_type == "isis":
```
ladoramkershner, line 544: Since there is only one if case for cam_type should we 

a) check and make sure an invalid cam_type is not passed in?
b) have an else clause to state passed cam_type is not supported?

autocnet/spatial/overlap.py
```diff
@@ -518,3 +517,49 @@ def place_points_in_image(image,
             points.append(point)
     print(f'Able to place {len(points)} points.')
     Points.bulkadd(points, ncg.Session)
+
+def add_measures_to_point(pointid, cam_type='isis', ncg=None, Session=None):
```
ladoramkershner, line 521: Doc string please!

PR #535, opened 2021-03-22
Added missing output in python workshop nb

jessemapel:


jessemapel:
More workshop PR stuff.

PR #547, opened 2021-04-27
Adds update from jigsaw and fixes rotted tests.

jlaura:
Comments are inline. The goal here was to get the update from jigsaw in. For Valles Marineris (sp?) it take ~120 seconds to update the 1.5M measures.

As a secondary goal, I tried to fix as many tests as possible. The two integration tests are still failing and can be fixed in a different PR.

### DIFF
--- a/autocnet/graph/network.py
+++ b/autocnet/graph/network.py
@@ -1813,7 +1813,7 @@ class NetworkCandidateGraph(CandidateGraph):
         cnet.to_isis(df, path, targetname=target)
         cnet.write_filelist(fpaths, path=flistpath)
 
-    def update_from_jigsaw(self, path, k=10):
+    def update_from_jigsaw(self, path, pointid_func=lambda x: int(x.split('_')[-1])):
         """
         Updates the measures table in the database with data from
         a jigsaw bundle adjust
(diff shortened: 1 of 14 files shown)


autocnet/io/db/tests/test_model.py
```diff
@@ -150,35 +131,6 @@ def test_update_point_geom(session, data, new_adjusted, expected):
 def test_measures_exists(tables):
     assert model.Measures.__tablename__ in tables
 
-@pytest.mark.parametrize("measure_data, point_data, image_data", [({'id': 1, 'pointid': 1, 'imageid': 1, 'serial': 'ISISSERIAL', 'measuretype': 3, 'sample': 0, 'line': 0},
```
jlaura, line 153: This test was removed because the new test was added for merging jigsaw results with an autocrat project.

autocnet/matcher/cpu_outlier_detector.py
```diff
@@ -128,7 +128,7 @@ def spatial_suppression(df, bounds, xkey='x', ykey='y', k=60, error_k=0.05, nste
         if n_y_cells <= 0:
             n_y_cells = 1
 
-        grid = np.zeros((n_y_cells, n_x_cells), dtype=np.bool)
+        grid = np.zeros((n_y_cells, n_x_cells), dtype=bool)
```
jlaura, line 131: as above np types are deprecated

conftest.py
```diff
@@ -71,8 +70,8 @@ def default_configuration():
                   'username': 'postgres',
                   'password': 'NotTheDefault',
                   'host': 'localhost',
-                  'port': 5432,
-                  'pgbouncer_port': 5432,
+                  'port': 35432,
```
jlaura, line 73: This should be a permanent update once GitHub CI is working. This keeps the testing off the 'normal' db port.

autocnet/graph/network.py
```diff
@@ -2133,14 +2106,13 @@ def clear_db(self, tables=None):
                     except Exception as e:
                         warnings.warn(f'Failed to reset primary id sequence for table {t}')
 
-    def place_points_from_cnet(self, cnet):
+    """def place_points_from_cnet(self, cnet):
```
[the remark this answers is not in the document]
ladoramkershner: do you want to remove this functionality all together? Commenting it out seems a little too wild west

autocnet/matcher/tests/test_outlier_detector.py
```diff
@@ -67,9 +67,11 @@ def test_suppress(self):
     def test_suppress_non_optimal(self):
         with warnings.catch_warnings(record=True) as w:
             mask, k = cpu_outlier_detector.spatial_suppression(self.df, self.domain, k=30, xkey='lon', ykey='lat')
-            self.assertEqual(len(w), 1)
             self.assertEqual(k, 59)
             self.assertTrue(issubclass(w[0].category, UserWarning))
+            # The old check for # of warning is brittle because deprecation warnings in upstream packages are counted.
```
ladoramkershner, line 72: This comment doesn't need to be in the actual code base

autocnet/io/db/controlnetwork.py
```diff
@@ -93,62 +96,81 @@ def db_to_df(engine, sql = """
...
-              the network candidate graph associated with the measure/database
-              being updated.
+    cnet : pd.DataFrame
+           plio.io.io_control_network loaded dataframe
+
+    measures : pd.DataFrame
+               of measures from a database table. 
+    
+    connection : object
+                 An SQLAlchemy DB connection object
```
ladoramkershner, line 122: could you define pointid_func?

autocnet/graph/network.py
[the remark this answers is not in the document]
jlaura: Happy to delete. If I do, the chances of this functionality coming back are close to zero. We will not remember that this was a thing. Nominally, I am fine with that - I don't remember why this was written. If that is good with you I'll remove this and the associated test.

autocnet/io/db/controlnetwork.py
jlaura: Done.

autocnet/matcher/tests/test_outlier_detector.py
jlaura: Removed.

autocnet/graph/network.py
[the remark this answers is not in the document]
ladoramkershner: It was to encourage the cross use of ISIS and AutoCNet, this function is the main functionality of ncg.from_cnet()..

Is there another path a user can take if they have an ISIS control network and they want to read it into AutoCNet to do outlier rejection or some thing else?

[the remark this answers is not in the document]
jlaura: Not as far as I know. This function is, as far as I know, still working. I turned it off because of that test. The test is keeping an extra session open which then causes all subsequent tests to fail. One option would be to remove the test knowing that this function is untested and then, at some other time, write a functioning test. Thoughts on that approach?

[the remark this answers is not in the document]
ladoramkershner: I like that option better than removing this function. I don't really trust us (mostly me) to remember that this function is untested. Didn't we create the ncg session_scope to avoid orphaned sessions? Would leveraging that help with the test failure?

My top pick is to fix the test, but if altering the test is more a pain than what I am thinking I would be okay with commenting out the test. It could serve as a place holder and hopefully be more obvious than a no test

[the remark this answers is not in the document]
jlaura: I spent maybe 30 minutes or so working on the test and trying to use a scoped session. I was not able to get this working inside of the testing harness with the scoped session and am honestly not 100% sure why. This PR has two other known failing tests and I would just leave this one failing. The problem with that approach is that a failure on this test cascades into all subsequent DB tests. This is because each test ensures that the session is properly closed by counting the number of sessions and this test / function not closing the session means all later tests fail with n+1 sessions open.

ladoramkershner (approved):

PR #549, opened 2021-04-30
Single clean commit for parallel ground

jlaura:
This is a single, clean commit PR, that adds ground. I tried to incorporate all past comments into this. I will try to get some tests added shortly.


autocnet/matcher/ground.py
```diff
@@ -0,0 +1,353 @@
...
+    with ncg.session_scope() as session:
+        # Check to see if the point already exists
+        g = Points(path=ground_mosaic.file_name,
+                    choosername='find_most_interesting_ground',
+                    aprioriline=line,
+                    apriorisample=sample,
+                    line=newline,
+                    sample=newsample,
+                    geom=p,
+                    ignore=False)
```
ladoramkershner, line 253: shouldn't this be appending to the CandidateGroundPoints database model?

autocnet/matcher/ground.py
```diff
@@ -0,0 +1,353 @@
...
+        return
+
+    left_x, _, top_y, _ = image.image_extent
+    newsample = left_x + interesting.x
+    newline = top_y + interesting.y
+
+    # @LAK - this needs eyeballs to confirm correct oc/og
+    newpoint = isis.point_info(ground_mosaic.file_name, newsample, newline, 'image')
+    p = Point(newpoint[0].get('PositiveEast360Longitude'),
+              newpoint[0].get('PlanetocentricLatitude'))
```
ladoramkershner, line 242: This looks good to me! ISIS operates in ocentric, the reason we needed to switch to ographic is when we were working with pyproj to convert lat/lon to/from BCBF.

autocnet/matcher/ground.py
jlaura: Updated! Good catch. :)

jlaura:
@ladoramkershner Updated based on your comment (locally; yesterday) and then synced the changes (today). I think this should be good to go.

autocnet/io/db/model.py
```diff
@@ -443,7 +477,7 @@ def try_db_creation(engine, config):
     Points.rectangular_srid = rectangular_srid
     Points.semimajor_rad = spatial['semimajor_rad']
     Points.semiminor_rad = spatial['semiminor_rad']
-    for cls in [Points, Overlay, Images, Keypoints, Matches]:
+    for cls in [Points, Overlay, Images, Keypoints, Matches, CandidateGroundPoints()]:
```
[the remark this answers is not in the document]
jlaura: Fixed this - I have no idea why I included an instance and not a class...amazing - it works either way. sqlalchemy must be doing some introspection that works either way.

PR #555, opened 2021-05-10
ticked version

Kelvinrr:
Not sure if we really need to change much else? I recall version 0.4.0 was the target. 

Does this version have to be manually pushed to https://anaconda.org/usgs-astrogeology/autocnet after this gets merged?


jlaura:
We also need to tick the recipe version.

PR #558, opened 2021-05-14
Added Fourier Mellen registration func

Kelvinrr:
Tried to keep the implementation as simple as possible to make it easier to copy+paste into whatever way we refactor subpixel,py

### DIFF
--- a/autocnet/matcher/subpixel.py
+++ b/autocnet/matcher/subpixel.py
@@ -8,7 +8,6 @@ import numbers
 
 import sys
 
-from skimage import registration as reg
 from skimage import transform as tf
 from skimage import data
 from skimage import registration 
@@ -272,7 +271,7 @@ def subpixel_phase(sx, sy, dx, dy,
         if (s_image is None) or (d_template is None):
             return None, None, None
 
-    (shift_y, shift_x), error, diffphase = reg.phase_cross_correlation(s_image, d_template, **kwargs)
+    (shift_y, shift_x), error, diffphase = registration.phase_cross_correlation(s_image, d_template, **kwargs)
     dx = d_roi.x - shift_x
     dy = d_roi.y - shift_y
 
@@ -1761,16 +1760,21 @@ def register_to_base(pointid,
     
 
 
-def estimate_logpolar_transform(img1, img2, low_sigma=1, high_sigma=5, verbose=False): 
+def estimate_logpolar_transform(img1, img2, low_sigma=0.5, high_sigma=30, verbose=False): 
     """
     Estimates the rotation and scale difference for img1 that maps to img2 using phase cross correlation on a logscale projection. 
     
-    Essentially, scale and angular changes in cartesian space become translation in log-polar space. Translation from subpixel registration 
+    Scale and angular changes in cartesian space become translation in log-polar space. Translation from subpixel registration 
     in log-polar space can then be translated into scale/rotation change in the original cartesian images. This scale + roation 
     change estimation is then returned as an affine transform object. This can then be used before other subpixel registration 
     methods to enable scale+rotation invariance. 
     
-    
+    See Also 
+    --------
+
+    skimage.filters.difference_of_gaussians : Bandpass filtering using a difference of gaussians 
+    skimage.filters.window : Simple wondowing function to remove spectral leakage along the axes in the fourier transform
+
     References
     ----------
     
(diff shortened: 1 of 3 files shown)


autocnet/matcher/subpixel.py
```diff
@@ -1752,4 +1758,190 @@ def register_to_base(pointid,
...
+        ax[3].set_title("Log-Polar-Transformed\nModified FFT")
+        ax[3].imshow(warped_img2_fs, cmap='magma')
+        fig.suptitle('Working in frequency domain can recover rotation and scaling')
+        plt.show()
+
+        print(f"Recovered value for cc rotation: {recovered_angle}")
+        print()
+        print(f"Recovered value for scaling difference: {shift_scale}")
+
+    shift_y, shift_x = np.asarray(img1.shape) / 2 - 0.5
```
jlaura, line 1866: Need a comment here on why -0.5. The hard coded subpixel translations become hellish to debug later if someone else thinks this is an error.

autocnet/matcher/subpixel.py
```diff
@@ -1752,4 +1758,190 @@ def register_to_base(pointid,
...
+    tf_scale = tf.SimilarityTransform(scale=shift_scale)
+    tf_rotate = tf.SimilarityTransform(rotation=np.deg2rad(recovered_angle))
+    tf_shift = tf.SimilarityTransform(translation=[-shift_x, -shift_y])
+    tf_shift_inv = tf.SimilarityTransform(translation=[shift_x, shift_y])
```
jlaura, line 1870: This looks like it is pulling from the Affine work I did a bit ago on sub windows. Are we sure that this is correct? I am asking because the sub window affine stuff I did does not align with the full affine transform on an image that you did (hence the different sub pixel transformations).

autocnet/matcher/tests/test_subpixel.py
```diff
@@ -108,6 +133,26 @@ def test_subpixel_transformed_template(apollo_subsets):
     assert ny == pytest.approx(54.0081)
 
 
+def test_estimate_logpolar_transform(iris_pair):
+    img1, img2 = iris_pair 
+    affine = sp.estimate_logpolar_transform(img1, img2) 
+    print(affine.scale, affine.rotation, affine.translation)
```
jlaura, line 139: yank the print?

autocnet/matcher/tests/test_subpixel.py
```diff
@@ -108,6 +133,26 @@ def test_subpixel_transformed_template(apollo_subsets):
...
+
+    assert pytest.approx(affine.scale, 0.1) == 0.71
+    assert pytest.approx(affine.rotation, 0.1) == 0.34 
+    assert pytest.approx(affine.translation[0], 0.1) == 283.68
+    assert pytest.approx(affine.translation[1], 0.1) == -198.62
```
jlaura, line 144: These are verified off something? Asking because we had tests like this and to prove to ourselves that our sub pixel was correct, we went to the canned, binary ndarrays that Evin created that had boxes, crosses, etc.

jlaura (changes requested):

autocnet/matcher/tests/test_subpixel.py
Kelvinrr: Largely based on my toy example where I made sure that the initial rotation and scale difference was successfully reproduced. I can switch to a cross but the IRIS test seems to be doing the same thing? 


Screen Shot 2021-05-14 at 8 58 16 AM

autocnet/matcher/subpixel.py
Kelvinrr: I think our attempts to get an ROI transformed have had problems because we were handling the origin of the affine correctly. In the case of the camera model, the origin for the returned affine is `(0,0)`  of the complete image, and I think there is something happening in trying to change the origin for the complex affine to be the ROI origin instead that gets screwed up. 

In this case, this is rotation from the center (hence the weird offsets) and scale from the origin. So it's significantly simpler. That being said, figuring that out is probably where most of my time when since Scikit-image's docs on affine transforms seem opaque or unintuitive in a lot of ways about precisely how they do things.

autocnet/matcher/subpixel.py
```diff
@@ -1752,4 +1758,190 @@ def register_to_base(pointid,
...
+    radius = shape[0] // 4  # only take lower frequencies
+    warped_img1_fs = tf.warp_polar(img1_fs, radius=radius, output_shape=shape,
+                                 scaling='log', order=0)
+    warped_img2_fs = tf.warp_polar(img2_fs, radius=radius, output_shape=shape,
+                               scaling='log', order=0)
+
+    warped_img1_fs = warped_img1_fs[:shape[0] // 2, :]
+    warped_img2_fs = warped_img2_fs[:shape[0] // 2, :]
+    shifts, error, phasediff = registration.phase_cross_correlation(warped_img1_fs,
+                                                       warped_img2_fs,
```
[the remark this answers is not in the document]
Kelvinrr: looks like I had both imports in, I'll remove the `as reg` version

autocnet/matcher/subpixel.py
```diff
@@ -1752,4 +1758,190 @@ def register_to_base(pointid,
...
+    Parameters
+    ----------
+    
+    img1: np.ndarray 
+          The source image, this is the image that is used as a base as img2 is registered to the center on img1
+    
+    img2: np.ndarray 
+          The image that will be moved to match img1
+
+    low_sigma : float, list, np.array
```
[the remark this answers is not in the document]
Kelvinrr: I can add some further explanations into the bandpass filter and windowing, but from what I understand, these don't really have optimal choices and is super data dependent. Maybe better to link scikit-image's explanations rather than rewriting things?

autocnet/matcher/subpixel.py
```diff
@@ -266,7 +272,7 @@ def subpixel_phase(sx, sy, dx, dy,
         if (s_image is None) or (d_template is None):
             return None, None, None
 
-    (shift_y, shift_x), error, diffphase = register_translation(s_image, d_template, **kwargs)
+    (shift_y, shift_x), error, diffphase = reg.phase_cross_correlation(s_image, d_template, **kwargs)
```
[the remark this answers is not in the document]
Kelvinrr: Couldn't find much on this, the depreciation wasn't mentioned in the change-logs for any recent releases. 

Judging from the PR, seems we want to pin to >=0.17? 

https://github.com/scikit-image/scikit-image/commit/200ae1b41beb531beb529ac872314e120fbe6094

autocnet/matcher/subpixel.py
jlaura: :+1: Agree 100% on the affine docs being hard to detangle! Hopefully some of what you learned here, we can bring back for the remaining AutoCNet hours this FY.

jlaura (approved):

PR #559, opened 2021-05-20
MI Matcher Addition

acpaquette:
Adds the mutual information matcher for use within autocnet. Mainly through the `subpixel_template` method where the mutual_information function should be passed in as the `func=mutual_information` argument.


jessemapel (commented):
It looks like we can merge this as-is right now and in the future we can test swapping in the scikit image MI matcher function. This looks good to me, but I'm really bad at reviewing autocnet so I'm going to let some other's chime in too.

autocnet/matcher/tests/test_mutual_information.py
```diff
@@ -0,0 +1,31 @@
...
+    assert corrilation == pytest.approx(0)
+
+def test_mutual_information():
+    d_template = np.array([[i for i in range(50, 100)] for j in range(50)])
+    s_image = np.array([[150 - i for i in range(150)] for j in range(150)])
+
+    x_offset, y_offset, max_corr, corr_map = mutual_information.mutual_information(d_template, s_image, bins=20)
+    assert x_offset == -25
+    assert y_offset == -25
+    assert max_corr == 2.9755967600033015
```
jessemapel, line 31: I would add a simple sanity check on the correlation map. Maybe just check the dimensions and that the min/max are within the expected range, I think between 0 and the number of bins.

autocnet/matcher/mutual_information.py
```diff
@@ -0,0 +1,95 @@
+import numpy as np
+
+def mi(t1, t2, **kwargs):
```
jlaura, line 3: Can we name this something more descriptive? Or use it as a context inside of mutual_information?

autocnet/matcher/mutual_information.py
```diff
@@ -0,0 +1,95 @@
...
+
+    Parameters
+    ----------
+
+    t1 : ndarray
+         First image to use in the histogram comparison
+
+    t2 : ndarray
+         Second image to use in the histogram comparison
+
```
jlaura, line 17: Need a see also for what the kwargs coming in are. Right now, I wouldn't know where to even look if I used just the doc string.

autocnet/matcher/mutual_information.py
jlaura: Also - did my gist include a 'from' line that attributed this code to the blog post that I initial found it on?

autocnet/matcher/mutual_information.py
```diff
@@ -0,0 +1,95 @@
...
+    px_py = px[:, None] * py[None, :] # Broadcast to multiply marginals
+    # Now we can do the calculation using the pxy, px_py 2D arrays
+    nzs = pxy > 0 # Only non-zero pxy values contribute to the sum
+    return np.sum(pxy[nzs] * np.log(pxy[nzs] / px_py[nzs]))
+
+def mutual_information(d_template, s_image, bins=100, func=mi):
+    """
+    Applys the mutual information matcher function over a search image using a
+    defined template. Where the search area is 2x the size of the template image
+
```
jlaura, line 40: Where is the 2x the size defined? Looks like I could run this with identically sized d_template, s_image ndarrays and it would work fine.

autocnet/matcher/mutual_information.py
```diff
@@ -0,0 +1,95 @@
...
+            if corr > max_corr:
+                max_corr = corr
+                max_i = i
+                max_j = j
+            corr_map[i, j] = corr
+
+    # This is still operating at the pixel scale. Use the template_match_autoreg
+    # logic to achieve submpixel weighting.
+    x_offset = max_j - (template_size[1]/2)
+    y_offset = max_i - (template_size[0]/2)
```
jlaura, line 93: This is not sub pixel enabled right now. Is that intentional? Subpixel weighting should be very easy to copy/paste in (cleaner would be to refactor to get the subpixel weighting into a function that both this and the linked call).

Lines 49-72 show how to get a subpixel weighting using the corrmap: https://github.com/USGS-Astrogeology/autocnet/blob/dev/autocnet/matcher/naive_template.py#L49

jlaura (changes requested):

autocnet/matcher/mutual_information.py
acpaquette: There is no `from` line but there is a link to scikit image if that is where you obtained the implementation. Then scikit image also has a link to [this paper](https://www.sciencedirect.com/science/article/abs/pii/S0031320398000910?via%3Dihub) that outlines their implementation

autocnet/matcher/mutual_information.py
acpaquette: This kinda depends if we use scikit images implementation or the matcher that was in your gist

autocnet/matcher/mutual_information.py
acpaquette: It can in its current state. There are a few options:
- Force user to provide the expected dimensions
- Only ask for an image or template size then calculate the other
- Ignore the dimension constraint but don't search a subset of the search space
- Buffer the search space so that the template can still exhaust the whole search space (scikit image approach)

autocnet/matcher/mutual_information.py
acpaquette: This was copied from the notebook and I wasn't sure if we wanted to shoot for subpixel registration for the image registration sprint. This is something I can add to the PR

autocnet/matcher/mutual_information.py
jlaura: All of those options suggest to me that the docs string is wrong and needs to be updated to be correct. Does this code have a size requirement?

acpaquette: Yes, that's correct. The only real size requirement is that the template must be smaller than the search image size

acpaquette: Is that something that should be addressed in the `subpixel_template_classic` or in this function?

autocnet/matcher/mutual_information.py
jessemapel: Making this a contextual function means we can't unit test it anymore. I would move this back out to a module function and rename these. This function should be mutual_information as it computes the mutual information between two images. The other function should be called mutual_information_match as it matches two images using the mutual information.

autocnet/matcher/tests/test_mutual_information.py
```diff
@@ -0,0 +1,35 @@
...
+    s_image = np.ones((100, 100))
+
+    s_image[25:75, 25:75] = d_template
+
+    x_offset, y_offset, max_corr, corr_map = mutual_information.mutual_information(d_template, s_image, bins=20)
+    assert x_offset == 0.9633527901853505
+    assert y_offset == 0.5
+    assert max_corr == 2.9755967600033015
+    assert corr_map.shape == (51, 51)
+    assert np.average(corr_map) == 1.3199548152066989
```
jessemapel, line 35: Average is a really bad sanity check as it's going to be highly sensitive and doesn't actually tell us anything. I did a little bit more research and it turns out bounds on MI are rather complex, so here we should just sanity check the lower bound of 0.

autocnet/matcher/mutual_information.py
jlaura: > Where the search area is 2x the size of the template image

Needs to be changed. from your comment:

> Yes, that's correct. The only real size requirement is that the template must be smaller than the search image size

The doc string is wrong.

jlaura (commented):
This has 2 outstanding comments and then likely needs a rebase. After that, I think this is good togo.

jlaura:
@acpaquette Here is the improved center of gravity based subpixel approach: https://github.com/USGS-Astrogeology/autocnet/blob/dev/autocnet/matcher/naive_template.py#L62; using an implementation right from scipy and then the same logic to back out the transformation chain as the naive template matcher.

autocnet/matcher/tests/test_mutual_information.py
acpaquette: Fixed with a lower bound test!

autocnet/matcher/mutual_information.py
acpaquette: Moved the contextual mutual_information function out of the matcher

jlaura (approved):

jessemapel (approved):

jlaura:
@acpaquette Looks like the MI test needs a `pytest.approx` to some reasonable number of decimal places. One other minor request - would you mind updating the changelog to add this PR?

I'll open an issue that the overlay test is leaving a dangling session and we can fix that in another PR.

Issue #557, opened 2021-05-11
CodeCov on CI is not working


jlaura:
On merged PRs the code coverage report is not being uploaded. Right now, CI is set to fail if the coverage is not uploaded.

jlaura:
CodeCov is working on PRs.

PR #562, opened 2021-05-27
Update environment.yml

jlaura:
Updates the environment to use the new plio release with PVL >= 1.0

Depending upon CI to test this. Will mark as a draft until I have passing CI.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -20,7 +20,7 @@ requirements:
       - knoten>=0.2.1
       - networkx=2.4
       - opencv<=3.5
-      - plio>=1.2.3
+      - plio>=1.2.5
       - plurmy>=0.1.1
       - pysis>=0.1.1
       - pyyaml=5.3.1
@@ -43,14 +43,14 @@ requirements:
       - jupyter
       - knoten
       - ncurses
-      - networkx
+      - networkx>=2,<3
       - numpy
       - opencv<=3.5
       - pillow
-      - plio>=1.1.0
+      - plio>=1.2.5
       - plurmy
+      - proj>=7.0.0,<8
       - psycopg2
-      - pvl<1.0.0
       - pyproj
       - pysis
       - pytest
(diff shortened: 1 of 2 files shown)


jlaura:
The environment is properly setting up and the tests are passing as expected (2 xfails - intentional fails). The CI failure is because of the code coverage upload issue (#557)

environment.yml
```diff
@@ -22,15 +22,15 @@ dependencies:
...
+  - plio >= 1.2.5
   - plurmy
   - psycopg2
   - proj >= 7.0.0, <8
   - pyproj
   - pytest
   - pytest-cov
   - pysis
-  - pvl < 1.0
+  - pvl >= 1.0
```
jessemapel, line 33: Does AutoCNet actually use PVL 1.0 features or is this just a transitive dependency from PLIO?

jlaura: Transitive - so probably safe to remove from here now that we aren't trying to do weird version mixing things. Ohh -this needs to update the meta.yml too! PR inbound with both changes.

conda/meta.yaml
```diff
@@ -43,27 +43,27 @@ requirements:
...
       - pyyaml
       - redis-py
-      - scikit-image
+      - scikit-image>=0.17
       - scikit-learn
       - scipy<=1.2.1
       - shapely
-      - sqlalchemy
-      - sqlalchemy-utils
+      - sqlalchemy>-1.4,<1.5
```
acpaquette, line 65: What does the `>-` mean? Or is it a typo?

acpaquette (commented):
The plio and pvl changes look in line with the updates. One small question but the other changes look good!

rbeyer:
If *this* package has `import pvl` in its code (which it does), then shouldn't the line for `pvl` in these files just be updated and not deleted?  Otherwise you're relying on other dependencies to install *pvl* for you.  Which they will, but seems strange to me.

PR #567, opened 2021-05-30
Tested Cluster Submission

jlaura:
This PR moves what was `acn_submit` into the library (autocnet.graph.cluster_submit) and then adds tests for acn_submit functionality. 

This PR also yanks the .ropeproject files that are a legacy from some past IDE out.


autocnet/graph/network.py
```diff
@@ -1420,6 +1420,14 @@ def _setup_dem(self):
...
             self.dem = None
 
+    @property 
+    def Session(self):
+        return self._Session
+
+    @Session.setter
+    def Session(self, Session):
+        self._Session = Session
+
```
jlaura, line 1430: This is a property now to improve testability.

autocnet/graph/cluster_submit.py
```diff
@@ -88,23 +88,71 @@ def main(msg):
...
+    queue_name : str
+                 The name of the queue to remove an object from
+
+    remove_key : obj
+                 The message to remove from the list
+    """
+    # The operation completed. Remove this message from the working queue.  
+    queue.lrem(queue_name, 0, remove_key)
+
+def manage_messages(args, queue):
```
jessemapel, line 131: Needs a doc string to indicate what's expected in args.

autocnet/graph/network.py
```diff
@@ -2225,6 +2241,17 @@ def queue_length(self):
         llen = self.redis_queue.llen(self.config['redis']['processing_queue'])
         return llen
 
+    def get_work_queue_messages(self):
+        return self.redis_queue.lrange(self.working_queue, 0, -1)
+
+    def get_timed_out_jobs(self, clear=False):
```
jessemapel, line 2247: It looks like clear is unused here.

autocnet/graph/network.py
```diff
@@ -2225,6 +2241,17 @@ def queue_length(self):
         llen = self.redis_queue.llen(self.config['redis']['processing_queue'])
         return llen
 
+    def get_work_queue_messages(self):
```
jessemapel, line 2244: Both these functions need doc strings

autocnet/graph/network.py
```diff
@@ -1711,16 +1724,19 @@ def apply(self,
         rconf = self.config['redis']
         rhost = rconf['host']
         rport = rconf['port']
-        processing_queue = rconf['processing_queue']
-
+        try:
+            processing_queue = getattr(self, redis_queue)
+        except AttributeError:
+            print(f'Unable to find attribute {redis_queue} on this object. Valid queue names are: "processing_queue" and "working_queue".')
```
jessemapel, line 1730: Add these options to the argument doc string.

autocnet/graph/tests/test_cluster_submit.py
```diff
@@ -0,0 +1,186 @@
...
+def _generate_obj(msg, ncg):
+    return Foo()
+
+@pytest.mark.parametrize("along, func, msg_additions", [
+                            ('edge', _do_nothing, {'id':(0,1), 'image_path':('/foo.img', '/foo2.img')}),  # Case: callable func
+                            ('node', _do_nothing, {'id':0, 'image_path':'/foo.img'}),   # Case: callable func
+                            ('edge', 'test', {'id':(0,1), 'image_path':('/foo.img', '/foo2.img')}),  # Case: callable func
+                            ('node', 'test', {'id':0, 'image_path':'/foo.img'}),   # Case: callable func
+                            ('edge', 'graph.tests.test_cluster_submit._do_nothing', {'id':(0,1), 'image_path':('/foo.img', '/foo2.img')}),  # Case: callable func
+                            ('node', 'graph.tests.test_cluster_submit._do_nothing', {'id':0, 'image_path':'/foo.img'}),   # Case: callable func
```
jessemapel, line 101: These comments are confusing. If these are all testing the same case then why are there multiple tests? Be more specific about what each case is supposed to test or just remove them.

jessemapel (approved):
Better testing is always good to see. Just needs some small docs clean-up.

autocnet/graph/tests/test_cluster_submit.py
jlaura: Updated to indicate (1) generic func, (2) method on a fake obj, (3) imported func from the module.

autocnet/graph/network.py
jlaura: Added.

autocnet/graph/network.py
jlaura: I removed these. YAGNI convenience funcs that I thought that I removed.

autocnet/graph/network.py
jlaura: removed.

autocnet/graph/cluster_submit.py
jlaura: Doc string added here and to process.

jlaura:
@jessemapel This should be good to go for a re-review.

autocnet/graph/network.py
jessemapel: I don't see them in the doc string. I'd expect this to be the name of a queue, not the name of an attribute on the NCG.

jlaura: Here is the added doc string.
```    
 redis_queue : str
                      The redis queue to push messages to that are then pulled by the
                      cluster job this call launches. 
```

So, this could just be the string name. The issue with that is that the config specifies a base name and then munges that to add a prefix `:prefix`. The user would need to specify the full queue name. As a convenience the NCG has attributes that are the names of the queues. That is why I am pulling the queue name via the attribute.

Via attribute is also a safety net for users that are duplicating notebooks. This way they don't have to remember the base name that they are using.

jessemapel (approved):

jlaura:
I'll resolve based on the other merged PR and then merge. Thanks for the review!

jessemapel (approved):

PR #569, opened 2021-06-01
PVL does exist

### LINKED ISSUE #568: pvl still being imported and used
Not sure how I missed these before, but grep is your friend, it tells me that *pvl* is still being imported and used in *autocnet* in these modules:

* graph/node.py (now hidden behind a NotImplementedError, so should never be executed)
* io/db/controlnetwork.py
* matcher/subpixel.py
* spatial/isis.py
* utils/hirise.py

So adding *pvl* back to the requirements.txt and environment.yml is needed unless you think these uses can also be scrubbed (I don't know the code well enough to make that call, but doesn't seem like it to me).

I *think* you can just use *pvl* 1.0+ now without issue, but it is hard to tell, and harder since none of the existing tests seem to exercise the pvl usage.  I'm happy to tackle issues as they come up.

jlaura:
Fixes #568.

A bunch of those imports are a surprise to me. Legacy of old stuff, etc. That doesn't matter though! So, PVL is added back in. Looking at the code, this should be fine as a straight insert of PVL 1.0. The usage is almost always ISIS cubes.

### DIFF
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: autocnet
-  version: 0.4.0
+  version: 0.4.1
   
 channels:
   - conda-forge
@@ -50,6 +50,7 @@ requirements:
       - plio>=1.3
       - plurmy
       - proj >= 7.0.0, <8
+      - pvl >= 1.0
       - psycopg2
       - pyproj
       - pysis
--- a/environment.yml
+++ b/environment.yml
@@ -26,6 +26,7 @@ dependencies:
   - plurmy
   - psycopg2
   - proj>=7.0.0,<8
+  - pvl>=1.0
   - pyproj
   - pytest
   - pytest-cov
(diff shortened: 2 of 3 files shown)


rbeyer (approved):
Looks good to me.

However, I don't think that the part of *autocnet* that uses *pvl* is actually exercised by tests, so we may encounter some issues in practical usage, but this will maybe force those errors.

Alternately, put back in the `pvl < 1.0` for this PR.  Write some tests, and then submit another PR with pvl >= 1.0 and those tests passing.

Meh, po-tay-to, po-tah-to.

jlaura:
@rbeyer - I agree that writing tests for these would be better. I skipped it because:

(1) matcher/subpixel.py is up for a major rework next FY to get the API fixed. In doing that pvl should be refactored out into the ROI object.
(2) some of the other cases, I'm not sure if that code can simply be removed since it is older / unused experiments.

PR #574, opened 2021-06-14
Test fixes for 571 disabling autoflush

jlaura:
@ladoramkershner This is a test fix. I am not sure how to test given that you are seeing issues only at a really large scale.


jlaura:
Weird - these pass locally. I need to figure out what to do with the session inside of the tests. I wonder if I am seeing some OSX vs. Linux differences or something else.

PR #578, opened 2021-07-23
Updates to get kalisiris working and improved subpix

jlaura:
@rbeyer  Would you mind taking a look at the change in spatial.isis? I had to remove the .value call. My env has PVL version 1.2.1 in it.


rbeyer:
So when I remove the .value call, I get a test error:
```
pytest autocnet/spatial/tests/test_isis.py
================================ test session starts ================================
platform darwin -- Python 3.7.10, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /Users/rbeyer/software/planetarypy/autocnet, configfile: pytest.ini
plugins: cov-2.12.1, mock-3.6.1
collected 4 items

autocnet/spatial/tests/test_isis.py ..F.                                      [100%]

===================================== FAILURES ======================================
___________________________ TestISIS.test_image_to_ground ___________________________

self = <autocnet.spatial.tests.test_isis.TestISIS testMethod=test_image_to_ground>

    def test_image_to_ground(self):
        lon, lat = si.image_to_ground(self.cube, 20, 20)
>       self.assertEqual(274.14948072713, lon)
E       AssertionError: 274.14948072713 != Quantity(value=274.14948072713, units='DEGREE')

autocnet/spatial/tests/test_isis.py:121: AssertionError
```

Because now you're getting a pair of pvl Quantity tuples rather than a pair of floats, which seems like the wrong output here.

Why did you have to remove it?  What problem were you trying to solve by doing so?

jlaura:
@rbeyer I am getting a float has no attribute value. So my call is returning a float, not a pal type.

rbeyer:
Okay, but are you getting that the unit test fails before removing `.value` or after?

I suspect that the test will fail for you as it has for me after you remove it, but somewhere else, you're calling `image_to_ground()` with some arguments that are causing the internal `point_info()` to return something that isn't what is expected by this function.

So, how are you calling it when it fails for you, as written, and you must change it to make it work?

jlaura:
The call is being made from here: https://github.com/USGS-Astrogeology/autocnet/blob/dev/autocnet/matcher/subpixel.py#L899 That looks like the vanilla case?

Also, looking at the code, why would the first, if sequence not need to use .values accessor, but the second not a sequence needs too? I would expect the API to be either a returned array of PVL objs or a single PVL obj. Right now, it looks like the API is __getitem__ access for a sequence and obj access for a single response. I have to look at the test setup to see what's up in there.

rbeyer:
Okay, my test suite was not complete enough, and `image_to_ground()` is not working correctly or consistently.  There are four "flavors" of input (single values or arrays, and then images which are map projected and those which are not).  The test suite only covered half of those cases, and the two that it didn't: fail (surprise!).  I can now build a minimal failing test that gets your error, and the other thing you noticed about the array code provides the other failure.  You know, I noticed that earlier, but thought: "the test is working," and didn't poke at it hard enough to discover the problem.

Depending on my afternoon, there may be a PR before EOD, certainly by Monday.

rbeyer:
When #580 gets merged, you can and should remove your change to `spatial/isis.py` which is only a partial solution, whereas #580 should cover *all* the ways that `image_to_ground()` was previously broken.

jlaura:
@rbeyer @Kelvinrr This is updated and tested locally matching CTX to THEMIS properly. Note the changes in the isis image to ground (0.5 pixel handling) for `mappt`.  Otherwise, this should be really straightforward! (I hope...)

jlaura:
I'll get a change in to maintain lon, lat. I'll also think about getting some sort of architecture.md document in that document a lon, lat convention.

autocnet/spatial/isis.py
```diff
@@ -215,16 +214,20 @@ def point_info(
                     coordsys="UNIVERSAL"
                 ),
                 "image": dict(
-                    sample=xx,
-                    line=yy,
+                    # Convert from PLIO pixels to ISIS pixels
+                    sample=xx+0.5,
+                    line=yy+0.5,
```
Kelvinrr, line 219: in line 198: 

```
    if point_type == "image":
        # convert to ISIS pixels
        for i in range(len(x_coords)):
            x_coords[i] += 0.5
            y_coords[i] += 0.5
```

Assuming I am not missing something, if `point_type="image"` and it's a projected cube, are you not adding the offset twice?

jlaura: Yes - I think that you are correct. I want to check this to make sure that removing the second +0.5 does not cause the good results I am seeing to revert. If it does, I'll have to look to see where a 0.5 pixel offset could be coming from.

jlaura: @Kelvinrr Testing this this morning and I am only seeing quality registrations when the +0.5 is in BOTH places. What I am not sure about though is why this might be the case. Is this a convenient fiction? Is this because of how the arguments are being passed. I am test more with a bunch of print debugging to understand what is going on.

jlaura: @Kelvinrr I can confirm that these are not being double called. What I am not sure of is why they are not being double called. Below is my current thinking and some logs to get your input.

So the point_type is ground, but the coordinates are in image space (not lon, lat).

```
# This is when projecting the 4 corners and the center to compute the affine. First, image (projected) to ground is called:

# Type is ground so the point_type check should not shift anything.
Original [6744] [3007] # ground pixel (not lon, lat) coordinate
After point type check: [6744] [3007]
Has mapping! [6744] [3007]  # confirm map projected
MAPPT ARGS:  {'sample': 6744.5, 'line': 3007.5, 'allowoutside': False, 'type': 'image'} image. # Now apply the shift; this image coordinates to ground coordinates

# Now this is the ground to image to get backup into the CTX image
Original [326.37688451778] [24.928005738351] # lon lat
After point type check: [326.37688451778] [24.928005738351] # shouldn't change

# Next ground pt
Original [6744] [3057]
After point type check: [6744] [3057]
Has mapping! [6744] [3057]
MAPPT ARGS:  {'sample': 6744.5, 'line': 3057.5, 'allowoutside': False, 'type': 'image'} image
Original [326.37688451778] [24.84365271406]
After point type check: [326.37688451778] [24.84365271406]
Original [6794] [3057]
After point type check: [6794] [3057]
Has mapping! [6794] [3057]
MAPPT ARGS:  {'sample': 6794.5, 'line': 3057.5, 'allowoutside': False, 'type': 'image'} image
Original [326.46123754207] [24.84365271406]
After point type check: [326.46123754207] [24.84365271406]
Original [6794] [3007]
After point type check: [6794] [3007]
Has mapping! [6794] [3007]
MAPPT ARGS:  {'sample': 6794.5, 'line': 3007.5, 'allowoutside': False, 'type': 'image'} image
Original [326.46123754207] [24.928005738351]
After point type check: [326.46123754207] [24.928005738351]
Original [6769.0] [3032.0]
After point type check: [6769.0] [3032.0]
Has mapping! [6769.0] [3032.0]
MAPPT ARGS:  {'sample': 6769.5, 'line': 3032.5, 'allowoutside': False, 'type': 'image'} image
Original [326.41906102992] [24.885829226206]
After point type check: [326.41906102992] [24.885829226206]
Estimation of the transformation took 19.385334253311157 seconds.
```

Kelvinrr: wwhhhaaaaaat, does that mean we should add +5 to the mappt call as well? presumably it's not getting the +5 added? Seems like that loop is not changing the values in place correctly.

jlaura: The mappt call has the values being added correctly (see the two subsequent lines that start 'Has mapping!' and 'MAPPT ARGS'. When the mapping group exists, we are properly adding the 0.5 offset. This is with the code as it is in the PR. If I revert and remove the +0.5 on the mappt image dict, the results are incorrect. If I remove the +0.5 on the for loop when point_type == 'image' the results are also incorrect. The 0.5 pixel offset is not being added twice when we are running ground to image because the point_type == ground.

Kelvinrr: wouldn't it just be cleaner to add the +0.5 in line 239 (where the point list is being created)? And just make them match in the way the offset is being added?

jlaura: Ahh - I gotcha I think. 

Remove:

```
    if point_type == "image":
        # convert to ISIS pixels
        for i in range(len(x_coords)):
            x_coords[i]+0.5
            y_coords[i]+0.5
```

and swap into:
```
            p_list = [
                f"{samp}, {line}" for samp, line in zip(x_coords, y_coords)
            ]
```

to make:
```
            p_list = [
                # plio pixels to isis pixels
                f"{samp+0.5}, {line+0.5}" for samp, line in zip(x_coords, y_coords)
            ]
```

Kelvinrr: yeah

jlaura: New commit in for a review with the change. Thanks for the comment and the discussion.

Kelvinrr (approved):

PR #580, opened 2021-07-23
image_to_ground fix

### LINKED ISSUE #579: spatial/isis.py::image_to_ground() not providing good returns in all cases.
As discovered in #578, the `image_to_ground()` function fails to return reasonable data under two conditions:

1. a non-map-projected cube and arrays of latitude and longitude will return a pair of numpy arrays, the first of which will contain the values of the coordinates (improperly a mix of lon and lat), and the second will contain an array of identical strings which will be the "units" of the values in the first array, instead of one array of lons and one of lats.
2. a map-projected cube and single values of longitude and latitude will result in an `AttributeError: 'float' object has no attribute 'value'`

The alternate situations: non-map-projected cube and single values of lon and lat, or a map-projected cube given numpy arrays of lon and lat return correctly.

rbeyer:
The `spatial/isis.py::image_to_ground()` function calls the `point_info()` function which just returns the results dictionary parsed from the PVL-text return of ISIS `mappt` or `campt`.  Turns out that while both return PVL-text with keys like `PlanetocentricLatitude` one of those ISIS programs returns just a plain numeric value, and the other returns a value and a unit.  To a PVL parser, those are different things, and one is returned as a `float` or `int` and the other is returned as a `pvl.collections.Quantity` object, from which the value can be extracted, but they are different kinds of objects, and `image_to_ground()` was not properly handling that duality.

Now it does.  More robust tests were added for image_to_ground() and ground_to_image(), just for good measure.

Would close #579, and is relevant to discussion in #578.

### DIFF
--- a/autocnet/spatial/isis.py
+++ b/autocnet/spatial/isis.py
@@ -308,13 +308,28 @@ def image_to_ground(
     res = point_info(cube_path, sample, line, "image")
 
     if isinstance(sample, (abc.Sequence, np.ndarray)):
-        lons, lats = np.asarray([
-            [r[lontype], r[lattype]] for r in res
-        ]).T
+        lon_list = list()
+        lat_list = list()
+        for r in res:
+            lon_list.append(_get_value(r[lontype]))
+            lat_list.append(_get_value(r[lattype]))
+
+        lons = np.asarray(lon_list)
+        lats = np.asarray(lat_list)
     else:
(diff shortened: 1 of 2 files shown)


rbeyer:
Oh, I should mention that the `autocnet/utils/tests/test_serializers.py` test fails, but it was broken like that when I got it.

jlaura (approved):

PR #583, opened 2021-08-06
Setting up AutoCNet to interact with RTDs

ladoramkershner:
Added relevant documentation for autocnet to interface with ReadtheDocs. Had to incorporate a less than ideal work around for the kalasiris import that happens on autocnet import. This fix included using the conf.py file to create a bogus $ISISROOT/bin/xml/DeNile.xml file for kalasiris to scrape, otherwise it hard fails on import. 

I will make notes of where this fix is and will mention this PR in an issue to decouple the kalasiris import from the autocnet import.


docs/Makefile
```diff
@@ -180,3 +180,6 @@ pseudoxml:
 	$(SPHINXBUILD) -b pseudoxml $(ALLSPHINXOPTS) $(BUILDDIR)/pseudoxml
 	@echo
 	@echo "Build finished. The pseudo-XML files are in $(BUILDDIR)/pseudoxml."
+
+postclean:
+	rm -rf ../fake/
```
ladoramkershner, line 185: This is part of the kalasiris fix.

docs/conf.py
```diff
@@ -45,13 +45,20 @@ def __getattr__(cls, name):
 
 sys.path.insert(0, os.path.abspath('../'))
 
+os.environ['ISISROOT'] = '../fake/path/isis/root'
+os.environ['ISISDATA'] = '../fake/path/isis/data'
+os.system("mkdir -p ${ISISROOT}/bin/xml")
+os.system("touch ${ISISROOT}/bin/xml/DeNile.xml")
+os.system("mkdir -p ${ISISDATA}")
+
```
ladoramkershner, line 53: This is part of the kalasiris fix.

rbeyer:
So you don't actually have to make real directories to make RTD pass.  I thought you did, too, but you can actually take advantage of the mocking mechanism in Python to mock the existance of some modules if you don't actually need them to build your documentation (which you typically don't).

I had the same issue with my *hiproc* module.  In order to run the documents, it had to import *kalasiris*, and a bunch of other modules, but the documentation build didn't actually need them, it just needed the various imports that `import hiproc` triggered to succeed.

An example of doing this mocking approach can be found in https://github.com/rbeyer/hiproc/blob/main/docs/conf.py

And if you really want background you can read [this Issue](https://github.com/readthedocs/readthedocs.org/issues/8205) on RTD and [this one](https://github.com/sphinx-contrib/autoprogram/issues/49) against sphinx-autoprogram as I bumbled my way to understanding.

ladoramkershner:
Oh wow.. that looks WAY more elegant than my solution. I will give it a try right now. Thank you so much for the quick feedback @rbeyer

docs/environment.yml
```diff
@@ -0,0 +1,47 @@
+channels:
+  - conda-forge
+  - usgs-astrogeology
+  - menpo
+
+dependencies:
```
AustinSanders, line 6: I don't have much experience here, but I get the feeling that not all of these dependencies are strictly necessary for building the docs.  I could very easily be wrong here, but it might be worth checking.

AustinSanders (approved):

Issue #582, opened 2021-07-28
Create a provenance table for function results tracking


ladoramkershner:
Every apply jobs produces a result, that equates to the status message(s) given out by the applied function. We currently attach the result to a log status and the entire log message (which also includes function parameterize) gets printed out to the log file. The current information getting tracked is fairly thorough, but the information would be much more usable if it was put into a 'provenance' database table as opposed to a log file. 

Finding the relevant information currently requires grepping hundreds of thousands of log files which takes ~20 min to run. Then, if you have one 'space' out of place, your grep will return nothing.. and you have to start all over. It makes it really difficult to debug.

The major issue to solve with adding this, is that we cannot guarantee that every workflow will use the same processes. So the table would need to be dynamic in terms of number/naming of columns. We may be able to create a bare bones provenance table with foreign key information to imageid, pointid, measureid (all of which can be null) then add a column based on the chooser keyword passed into ncg.apply(). I am not sure if the add a column could be added as a function under the orm Provenance model, but that function can get called at the top of every ncg.apply call.

Kelvinrr:
Do you have an idea of what the schema for such a table would look like? Basically, specifically what you would want to store. 

Postgres has powerful json tools so we might be able to leverage that (e.g. store kwargs). From what I understand is, whenever we apply a function to a measure, point or image, we store the ID, the function name, function parameters, and some kind result? The former two are easy but the last one is kinda tricky.

jessemapel:
Maybe we can record the results by tracking changes to the ORM with something like this https://stackoverflow.com/questions/29921260/tracking-model-changes-in-sqlalchemy

acpaquette:
@Kelvinrr I agree with storing the JSON but from experience with PDS and large JSON stores in significantly large databases, it can be difficult to run queries against those JSON stores. I have only experimented with this briefly, so I could be wrong, but the majority of queryable aspects should probably be stored in columns in basic types like we have been doing for the majority of autocnet databases. From your comment something like storing the "...measure, point or image, ... the ID, the function name..." as common data types would make things much faster to query, then hold the function parameters as JSON on the record in the table. 

I'm sure you've given this some thought, I just figured I would chime in and mention that directly querying JSON can be seriously slow and painful.

Kelvinrr:
[...]

PR #588, opened 2021-08-23
Table and job history

Kelvinrr:
Addresses #582 

Still updating as I test things.


Kelvinrr:
Ran it through a full job and it all seems to be working fine. 

Screen Shot 2021-07-14 at 3 28 08 PM

autocnet/graph/cluster_submit.py
```diff
@@ -175,14 +181,30 @@ def manage_messages(args, queue):
...
+        # Apply the algorithm
+        response = process(msgdict)
+        # Should go to a logger someday!
+        print(response)
+    
+    out = stdout.getvalue()
+    # print to get everything on the logs in the directory
+    print(out)
+    serializedDict = json.loads(msg)
+    success = True if "success" in msgdict["results"][0]["status"].split(" ")[0].lower() else False
```
Kelvinrr, line 201: I think we need to homogenize how we use the results key. It seems only some really fill it in and others don't. It would be useful to have a key that's just a success boolean and a separate string message.

autocnet/io/db/model.py
```diff
@@ -591,8 +592,58 @@ def measuretype(self, v):
...
+
+class PointHistory(BaseMixin, Base): 
+    __tablename__ = 'points_history'
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    measureId = Column(Integer,  ForeignKey('points.id'))
+    eventTime = Column(DateTime)
+    executedBy = Column(String)
+    before = Column(JSONB)
+    after = Column(JSONB)
+
```
Kelvinrr, line 625: These can probably be programmatically generated like the triggers, but I think I've spent enough time on this and might save it as a future enhancement.

autocnet/io/db/model.py
```diff
@@ -591,8 +592,42 @@ def measuretype(self, v):
             v = MeasureType(v)
         self._measuretype = v
 
+
+class JobsHistory(BaseMixin, Base): 
+    __tablename__ = 'jobs_history'
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    jobId = Column("jobId", Integer)
```
jlaura, line 599: What is the rationale behind having these be camel case? We have to use quotes to work with them on the db side and camel case is weird on the python side. why not snake case or simply all lowercase?

autocnet/io/db/model.py
```diff
@@ -591,8 +592,42 @@ def measuretype(self, v):
...
+    args = Column(JSONB)
+    results = Column(JSONB)
+    logs = Column(String)
+    success = Column(Boolean, default=False)
+
+
+class MeasuresHistory(BaseMixin, Base): 
+    __tablename__ = 'measures_history'
+    id = Column(Integer, primary_key=True, autoincrement=True)
+    fk = Column(Integer)
```
jlaura, line 610: fk == foreign key? If so, why aren't these standard foreign key declarations? 

These are a one-to-many I believe? https://docs.sqlalchemy.org/en/14/orm/basic_relationships.html#one-to-many

autocnet/io/db/triggers.py
```diff
@@ -86,3 +86,126 @@
...
+$BODY$
+LANGUAGE sql IMMUTABLE STRICT;
+COMMENT ON FUNCTION jsonb_delete_left(jsonb, text) IS 'delete key in second argument from first argument';
+
+CREATE OPERATOR - ( PROCEDURE = jsonb_delete_left, LEFTARG = jsonb, RIGHTARG = text);
+COMMENT ON OPERATOR - (jsonb, text) IS 'delete key from left operand';
+
+--
+
+CREATE OR REPLACE FUNCTION jsonb_delete_left(a jsonb, b text[])
```
jlaura, line 114: Question, as this is new to me. These can have the same name, but different signature and PostgreSQL knows to dispatch properly based on the input args?

PR #603, opened 2021-09-16
Improving place_points_from_cnet speed

ladoramkershner:
Uses a copy method to load an ISIS controlnetwork into the AutoCNet style database as opposed to the previous for loop model add. Also added residuals and maxResidual columns to point table.

Tests still needed!


autocnet/graph/network.py
```diff
@@ -2222,50 +2222,102 @@ def clear_db(self, tables=None):
...
+
         semi_major, semi_minor = self.config["spatial"]["semimajor_rad"], self.config["spatial"]["semiminor_rad"]
 
+        if isinstance(cnet, str):
+            cnet = from_isis(cnet)
+        cnet = cnet.rename(columns={'id':'identifier'})
+
+        points = cnet.copy(deep=True) # this prevents Pandas value being set on copy of slice warnings
+        points.drop_duplicates(subset=['identifier'])
+        points.insert(0, 'id', list(range(1,len(points)+1)))
```
jlaura, line 2252: If this is the id from the DB, can't one just let the DB auto assign these?

autocnet/graph/network.py
```diff
@@ -2222,50 +2222,102 @@ def clear_db(self, tables=None):
...
+
+        with self.session_scope() as session:
+            imgs = session.query(Images.serial, Images.id).all()  
+        iid_map = {ii[0]: ii[1] for ii in imgs}
+        cnet['imageid'] = cnet.apply(lambda row: iid_map[row['serialnumber']], axis=1)
+        
+        cnet[['templateMetric', 'templateShift']] = None
+        cnet[['phaseError','phaseDiff','phaseShift']] = None
+        cnet[['sampler','liner','residual','rms']] = None
+        cnet['ChooserName'] = None
```
jlaura, line 2280: Set to be ISIS or pull from the input?

autocnet/graph/network.py
```diff
@@ -2222,50 +2222,102 @@ def clear_db(self, tables=None):
...
+        cnet['pointid']  = cnet.apply(lambda row: pid_map[row['identifier']], axis=1)
+
+        with self.session_scope() as session:
+            imgs = session.query(Images.serial, Images.id).all()  
+        iid_map = {ii[0]: ii[1] for ii in imgs}
+        cnet['imageid'] = cnet.apply(lambda row: iid_map[row['serialnumber']], axis=1)
+        
+        cnet[['templateMetric', 'templateShift']] = None
+        cnet[['phaseError','phaseDiff','phaseShift']] = None
+        cnet[['sampler','liner','residual','rms']] = None
```
jlaura, line 2279: Not for this PR. What if start from a post-jigsaw network?

autocnet/io/db/model.py
```diff
@@ -501,6 +503,22 @@ def pointtype(self, v):
...
+    @residuals.setter
+    def residuals(self, v):
+        self._residuals = v
+
+    @hybrid_property
+    def maxresidual(self):
+        return self._maxresidual
+
+    @maxresidual.setter
+    def maxresidual(self, max_res):
```
jlaura, line 519: max residual might be redundant? We can use max() on the python side a la: `[max(residual) for measure in measures for residual in residuals]`

On the SQL side we can do: `SELECT max(x) FROM unnest(residuals) as x;`

Is it nice to have though? - maybe? I'm torn here.

jlaura:
@ladoramkershner I labelled this as a patch bump because this looked most like a 'bug' fix for performance improvement instead of a new feature addition. Please feel free to drop the bump patch label and use the bump minor if you think this is more of a new feature addition.

Nice work on this PR!

autocnet/io/db/model.py
ladoramkershner: I put it in because it is a pain in the butt to calculate anything from the vector (especially in QGIS). Technically we don't need to have the residuals column in the points table at all because it already exists in the Measures table. I am not sure what the right answer is here though..because clogging up the tables is a real concern.

autocnet/graph/network.py
```diff
@@ -2222,50 +2222,102 @@ def clear_db(self, tables=None):
...
+        """
+
         semi_major, semi_minor = self.config["spatial"]["semimajor_rad"], self.config["spatial"]["semiminor_rad"]
 
+        if isinstance(cnet, str):
+            cnet = from_isis(cnet)
+        cnet = cnet.rename(columns={'id':'identifier'})
+
+        points = cnet.copy(deep=True) # this prevents Pandas value being set on copy of slice warnings
+        points.drop_duplicates(subset=['identifier'])
```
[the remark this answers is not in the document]
ladoramkershner: The id's in the original network are the point ids, which should be independent (per point). I am doing this to lessen the number of geometric calculations we have to do, since they are expensive and all measures on a point will have the same geom info.

TLDR; I think this is valid as long as the assumption "all measures have the same point properties (geom, reference index, etc.)" is valid.

autocnet/graph/network.py
ladoramkershner: I think I need to do this ahead of time in order to populate the measures "pointid" column.

autocnet/graph/network.py
ladoramkershner: I think checking for this (and the residuals) would be a good PR update!

autocnet/graph/network.py
jlaura: Depending on how the insert happens that is correct. So I think this is fine this way. The alternative is to insert the Point with associated measures in one ORM call. Since the inserts here happen via pandas, I think this is a great way to do this.

jlaura:
I think that this is good to go!

ladoramkershner:
I am adding in the templateMetric, templateShift, and residual calculations cause those do show up in teh cnet!

autocnet/graph/network.py
```diff
@@ -2222,50 +2222,107 @@ def clear_db(self, tables=None):
...
+        cnet.insert(0, 'id', list(range(1,len(cnet)+1)))
+        pid_map = {ident: pid for ident, pid in zip(points['identifier'], points['id'])}
+        cnet['pointid']  = cnet.apply(lambda row: pid_map[row['identifier']], axis=1)
+
+        with self.session_scope() as session:
+            imgs = session.query(Images.serial, Images.id).all()  
+        iid_map = {ii[0]: ii[1] for ii in imgs}
+        cnet['imageid'] = cnet.apply(lambda row: iid_map[row['serialnumber']], axis=1)
+ 
+        cnet['templateMetric'] = cnet.apply(lambda row: row['measureLog'][0].value if row['measureLog'] else None, axis=1)
```
ladoramkershner, line 2280: Okay...I do not like this line, but I cannot figure out how to do it differently. Here I am assuming that 'GoodnessOfFit' is always in the first index for measureLog, if there is a measureLog. I don't think that is a correct assumption.

But measure log is a list of measureLog objects and I can't figure out how to cycle through it (to check for 'GoodnessOfFit') in a list comprehension.

jlaura:
@ladoramkershner Just want to confirm that you are good with this merging before pushing the button? If so, I can merge tomorrow or feel free to merge as this is approved!

PR #614, opened 2021-11-10
Fixes #584

### LINKED ISSUE #584: AutoCNet requires an ISISROOT for import
When autocnet is imported it also ALWAYS imports the ISIS wrapper kalasiris (used to be pysis). ISIS is leveraged in AutoCNet for spatial interactions with images, namely image_to_ground and ground_to_image calls. Therefore, if you do not have an ISISROOT environment variable set at the time of autocnet import, it fails. However, AutoCNet has capabilities other than spatial image calls AND has an option for CSM cameras to handle those interactions. Therefore, AutoCNet should not be dependent on having an ISISROOT. Ideally, we would sequester the camera import and have hard failures only if a spatial call was made without proper ISIS/CSM camera setups.

I was running into this problem while trying to generate the docs and created a less than ideal work around, when this is resolved, the following files should be updated:
- docs/conf.py
- docs/Makefile
The specific changes to be removed are highlighted in PR #583 .

jlaura:
This fixes #584, but making the kalasiris import optional. I solved this problem by wrapping the import in a try except. On the except, a FailedImport class is matched to the module name. Then any calls to the module using dot notation (that calls `__getattribute__` will raise an ImportError. This way, we can run any functions / methods in a file that do not use the import that failed.

To be even more explicit about what is happening is that kalasiris is attempting to import, but when the os.environ['ISISROOT'] call fails, the import raises. When that import raises, the except block in autocnet is triggered instantiating the FailedImport class.

## How was this tested?

I tested this in an interactive python session by importing autocnet.spatial.isis and then attempting to make use of an `isis.*` call. That module imports kalasiris as isis.

```
>>> from autocnet.spatial.isis import isis
>>> type(isis)
<class 'autocnet.utils.utils.FailedImport'>
>>> isis.campt
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jlaura/github/autocnet/autocnet/utils/utils.py", line 26, in __getattribute__
    raise ImportError('Module was not imported. No classes, attributes, or variables are available.')
ImportError: Module was not imported. No classes, attributes, or variables are available.
```

### DIFF
--- a/autocnet/spatial/isis.py
+++ b/autocnet/spatial/isis.py
@@ -239,7 +239,7 @@ def point_info(
 
         # ISIS's campt needs points in a file
         with isis.fromlist.temp(p_list) as f:
-            cp = campt(
+            cp = isis.campt(
                 cube_path,
                 coordlist=f,
                 allowoutside=allowoutside,
--- a/autocnet/spatial/isis.py
+++ b/autocnet/spatial/isis.py
@@ -18,9 +18,9 @@ import numpy as np
 
 try:
     import kalasiris as isis
-except:
+except Exception as exception:
     from autocnet.utils.utils import FailedImport
-    isis = FailedImport()
+    isis = FailedImport(exception)
 
 import pvl
 
--- a/autocnet/spatial/isis.py
+++ b/autocnet/spatial/isis.py
@@ -15,10 +15,14 @@ from collections import abc
 from numbers import Number
 
 import numpy as np
-import kalasiris as isis
-import pvl
 
-import kalasiris as kal
+try:
+    import kalasiris as isis
+except:
+    from autocnet.utils.utils import FailedImport
+    isis = FailedImport()
+
+import pvl
 
 isis2np_types = {
         "UnsignedByte" : "uint8",
@@ -48,7 +52,7 @@ def get_isis_special_pixels(arr):
 
     """
     isis_dtype = np2isis_types[str(arr.dtype)]
-    sp_pixels = getattr(kal.specialpixels, isis_dtype)
+    sp_pixels = getattr(isis.specialpixels, isis_dtype)
 
     null = np.argwhere(arr==sp_pixels.Null)
     lrs = np.argwhere(arr==sp_pixels.Lrs)
@@ -235,7 +239,7 @@ def point_info(
 
         # ISIS's campt needs points in a file
         with isis.fromlist.temp(p_list) as f:
-            cp = isis.campt(
+            cp = campt(
                 cube_path,
                 coordlist=f,
                 allowoutside=allowoutside,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ release.
 ### Fixed
 - `update_from_jigsaw` failures due to stale code. Now uses a conntext on the engine to ensure closure
 - Fixes errors where reference measure index was being incorrectly tracked when placing measures would fail [#606](https://github.com/USGS-Astrogeology/autocnet/issues/606)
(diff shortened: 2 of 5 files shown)


autocnet/spatial/isis.py
```diff
@@ -235,7 +239,7 @@ def point_info(
 
         # ISIS's campt needs points in a file
         with isis.fromlist.temp(p_list) as f:
-            cp = isis.campt(
+            cp = campt(
```
rbeyer, line 242: Doesn't this still need to be `isis.campt()`?

jlaura: Sure does....I'm chocking that up to my editor helping me with errant keystrokes.

rbeyer:
So I was thinking about how a user would experience this.  They would receive a very mysterious error message that would not help them understand what went wrong ("What %$@# module wasn't imported?" I hear the future users cry out).

Your new `FailedImport` class could take an optional argument on `__init__()` which is an object of type Exception.  And when `__getattribute__()` is called it should just raise that exception that it is holding on to (if it got one on init) rather than inventing its own (you may need to use `raise ImportError from self.exception` or something to get the stack trace to make sense, not sure without testing).

This way, if the module-owner provided a hopefully useful Exception then that might be more useful to the end-user, and also makes your `FailedImport` class more extensible.

jlaura:
Finally found some time to place around a bit with this. Here is a working snippet that I believe fulfills your great suggestion.

```python
class FailedImport():
    def __init__(self, exception):
        self.exception = exception
    def __getattr__(self, name: str):
        raise self.exception            
        
try:
    import foo as bar
except ImportError as exception:
    bar = FailedImport(exception)
```

A few changes and limitations.

  1. Swapped to getattr from getattribute after seeing max recursion errors. This make sense based on [this post](https://stackoverflow.com/questions/3278077/difference-between-getattr-vs-getattribute).
  2. Made the exception a required argument. Why? Because we can not do an attribute lookup in getattr without hitting max recursion calling, you guessed it(!) getattr.
  3. Due to how getattr works, we can not partially turn an import on/off. In other words, this will always raise, if bar.subtract exists, but bar.add fails to import, the whole module is assumed to be turned off.

Here is the new stack trace (from a jupyter notebook, ignore the notebook specific stuff):

```bash
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
<ipython-input-18-acf7fadaacf0> in <module>
----> 1 bar.add

<ipython-input-17-04cfcdceb245> in __getattr__(self, name)
      3         self.exception = exception
      4     def __getattr__(self, name: str):
----> 5         raise self.exception
      6 
      7 try:

<ipython-input-17-04cfcdceb245> in <module>
      6 
      7 try:
----> 8     import foo as bar
      9 except ImportError as exception:
     10     bar = FailedImport(exception)

ModuleNotFoundError: No module named 'foo'
```

rbeyer:
I think making exception required is fine, I mean, the only reason to implement this is because you're catching some kind of exception related to imports that you want to hold on to.  You could still implement it, by making `exception=None` in `__init__()` and then putting an `if self.exception is None` check in `__getattr__()` but I don't think it is needed.  You could also get fancy and write this as a context manager to streamline things, but that's certainly extra credit.

Regarding partial import, I'm not sure what you're talking about, when you say `import x` either the symbol `x` gets imported or it doesn't.  If you separately say `import subtract from x` or `import add from x` those are two separate import activities and could each be separately wrapped with a try/except, and after that Python doesn't particularly remember if `subtract` and `add` came from the same module.  So I think this is a non-issue.

rbeyer (approved):

PR #620, opened 2021-12-07
Updated graph module to use logging 

gsn9:
So far I've updated the `cluster_submit.py` and `edge.py` to use logging over warnings. I'm currently working on converting `network.py` over  but I just wanted to open a draft up before the end of the day.

### DIFF
--- a/autocnet/graph/cluster_submit.py
+++ b/autocnet/graph/cluster_submit.py
@@ -5,7 +5,7 @@ import copy
 import os
 import json
 import sys
-import warnings
+import logging
 
 from io import StringIO 
 from contextlib import redirect_stdout
@@ -19,6 +19,9 @@ from autocnet.io.db.model import Points, Measures, Overlay
 from autocnet.utils.utils import import_func
 from autocnet.utils.serializers import JsonEncoder, object_hook
 from autocnet.io.db.model import JobsHistory
+
+# set up the logging file
+log = logging.getLogger(__name__)
 
 def parse_args():  # pragma: no cover
     parser = argparse.ArgumentParser()
@@ -178,7 +181,7 @@ def manage_messages(args, queue):
         
         if msg is None:
             if args['queue'] == False:
-                warnings.warn('Expected to process a cluster job, but the message queue is empty.')
+                log.warning('Expected to process a cluster job, but the message queue is empty.')
                 return
             elif args['queue'] == True:
                 print(f'Completed processing from queue: {queue}.')
@@ -197,8 +200,9 @@ def manage_messages(args, queue):
         with redirect_stdout(stdout):
             # Apply the algorithm
             response = process(msgdict)
-            # Should go to a logger someday!
-            print(response)
+            # Should go to a logger someday! (today is that day!)
+            log.info(response)
+            
 
         out = stdout.getvalue()
         # print to get everything on the logs in the directory
@@ -226,6 +230,9 @@ def manage_messages(args, queue):
 
 def main():  # pragma: no cover
     args = vars(parse_args())
+    # set up the logger
+    logging.basicConfig(level=os.environ.get("autocnet_loglevel", "INFO"))
+    log.error("hello")
     # Get the message
     queue = StrictRedis(host=args['host'], port=args['port'], db=0)
     manage_messages(args, queue)
--- a/autocnet/graph/tests/test_cluster_submit.py
+++ b/autocnet/graph/tests/test_cluster_submit.py
@@ -5,6 +5,7 @@ from unittest.mock import patch
 
 import numpy as np
 import pytest
+import logging
 
 from autocnet.utils.serializers import JsonEncoder, object_hook
 from autocnet.graph import cluster_submit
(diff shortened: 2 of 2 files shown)


autocnet/graph/tests/test_node.py
```diff
@@ -127,11 +130,11 @@ def test_isis_serial(self, node):
...
 
-    def test_save_load(self, node, tmpdir):
+    def test_save_load(self, node, tmpdir,caplog):
         # Test that without keypoints this warns
-        with pytest.warns(UserWarning) as warn:
-            node.save_features(tmpdir.join('noattr.npy'))
-        assert len(warn) == 1
+        node.save_features(tmpdir.join('noattr.npy'))
+        expected = 'Node {} has not had features extracted.'.format(node['node_id'])
+        assert expected in caplog.text
```
gsn9, line 138: I changed the way this test is testing for success. Rather than checking that the warning(log) is of length 1, it's now testing to ensure the log contains the expected message. I think this is an okay change? unless we want to ensure that there is only 1 log/warning being generated here.

autocnet/graph/cluster_submit.py
```diff
@@ -20,6 +20,9 @@
 from autocnet.utils.serializers import JsonEncoder, object_hook
 from autocnet.io.db.model import JobsHistory
 
+# set up the logging file
```
jlaura, line 23: I would yank the comment. It is obvious what the next line does.

autocnet/graph/cluster_submit.py
```diff
@@ -226,6 +230,8 @@ def manage_messages(args, queue):
 
 def main():  # pragma: no cover
     args = vars(parse_args())
+    # set up the logger
+    logging.basicConfig(level=os.environ.get("autocnet_loglevel", "INFO"))
```
jlaura, line 234: Can we go all caps on an environment variable?

autocnet/graph/edge.py
```diff
@@ -176,7 +179,7 @@ def ring_match(self, *args, **kwargs):
         tar_desc = self.destination.descriptors
 
         if not 'xm' in ref_kps.columns:
-            warnings.warn('To ring match body centered coordinates (xm, ym, zm) must be in the keypoints')
+            log.warning('To ring match body centered coordinates (xm, ym, zm) must be in the keypoints')
```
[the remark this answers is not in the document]
jlaura: Leaving them untested is fine. I don't think we want the scope of this to get really large.

autocnet/graph/tests/test_cluster_submit.py
```diff
@@ -102,9 +104,11 @@ def test_finalize_message_from_work_queue(args, queue, simple_message):
     cluster_submit.finalize_message_from_work_queue(queue, args['working_queue'], remove_key)
     assert queue.llen(args['working_queue']) == 0
     
-def test_no_msg(args, queue):
-    with pytest.warns(UserWarning, match='Expected to process a cluster job, but the message queue is empty.'):
-        cluster_submit.manage_messages(args, queue)
+def test_no_msg(caplog,args, queue):
+    cluster_submit.manage_messages(args, queue)
+    expected_log = 'Expected to process a cluster job, but the message queue is empty.'
+    assert expected_log in caplog.text
```
jlaura, line 110: :+1:

autocnet/graph/tests/test_cluster_submit.py
```diff
@@ -181,6 +190,7 @@ def test_process_generic(along, func, msg_additions, mocker):
     mocker.patch('autocnet.graph.cluster_submit._instantiate_obj', side_effect=_generate_obj)
     mocker.patch('autocnet.graph.network.NetworkCandidateGraph.Session', return_value=True)
     mocker.patch('autocnet.graph.network.NetworkCandidateGraph.config_from_dict')
+    _do_something()
```
jlaura, line 193: This should be in a stand along test, not piggy backed into the generic process test.

You could copy one of the generic process argument sets: `([1,2,3,4,5], _do_nothing, {})`, swapping in the `_do_something` func. The `_do_something` func could then take a single argument, the log level. It *should* be possible to then pass in a log level (e.g., debug or error) dynamically to `_do_something` and have it return a message at the correct log level. Something like:

```python
def _do_something(log_level):
    return getattr(log, log_level)(f'Logging at the {log_level}')
```

which should return a logging string at the passed log level. You could pass an expected message, log level, and a log level env variable to set, into the test using pytest.mark.parametrize and test log level / env variable combinations to ensure everything is setup properly.

autocnet/graph/tests/test_edge.py
```diff
@@ -16,6 +16,8 @@
 from .. import edge
 from .. import node
 
+import logging
+log = logging.getLogger(__name__)
```
jlaura, line 20: What is the rationale behind having the logger in the testing suite?

gsn9: I could be wrong, but I was under the impression that the logger needed to be defined to test the stdout of the logs that are being generated.

PR #621, opened 2021-12-07
converted spatial module to use logging

gsn9:
I converted the few warning statements that were in the `overlap.py` file, there weren't any test's for this file and the other two files in this module didn't seem to using warnings at all. I'm wondering if it would be worth it to create a `test_overlap.py` file to test the logging and possibly the rest of the code in this file.

### DIFF
--- a/autocnet/spatial/overlap.py
+++ b/autocnet/spatial/overlap.py
@@ -1,5 +1,5 @@
 import time
-import warnings
+import logging
 import json
 from subprocess import CalledProcessError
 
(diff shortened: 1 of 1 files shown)


jlaura (approved):

jlaura:
I don't know that we need explicit tests at the module level to check for logging. I wonder if more thorough tests at the graph/cluster_submit.py level is not adequate because the 'logic' for logging is generic across the entire rest of the library.

PR #622, opened 2021-12-07
updated transformation module to use logging

gsn9:
Only file that needed to be changed to use logging was `fundamental_matrix.py`


autocnet/transformation/tests/test_fundamental_matrix.py
```diff
@@ -1,7 +1,7 @@
 import os
 import sys
 import unittest
-import warnings
```
gsn9, line 4: removed unused import

jlaura (approved):

PR #623, opened 2021-12-08
Adds an integration test for sub pixel matching

jlaura:
This PR adds an integration tests using live CTX data for the adaptive sub pixel matcher. In making the test, this also made a change to the isis wrapper where a None value can be returned. This was added to handle a case where ground_to_image or image_to_ground calls with sequences of coordinates and the allowoutside flag set to True would not abort and raise if ISIS returned a valid NULL for the ground/image coordinates.


autocnet/spatial/isis.py
```diff
@@ -253,11 +253,15 @@ def point_info(
                 # convert all pixels to PLIO pixels from ISIS
                 r["Sample"] -= .5
                 r["Line"] -= .5
+                
                 results.append(r)
             else:
-                raise ValueError(
-                    f"ISIS campt completed, but reported an error: {r['Error']}"
-                )
```
jlaura, line 260: @rbeyer @Kelvinrr What are your thoughts about this change?

My thinking: multiple sequential campt calls are expensive when we have to keep paying the os.subprocess startup costs. This functions supports sequences, but all of the passed elements need to be valid under the current logic in this wrapper. On the ISIS side, we get back PVL filled with NULLs. So this func should propagate those to the caller to handle as they wish. 

Downside: Using 'None' loses the error message. An alternative would be to return a string or a dict. The downside to that is that the caller would get an iterable that is of mixed types.

Kelvinrr: I personally prefer this method. Doing `None` checks is just part of the python life. The only alternative I can think of is to simply not append and throw a warning that projection failed on index N.

rbeyer: This is either/or for me.  If you want to guarantee that the value returned always has "good" data, then leave it (that was the original design, if there was an Error, then that's a problem).  If you want to pay the `is None` price later, this change is fine.  Itis largely a philosophy question.

If you make this change, you should also change the docstring for this function to let callers know that there might be `None` values in there.

jlaura:
I'm not 100% sure why the CI is failing. It is on the curl command pulling the new test data. The one test failure that I am seeing is on the logging stuff that was previously added. I can get an issue opened about the test that I am seeing fail. It will be in a > 1.0.0 milestone.

PR #624, opened 2021-12-09
Logging for matcher module

gsn9:
This might be a longer pr as I did all of the matcher module's files in one go. I also added logging in several places where we are raising a value error. I can remove them if it's unneeded but I was thinking it would be good to log anytime a value error occurs.

### DIFF
--- a/autocnet/matcher/ciratefi.py
+++ b/autocnet/matcher/ciratefi.py
@@ -1,5 +1,4 @@
 import math
-# import warnings
 import logging
 from bisect import bisect_left
 
--- a/autocnet/matcher/ciratefi.py
+++ b/autocnet/matcher/ciratefi.py
@@ -1,5 +1,6 @@
 import math
-import warnings
+# import warnings
+import logging
 from bisect import bisect_left
 
 import cv2
@@ -11,6 +12,7 @@ from scipy.ndimage.interpolation import zoom
 
 import autocnet.utils.utils as util
 
+log = logging.getLogger(__name__)
 
 def cifi(template, search_image, thresh=90, use_percentile=True,
          radii=list(range(1,12)), scales=[0.5, 0.57, 0.66,  0.76, 0.87, 1.0], verbose=False):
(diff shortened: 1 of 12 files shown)


autocnet/matcher/ciratefi.py
```diff
@@ -75,17 +76,19 @@ def cifi(template, search_image, thresh=90, use_percentile=True,
 
     radii = np.asarray(radii)
     if not radii.size or not np.any(radii):
+        log.error('Input radii list is empty')
         raise ValueError('Input radii list is empty')
```
gsn9, line 81: these are the log's I mentioned before

autocnet/matcher/tests/test_ciratefi.py
```diff
@@ -51,10 +52,16 @@ def offset_template(img, img_coord):
...
+def test_cifi_radii_too_large(template, search, caplog):
+    # check all logs
+    ciratefi.cifi(template, search, 1.0, radii=[100], use_percentile=False)
+    num_pattern = '\d+'
+    captured_log = caplog.records[0].getMessage()
+    match = re.findall(num_pattern, captured_log)
+    assert (f'Max Radii is larger than original template, '\
+        f'this may produce sub-par results.Max radii: {match[0]} max template dimension: {match[1]}'\
+             == caplog.records[0].getMessage())
+
```
gsn9, line 64: I kinda changed the way this test is testing for the right log/warning now. With using logging I didn't find a simple way to pull the template dimensions so rather then going down that rabbit hole I decided to use a regular expression. Basically I'm just searching the captured log for the dimensions and then just checking to make sure that the expected matches the results.

autocnet/matcher/tests/test_outlier_detector.py
```diff
@@ -65,11 +64,11 @@ def test_suppress(self):
         self.assertEqual(mask.sum(), 34)
 
     def test_suppress_non_optimal(self):
-        with warnings.catch_warnings(record=True) as w:
+        # with warnings.catch_warnings(record=True) as w:
+        with self.assertLogs() as captured:
             mask, k = cpu_outlier_detector.spatial_suppression(self.df, self.domain, k=30, xkey='lon', ykey='lat')
             self.assertEqual(k, 59)
-            self.assertTrue(issubclass(w[0].category, UserWarning))
```
gsn9, line 71: I removed this line as I removed the warnings import, I think this should be okay?

Kelvinrr (approved):
Looks reasonable to me assuming CI failures are expected.

PR #626, opened 2021-12-15
updated mutual _information interface

gsn9:
Updated the mutual information interface

### DIFF
--- a/autocnet/matcher/mutual_information.py
+++ b/autocnet/matcher/mutual_information.py
@@ -3,8 +3,9 @@ from math import floor
 import numpy as np
 
 from scipy.ndimage.measurements import center_of_mass
+from skimage.transform import AffineTransform
 
-def mutual_information(t1, t2, **kwargs):
+def mutual_information(reference_roi, walking_roi, affine=AffineTransform(), **kwargs):
     """
     Computes the correlation coefficient between two images using a histogram
     comparison (Mutual information for joint histograms). The corr_map coefficient
@@ -13,12 +14,13 @@ def mutual_information(t1, t2, **kwargs):
     Parameters
     ----------
 
-    t1 : ndarray
-         First image to use in the histogram comparison
-
-    t2 : ndarray
-         Second image to use in the histogram comparison
-
+    reference_roi : Roi
+                    First image to use in the histogram comparison
+    
+    moving_roi : Roi
+                   Second image to use in the histogram comparison
+    
+    
     Returns
     -------
 
@@ -30,16 +32,23 @@ def mutual_information(t1, t2, **kwargs):
     --------
     numpy.histogram2d : for the kwargs that can be passed to the comparison
     """
-
-    if np.isnan(t1).any() or np.isnan(t2).any():
+    
+    reference_image = reference_roi.clip()
+    walking_template = walking_roi.clip(affine)
+    
+    # walking_template = tf.warp(walking_template, affine, order=3)
+    
+    
+    if reference_roi.ndv == None or walking_roi.ndv == None:
         print('Unable to process due to NaN values in the input data')
         return
     
-    if t1.shape != t2.shape:
+    # print(reference_roi.size_y, walking_roi.size_y)
+    if reference_roi.size_y != walking_roi.size_y and reference_roi.size_x != walking_roi.size_x:
         print('Unable compute MI. Image sizes are not identical.')
         return
 
-    hgram, x_edges, y_edges = np.histogram2d(t1.ravel(),t2.ravel(), **kwargs)
+    hgram, x_edges, y_edges = np.histogram2d(reference_image.ravel(), walking_template.ravel(), **kwargs)
 
     # Convert bins counts to probability values
     pxy = hgram / float(np.sum(hgram))
--- a/autocnet/matcher/tests/test_mutual_information.py
+++ b/autocnet/matcher/tests/test_mutual_information.py
@@ -3,14 +3,14 @@ import os
 import sys
 import unittest
 from unittest.mock import patch
-
+from autocnet.transformation.roi import Roi
 import pytest
 import numpy as np
 
 from .. import mutual_information
 
 def test_good_mi():
-    test_image1 = np.array([[i for i in range(50)] for j in range(50)])
+    test_image1 = Roi(np.array([[i for i in range(50)] for j in range(50)]), 25, 25, 25, 25, ndv=22222222)
     corrilation = mutual_information.mutual_information(test_image1, test_image1)
     assert corrilation == pytest.approx(2.30258509299404)
 
--- a/autocnet/transformation/affine.py
+++ b/autocnet/transformation/affine.py
@@ -65,9 +65,13 @@ def estimate_affine_transformation(reference_image,
     x_coords = [base_startx, base_startx, base_stopx, base_stopx, bcenter_x]
     y_coords = [base_starty, base_stopy, base_stopy, base_starty, bcenter_y]
 
-    # Dispatch to the sensor to get the a priori pixel location in the input image
-    lons, lats = isis.image_to_ground(reference_image.file_name, x_coords, y_coords, allowoutside=True)
-    xs, ys = isis.ground_to_image(moving_image.file_name, lons, lats, allowoutside=True)
+    try:
+        # Dispatch to the sensor to get the a priori pixel location in the input image
+        lons, lats = isis.image_to_ground(reference_image.file_name, x_coords, y_coords, allowoutside=True)
+        xs, ys = isis.ground_to_image(moving_image.file_name, lons, lats, allowoutside=True)
+    except Exception as e:
+        print(e.stderr)
+        raise e
 
     # Check for any coords that do not project between images
     base_gcps = []
--- a/autocnet/transformation/roi.py
+++ b/autocnet/transformation/roi.py
@@ -1,7 +1,6 @@
 from math import modf, floor
 import numpy as np
 
-from autocnet.transformation import dtypes
 import pvl
 
 class Roi():


autocnet/matcher/mutual_information.py
```diff
@@ -1,10 +1,11 @@
 from math import floor
-
+from autocnet.transformation.roi import Roi
 import numpy as np
 
 from scipy.ndimage.measurements import center_of_mass
+from skimage.transform import AffineTransform
 
-def mutual_information(t1, t2, **kwargs):
+def mutual_information(reference_roi, walking_roi, affine=AffineTransform(), **kwargs):
```
jlaura, line 8: I prefer moving_roi, as it is used below, but I defer to you all on what you like better. Both are descriptive. One seems anthropomorphic. @acpaquette @Kelvinrr

autocnet/matcher/mutual_information.py
```diff
@@ -13,12 +14,13 @@ def mutual_information(t1, t2, **kwargs):
     Parameters
     ----------
 
-    t1 : ndarray
-         First image to use in the histogram comparison
-
-    t2 : ndarray
-         Second image to use in the histogram comparison
-
+    reference_roi : Roi
```
jlaura, line 17: I thought that this func is basically just a helper that performs the actual computation of the MI metric. If that is the case, should this just stay ndarrays (because nothing in here needs any additional information the ROI carries and this is a very low level, almost helper func)?

autocnet/matcher/mutual_information.py
```diff
@@ -30,16 +32,17 @@ def mutual_information(t1, t2, **kwargs):
...
-        return
     
-    if t1.shape != t2.shape:
-        print('Unable compute MI. Image sizes are not identical.')
-        return
+    reference_image = reference_roi.array
+    walking_template = walking_roi.array
+    
+    if reference_roi.ndv == None or walking_roi.ndv == None:
+        raise Exception('Unable to process due to NaN values in the input data')
```
jlaura, line 40: This breaks the logic. Line 34 above checks if the ndaray has any null values in it. The new line 39 checks if the no data value attribute has been set on the ROI object. Those are very different checks. I would advocate that this func maintain using ndarrays and keep the above checks.

autocnet/matcher/mutual_information.py
```diff
@@ -30,16 +32,17 @@ def mutual_information(t1, t2, **kwargs):
...
-
-    if np.isnan(t1).any() or np.isnan(t2).any():
-        print('Unable to process due to NaN values in the input data')
-        return
     
-    if t1.shape != t2.shape:
-        print('Unable compute MI. Image sizes are not identical.')
-        return
+    reference_image = reference_roi.array
+    walking_template = walking_roi.array
```
jlaura, line 37: reference_array and moving_array maybe? The whole image/template paradigm should go for readability.

autocnet/matcher/mutual_information.py
```diff
@@ -92,50 +95,63 @@ def mutual_information_match(d_template, s_image, subpixel_size=3,
     if func == None:
         func = mutual_information
 
-    image_size = s_image.shape
```
jlaura, line 95: This is the user called func that should be using the ROI objects as far as I know.

autocnet/matcher/mutual_information.py
```diff
@@ -92,50 +95,63 @@ def mutual_information_match(d_template, s_image, subpixel_size=3,
...
-    template_size = d_template.shape
+
+    if isinstance(s_image, Roi):
+        image_size = s_image.array.shape#(s_image.size_x, s_image.size_y)
+        template_size = d_template.array.shape# (d_template.size_x, d_template.size_y)
+
+    else:
+        image_size = s_image.shape
+        template_size = d_template.shape
```
jlaura, line 105: If the API is just supporting ROIs (which I thought we agreed it should?), then this should support the ROI. Also, ROI reading is expensive (it is disk I/O), so things like shape should be pulled as attributes off the ROI object whenever possible.

autocnet/matcher/mutual_information.py
```diff
@@ -92,50 +95,63 @@ def mutual_information_match(d_template, s_image, subpixel_size=3,
...
     max_j = -1  # x
-    for i in range(y_diff+1):
-        for j in range(x_diff+1):
-            sub_image = s_image[i:i+template_size[1],  # y
-                                j:j+template_size[0]]  # x
-            corr = func(sub_image, d_template, **kwargs)
+
+    s_image_extent = s_image.image_extent
+
+    for i in range(s_image_extent[2],s_image_extent[3]):
```
jlaura, line 117: for me, x_diff, y_diff, while note great, are much better than indices into extents when I have no idea what the extent ordering is (e.g., [xstart, xstop, ystart, ystop], [xstart, ystart, xstop, ystop], [ystart, ystop, xstart, xstop], etc...)

autocnet/matcher/mutual_information.py
```diff
@@ -92,50 +95,63 @@ def mutual_information_match(d_template, s_image, subpixel_size=3,
...
+    x = abs(x - (corr_map.shape[0])/2)
     y += subpixel_y_shift
     x += subpixel_x_shift
-
-    # Compute the idealized shift (image center)
-    y -= (s_image.shape[0] / 2) - (d_template.shape[0] / 2)
-    x -= (s_image.shape[1] / 2) - (d_template.shape[1] / 2)
-
-    return float(x), float(y), float(max_corr), corr_map
+    new_affine = AffineTransform(translation=(-x, -y))
```
jlaura, line 155: @acpaquette  @Kelvinrr  Just a thought - instantiating an affine transform object here tightly couples this code to the skim age AffineTransform. Probably okay. Cleaner to return a 3x3 ndarray and then let the caller pack that into an affine?

autocnet/transformation/affine.py
```diff
@@ -65,9 +65,13 @@ def estimate_affine_transformation(reference_image,
     x_coords = [base_startx, base_startx, base_stopx, base_stopx, bcenter_x]
     y_coords = [base_starty, base_stopy, base_stopy, base_starty, bcenter_y]
 
-    # Dispatch to the sensor to get the a priori pixel location in the input image
-    lons, lats = isis.image_to_ground(reference_image.file_name, x_coords, y_coords, allowoutside=True)
-    xs, ys = isis.ground_to_image(moving_image.file_name, lons, lats, allowoutside=True)
+    try:
```
jlaura, line 68: why is the try/except wrapped? Should this be using the logger?

autocnet/matcher/tests/test_mutual_information.py
```diff
@@ -26,9 +26,11 @@ def test_mutual_information():
 
     s_image[25:75, 25:75] = d_template
 
-    x_offset, y_offset, max_corr, corr_map = mutual_information.mutual_information_match(d_template, s_image, bins=20)
-    assert x_offset == 0.01711861257171421
```
jlaura, line 30: can't we still assert shifts by checking the new pixel location through applying the returned affine?

autocnet/matcher/tests/test_mutual_information.py
```diff
@@ -26,9 +26,11 @@ def test_mutual_information():
...
-    assert y_offset == 0.0
+    d_template = Roi(d_template, 25, 25, 25, 25, ndv=22222222)
+    s_image = Roi(s_image, 50, 50, 25, 25, ndv=22222222)
+    affine, max_corr, corr_map = mutual_information.mutual_information_match(d_template, s_image, bins=20)
+
     assert max_corr == 2.9755967600033015
-    assert corr_map.shape == (51, 51)
+    assert corr_map.shape == (50, 50)
```
jlaura, line 34: Can we also add a test with an odd sized s_image to check that we are handling even and odd sized rois properly inside these funcs.

autocnet/transformation/roi.py
```diff
@@ -163,7 +162,7 @@ def array(self):
         """
         pixels = self.image_extent
         if isinstance(self.data, np.ndarray):
-            data = self.data[pixels[2]:pixels[3]+1,pixels[0]:pixels[1]+1]
+            data = self.data[pixels[2]:pixels[3],pixels[0]:pixels[1]]
```
jlaura, line 165: Is this then the proper size? the NumPy index is not inclusive.

jlaura (changes requested):

autocnet/transformation/roi.py
acpaquette: @jlaura I was examining this in smaller case and say the following:
```
>>> roi.Roi(np.ones((50, 50)), 25, 25, 25, 25).array.shape
(50, 50)
>>> roi.Roi(np.ones((100, 100)), 25, 25, 25, 25).array.shape
(51, 51)
```

autocnet/matcher/mutual_information.py
Kelvinrr: I feel that would be a but clunky to shove the array into a constructor every time. Especially if we use the AffineTransform object to build them we would have to dump the matrix just to shove it back into an AffineTransform object? Unless skimage AffineTransform is super unstable or something I think it's fine.

autocnet/transformation/roi.py
jlaura: Does this have to do with floor/ceil calls when determining the size from a non-integer index?

autocnet/matcher/mutual_information.py
jlaura: I know that this was marked resolved, but nothing changed. Is this not the helper func that is called by mutual_information_match? If so, why is this using ROIs at all?

@acpaquette

autocnet/matcher/mutual_information.py
```diff
@@ -30,16 +32,22 @@ def mutual_information(t1, t2, **kwargs):
     --------
     numpy.histogram2d : for the kwargs that can be passed to the comparison
     """
+    
+    # grab ndarray from input Roi's 
+    reference_image = reference_roi.clip()
```
jlaura, line 37: If this remains, a new ROI will be clipped with every iteration from mutual_information_match(). The performance will be quite poor when every time the moving image has to step to a new pixel location in the reference a new ROI is clipped.

autocnet/matcher/mutual_information.py
jlaura: I see resolved, but I don't see any change. If I'm wrong - let me know why so that I can better understand the code flow.

jlaura (changes requested):

PR #629, opened 2021-12-17
Phase Roi Update

acpaquette:
Updates the iterative_phase function and subpixel_phase functions to use Roi objects rather than numpy arrays.

These updated functions are now also going to return affine transformations rather than updated coordinates. The affine matrices for both iterative_phase and subpixel_phase are walking_image to reference image.


autocnet/matcher/subpixel.py
```diff
@@ -210,37 +207,33 @@ def subpixel_phase(sx, sy, dx, dy,
...
+                 translation from the walking_roi to the reference_roi
+    : tuple
+      With the RMSE error and absolute difference in phase
     """
-    image_size = check_image_size(image_size)
-
-    s_roi = roi.Roi(s_img, sx, sy, size_x=image_size[0], size_y=image_size[1])
-    d_roi = roi.Roi(d_img, dx, dy, size_x=image_size[0], size_y=image_size[1])
+    reference_image = reference_roi.clip()
+    walking_template = walking_roi.clip(affine)
```
jlaura, line 229: It does not look like the ROI object was updated to take an affine in the subpixelapi branch. Did I miss the change? I also don't see the change in the PR.

autocnet/matcher/subpixel.py
```diff
@@ -210,37 +207,33 @@ def subpixel_phase(sx, sy, dx, dy,
...
     ----------
-    template : ndarray
-               The template used to search
-
-    search : ndarray
-             The search image
+    reference_roi : Object
+                    An Roi object from autocnet, the reference image to use when computing subpixel offsets.
+                    Contains either an ndarray or a GeoDataset Object
+    walking_roi : Object
```
jlaura, line 213: @gsn9 @Kelvinrr walking_roi vs moving_roi - can you all pick one and stick with it? It will make things more readable and let me write an architecture document about the sub pixel API more easily.

autocnet/matcher/subpixel.py
```diff
@@ -254,21 +247,18 @@ def subpixel_phase(sx, sy, dx, dy,
...
-    dx = d_roi.x - shift_x
-    dy = d_roi.y - shift_y
+        reference_image = reference_roi.clip()
+        walking_template = walking_roi.clip(affine)
 
-    return dx, dy, error, None
+    (shift_y, shift_x), error, diffphase = registration.phase_cross_correlation(reference_image,
+                                                                                walking_template,
+                                                                                **kwargs)
+    new_affine = AffineTransform(translation=(shift_x, shift_y))
```
jlaura, line 260: The template PR that was merged, I believe was applying -shift_x, -shift_y. This is shift_x, shift_y. Shouldn't these be consistent? Seem that the caller would need to apply direct vs inverse transformation if these are swapped.

@Kelvinrr

autocnet/matcher/subpixel.py
```diff
@@ -254,21 +247,18 @@ def subpixel_phase(sx, sy, dx, dy,
...
+    return new_affine, (error, diffphase)
```
jlaura, line 261: Similar comment. On the template PR, the return was a straight list, so unpacking is different than here. Shouldn't these be standardized?

jlaura (changes requested):

autocnet/matcher/subpixel.py
acpaquette: @jlaura I made this change but likely reverted it in the Roi clip function. I have been working on adding it back in but I am running into issues between the GeoDataset object and just a numpy array as the data. We basically lose all notion of a x, any y coordinate after cropping clipping the data. So applying the affine to get a subwindow of the Roi no longer operates on a GeoDataset object but a numpy array. 

This is difficult as our idea of an individual x, y coord that is agnostic of the window we are extracting is no longer present. If your x, y coord is (250, 250) and you extract a (200, 200) window over a numpy array, (250, 250) no longer work as array indices. It would be nice to maintain the idea of a center coordinate that is truly agnostic of the array size, which is what I have been fighting with in the Roi object

autocnet/matcher/subpixel.py
acpaquette: I figured we could do a pass once all of these work is finished and tidied up. I agree we should pick one. We never really decided which

autocnet/matcher/subpixel.py
acpaquette: We are trying to keep them consistent but the different subpixel matchers produce different x, y shifts. Some produce shifts from reference -> walking while others, like the Phase matcher produce x, y shifts from walking -> reference. I think it is reasonable to expect all of our subpixel matchers to produce one of the above forms but that means certain matchers will return `AffineTransform(translation=(shift_x, shift_y))` or AffineTransform(translation=(-shift_x, -shift_y))

autocnet/matcher/subpixel.py
acpaquette: Chain these algorithms together is something that we discussed but I don't know if we got to this in practice. I think a straight list is fine for now

jlaura: cool - this is something else that we can standardize post merge on all of these.

autocnet/matcher/subpixel.py
jlaura: Gotcha, without playing around in a notebook with this, I suspect any 'ideas' I have would be half baked and less than helpful. I'll leave this PR open until this change is pushed in though because I think without it, this code will not function. Is that a correct assumption?

acpaquette: I think that we can handle this differently in the algorithms to avoid this problem. Unfortunately it makes the algorithms messy. Rather than updating the affine per run of iterative affine, we can update the position of the x, y coordinate in the Roi and add that to the final affine. This prevents the iterative clipping, avoiding this problem mentioned previously. However, it does require that we maintain the state of the passed in Roi objects which is messy, since the algorithm updates the internal state of the Roi object

autocnet/transformation/roi.py
```diff
@@ -187,5 +186,28 @@ def clip(self, dtype=None):
...
+            self.size_x *= 2
+            self.size_y *= 2
+
+            array_to_warp = self.array
+
+            # if array_to_warp.shape != ((self.size_y * 2) + 1, (self.size_x * 2) + 1):
+            #     raise ValueError("Unable to enlarge Roi to apply affine transformation." +
+            #                      f" Was only able to extract {array_to_warp.shape}, when " +
+            #                      f"{((self.size_y * 2) + 1, (self.size_x * 2) + 1)} was asked for. Select, " +
+            #                      "a smaller region of interest" )
```
acpaquette, line 201: This can likely go since we are wrapping with mode set to reflect

